PR #9, opened 2019-10-10
Allow enabling legacy ingress provider, but not by default

### LINKED ISSUE #7: [Feature Request] Enable --providers.kubernetesingress
Adding an option to enable the compatibility to Kubernetes Ingress would be great.

dduportal:
Signed-off-by: dduportal <1522731+dduportal@users.noreply.github.com>

Fixes #7 .

WDYT?

### DIFF
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -80,4 +80,7 @@ spec:
           - "--api.insecure=true"
           - "--ping=true"
           - "--providers.kubernetescrd"
+          {{- if .Values.providers.kubernetesingress }}
+          - "--providers.kubernetesingress"
+          {{- end }}
           - "--log.level={{ .Values.logs.loglevel }}"
--- a/values.yaml
+++ b/values.yaml
@@ -20,6 +20,9 @@ ports:
 logs:
   loglevel: INFO
 #
+providers:
+  kubernetesingress: false
+#
 yaml:
   resources:
     requests:


sgandon:
I have tried you branch and enabling the ingress controller (which by the way should be enabled by default as looking and the root readme).
but none of my usual ingress are working.

dduportal:
Hi @sgandon , could you provide here:

- the result of `kubectl get pod --namespace=<namespace> <traefik-pod-name> -o yaml`
- The Helm command you used for the installation (with the values.yaml) 

please?

sgandon:
Hi @dduportal 
Thanks for looking into it.
In fact some things are working, the service that I used to deploy was expecting https call but it seems that by default traefik is not configured with ssl enabled.
I used to launch the traefik V1 chart with `--set ssl.enabled=true` to have the chart generate a certificate automatically.
When I adapted my service to no use ssl, this traefik V2 chart works fine.
So thank for taking the time to help me.

I have launched the chart using this command
```
helm install . -n traefik --set logs.loglevel=DEBUG --set providers.kubernetesingress=true --debug
```
I have looked at the dashboard and my ingress seems to have created the right routers

|rule|entrypoint|name|service|provider|
|---|---|---|---|---|
| PathPrefix(`/dashboard`) \|\| PathPrefix(`/api`) | web | default-traefik-dashboard-d012b7f875133eeab4e5@kubernetescrd | default-traefik-dashboard-d012b7f875133eeab4e5 |  k8s|
| Host(`iam.k8s`) && PathPrefix(`/`) | traefik                            web                            websecure | iam-k8s-@kubernetes | default-tpsvc-iam-idp-80 |  k8s|
| Host(`portal.k8s`) && PathPrefix(`/`) | traefik                            web                            websecure | portal-k8s-@kubernetes | default-tpsvc-portal-80 |  k8s|

dduportal:
> I used to launch the traefik V1 chart with --set ssl.enabled=true to have the chart generate a > certificate automatically.

It might be related to the fact that we dropped the support of annotations in Ingresses: we expect people to migrate to the new "CRD" `IngressRoute` instead, as explained in the intro of the page https://docs.traefik.io/v2.0/providers/kubernetes-crd/.

So any annotations of your previous ingresses cannot work: you have to migrate to `IngressRoute`. We provide a migration page in the documentation (https://docs.traefik.io/v2.0/migration/v1-to-v2/) and a migration tool: https://github.com/containous/traefik-migration-tool

sgandon:
not this is not related to that at all, we don't use any traefik specific annotation cause we want to be as agnostic as possible of the IngressControler. 
side not : Using specific CRDs is a lock on Traefik that we try to avoid.
Here my issue was on the fact that ssl was not anabled and that there is not way yet to enable it with automatic certificates creation as it is done on the traefik V1 chart.

kopaygorodsky:
Cert-manager uses Ingress resource, so it's not possible to get it working with traefik 2.0 
One more reason to have this option

dduportal:
Closed in favor of #34

PR #10, opened 2019-10-10
feat: allow service type NodePort

acouette:
Hello,

This PR allows to provision a service of type NodePort. Along with the associated nodePorts values for web and websecure ports.


templates/service.yaml
```diff
@@ -8,14 +8,20 @@ metadata:
...
     targetPort: web
+    {{- if (and (eq .Values.service.type "NodePort") (not (empty .Values.ports.web.nodePort)))}}
+    nodePort: {{ .Values.ports.web.nodePort }}
+    {{- end }}
   - port: 443
     name: websecure
     targetPort: websecure
+    {{- if (and (eq .Values.service.type "NodePort") (not (empty .Values.ports.websecure.nodePort)))}}
+    nodePort: {{ .Values.ports.websecure.nodePort }}
+    {{- end }}
```
dtomcej: needs newline

values.yaml
```diff
@@ -6,11 +6,16 @@ image:
 deployment:
   replicas: 1
 #
+service:
```

values.yaml
```diff
@@ -6,11 +6,16 @@ image:
...
+  type: LoadBalancer
+#
 ports:
   web:
     port: 8000
+    #nodePort: 30000
```
dtomcej:
```suggestion
    # nodePort: 30000
```

values.yaml
```diff
@@ -6,11 +6,16 @@ image:
...
   websecure:
     port: 8443
+    #nodePort: 30001
```
dtomcej:
```suggestion
    # nodePort: 30001
```

values.yaml
acouette: I am not sure to understand: if you comment the service section by default, the service won't have any type. Could you elaborate ?

dtomcej: The service should be commented out, as it should not be a requirement to set.

Please see https://github.com/containous/traefik-helm-chart/commit/08996411524c74d28895541a893116e5bc28eed0 as an example of what I mean about setting the default value if not set.

dtomcej: >Do you want me to add an if/else condition that is more resilient 

Yes please, that is preferable. Ideally we don't want to mandate more values in the values.yaml than is needed, as it ties our hands for backwards compatibility.

acouette: I tried to make it clean but the way helm implement the 'and' made me stack ifs.

PR #11, opened 2019-10-11
Add tolerations and node selector

XciD:


values.yaml
```diff
@@ -25,3 +25,5 @@ yaml:
     limits:
       cpu: "300m"
       memory: "150Mi"
+  nodeSelector: {}
```

values.yaml
```diff
@@ -25,3 +25,5 @@ yaml:
...
+  tolerations: []
```

PR #12, opened 2019-10-11
Add externalTrafficPolicy for LB service

XciD:
If `service.type` is NodePort or LoadBalancer, set `externalTrafficPolicy` to Local to enable [source IP preservation](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-typenodeport)

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -14,8 +14,8 @@ ports:
   traefik:
     port: 9000
 #
-service:
-  externalTrafficPolicy: Cluster
+#service:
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -14,6 +14,9 @@ ports:
   traefik:
     port: 9000
 #
+service:
```

values.yaml
```diff
@@ -14,6 +14,9 @@ ports:
...
+  externalTrafficPolicy: Cluster
```

PR #13, opened 2019-10-13
feat: .Values.service.annotations added for LB tune up

alex-roman:
In case when TLS is terminated by ELB service annotations needed to pass cloud provider's params.


dduportal:
Hi @alex-roman , thanks for this contribution! Do you think that #36 would solve this for you? (it aims to provide a broader way to setup the service)?

alex-roman:
@dduportal Hi Damien, yes it could. Thanks for the heads up! Can you please update the table in readme.md in your PR accordingly to the new params. Really love the chart, traefik, and what you guys doing. Thanks for the job!

PR #15, opened 2019-10-15
Add volume/volumeMounts and config.path property to yaml-section of values.yaml

### LINKED ISSUE #14: Be able to specify volumes and --configfile flag
In order to be able to use custom config-file from a config-map or secret it would be nice to be able to specify volumes/volumeMounts for deployment resource and --configfile="" argument in deployment args.

frelon:
Fixes #14

### DIFF
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -73,3 +73,4 @@ spec:
           - "--ping=true"
           - "--providers.kubernetescrd"
           - "--log.level={{ .Values.logs.loglevel }}"
+          - "--configfile={{ .Values.config.path }}"
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -64,6 +64,12 @@ spec:
         - name: traefik
           containerPort: {{ .Values.ports.traefik.port }}
           protocol: TCP
+        {{- if .Values.yaml }}
+        {{- if .Values.yaml.volumeMounts }}
+        volumeMounts:
+{{ toYaml .Values.yaml.volumeMounts | indent 10 }}
+        {{- end }}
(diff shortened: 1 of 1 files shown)


frelon:
Example config (using empty-dir instead of config-map):

```yaml
## Default values for Traefik
image:
  name: traefik
  tag: 2.0.2
#
deployment:
  replicas: 1
#
ports:
  web:
    port: 8000
  websecure:
    port: 8443
  traefik:
    port: 9000
#
logs:
  loglevel: INFO
#
yaml:
  resources:
    requests:
      cpu: "100m"
      memory: "50Mi"
    limits:
      cpu: "300m"
      memory: "150Mi"
  volumes:
  - name: config
    emptyDir: {}
  volumeMounts:
  - name: config
    mountPath: /config
#
config:
  path: /config/traefik-config.toml
```

crhuber:
Would this also work if one wanted to create a file provider?

frelon:
Sure, as long as the file is in the config-map one could reference it from the main config-file.

jasperkuperus:
Ok, I would like this behavior very much. I was at the point of creating a PR for `volumes` and `volumeMounts`. It's very useful for mounting secrets, e.g. when you want to configure Let's Encrypt with Google Cloud DNS and want to mount a service account.

The PR looks good to me. The only thing I don't understand is: Why namespace these under `yaml`, while e.g. `ports`, `env`, etc. are on top level in `values.yaml`?

What's the status of this PR?

mmatur:
Hi @frelon 

Thanks for your contribution. 

I will close your PR in favor of #84

Duplicate of #84

Issue #21, opened 2019-10-29
feat: File Based Provider Support


crhuber:
We currently require routing to services that are outside Kubernetes. Without File based provider we cant route to them, so it would be nice if in the helm chart you can pass args to enable it

dduportal:
Another use case: Specify a TLS wildcard certificate from a file/secret only available in Traefik's namespace (if you have 20 subdomains for 20 applications in their own 20 namespaces, you might not want to duplicate 20 times the wildcard certificate)

dduportal:
The Pull Request #34 will solve half of this issue, by allowing to enable the file provider with

```
helm install --name=traefik --set=additionalArguments="{--providers.file.filename=/dyn.yml}" .
```

for instance

mmatur:
Fixed by #34

igorbrigadir:
If anyone else is looking here - this is a good working example: https://traefik.io/blog/install-and-configure-traefik-with-helm/ the key was `--providers.file.filename=/data/traefik-config.yaml` and mounting a `type: configMap` volume where you can add any file based configs for middlewares, routers etc.

PR #26, opened 2019-11-15
Ports, service, args and dashboard configuration + minor style changes

elthariel:
The PR contains various improvements, allowing to :

- Configure the deployment/traefik/service ports to add extra entrypoints
- Disable the dashboard service and ingress
- Tune the rolling update config
- Use other service types for the main service
- Add arguments to traefik's startup

It also include minor changes in value to reflect the community standards I've observed in the official helm repo. I might be wrong on that one.

PS: Sorry for the all-in-one PR

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -2,31 +2,81 @@
 image:
   name: traefik
   tag: 2.0.5
+
+#
+# Configure the deployment
 #
 deployment:
+  # Number of pods of the deployment
   replicas: 1
+
+rollingUpdate:
+  maxUnavailable: 1
+  maxSurge: 1
+
+# Additional arguments to be passed at the startup of traeffik
+additionalArguments: []
+#  - "--argument1"
+
+#
+# Configure Traefik entrypoints
 #
 ports:
-  web:
-    port: 8000
-  websecure:
-    port: 8443
+  # The name of this one can't be changed as it is used for the readiness and
+  # liveness probes, but you can adjust its config to your liking
   traefik:
     port: 9000
-#
-#service:
-#  externalTrafficPolicy: Cluster
+    # Defines whether the port is exposed if service.type is LoadBalancer or
+    # NodePort.
+    #
+    # You SHOULD NOT expose the traefik port on production deployments. If you
+    # want to access it from outside of your cluster, use `kubectl proxy` or
+    # create a secure ingress
+    expose: false
+    # The exposed port for this service
+    exposedPort: 9000
+  web:
+    port: 8000
+    expose: true
+    exposedPort: 80
+  websecure:
+    port: 8443
+    expose: true
+    exposedPort: 443
+
+# Options for the main traefik service, where the entrypoints traffic comes
+# from.
+service:
+  type: LoadBalancer
+  # Additional annotations (e.g. for cloud provider specific config)
+  annotations: {}
+  # Additional entries here will be added to the service spec. Cannot contains
+  # type, selector or ports entries.
+  spec: {}
+    # externalTrafficPolicy: Cluster
+    # loadBalancerIp: "1.2.3.4"
+    # clusterIP: "2.3.4.5"
+
+dashboard:
+  service:
+    # Create a service exposing the traefik port in the cluster using a
+    # ClusterIP
+    enabled: true
+  ingress:
+    # Expose the dashboard and api service via an ingress route at /dashboard
+    # and /api This is not secure and SHOULD NOT be enabled on production
+    # deployments
+    enabled: true
 
 logs:
   loglevel: INFO
 #
-yaml:
-  resources:
-    requests:
-      cpu: "100m"
-      memory: "50Mi"
-    limits:
-      cpu: "300m"
-      memory: "150Mi"
-#  nodeSelector: {}
-#  tolerations: []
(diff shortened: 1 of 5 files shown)


values.yaml
```diff
@@ -2,31 +2,81 @@
...
-    requests:
-      cpu: "100m"
-      memory: "50Mi"
-    limits:
-      cpu: "300m"
-      memory: "150Mi"
-#  nodeSelector: {}
-#  tolerations: []
+resources: {}
+  # requests:
```
elthariel: The official helm documentation recommends to suggest values here but not to provide defaults

values.yaml
```diff
@@ -2,31 +2,81 @@
...
+
+# Options for the main traefik service, where the entrypoints traffic comes
+# from.
+service:
+  type: LoadBalancer
+  # Additional annotations (e.g. for cloud provider specific config)
+  annotations: {}
+  # Additional entries here will be added to the service spec. Cannot contains
+  # type, selector or ports entries.
+  spec: {}
```
elthariel: Not entirely sure of that one. We might want to just select the configurable fields ?

values.yaml
```diff
@@ -2,31 +2,81 @@
...
+  # Number of pods of the deployment
   replicas: 1
+
+rollingUpdate:
+  maxUnavailable: 1
+  maxSurge: 1
+
+# Additional arguments to be passed at the startup of traeffik
+additionalArguments: []
+#  - "--argument1"
```
elthariel: I'd gladly get a better idea of example ^-^

elthariel:
Might fix #7, #14, #21 or at least provide a workaround for them

dduportal:
Hi @elthariel , thanks for this PR, it is full of smart changes that could really help the end users!

May we ask you to split the PR to have atomic changes, as it is really a large one?

Let's say update the current PR to focus on one of the following "domains", and open 1 PR for each other domain, so the changes will be easier to track.

Domains concerned by this PR:

- Allow defining Entrypoints through values (including Traefik's CLI arguments `--entrypoints.*` and the Kubernetes Service exposing Traefik)
- Allow passing additional CLI argument to Traefik (which fixes issues related to other providers or custom Traefik configuration as #7 #14 and #21)
- Define a value section for "dashboard" with enable/disable possibility (+ tuning)  
- Style changes (default values, indent, if/with etc.)
- Passing custom annotations 

Thanks a lot for this GREAT GREAT contribution. Any help you need, please ask!

values.yaml
dduportal: In the spirit of #21, WDYT about 

```yml
#  - --provider.file.filename=/etc/traefik.dynamic.yml
```

?

values.yaml
dduportal: Good question. WDYT @dtomcej , any advise?

Pro: we let the user able to tune for whatever edge case and don't have to map 1000s of values
Cons: we might have to manage overriding (for instance if end user specifies `type:Cluster IP` here while letting the default `type: Loadbalancer`: key will be duplicated)

values.yaml
dduportal: good point!

values.yaml
dtomcej: We have a yaml section in the values.yaml that is applied to the deploment spec.

We may want to move the raw yaml section to each component:

Deployment:
  yaml:

Service:
  yaml:

This would allow users to add custom sections of the spec themselves without having to modify the base chart.

dtomcej: although I think that would be best in its own PR

values.yaml
elthariel: That would imply we provide a way to mount this as a configmap somewhere which I'm not doing

elthariel:
Splitting into 5 PRs is more work than the time it took to write the changes 😭 
I'll see what I can do when I've a moment. I haven't done this kind of git splitting work for a few years now haha

dduportal:
@elthariel if you create the 4 other PRs with one line change on each, I can help you by pushing the changes on the 5 if it is ok for you (because we'll be co-author of the PRs after this as I'll puhs commits of your works)
WDYT?

elthariel:
I was just whining a bit, mostly because it is friday. I'll take care of this don't worry

dduportal:
Ok, thanks, you rocks! Anything that can help: shout!

values.yaml
elthariel: I think `yaml` is a bit confusing. It seems to me that it is more standard to see the name of the section where it'll be merged like `spec`

elthariel:
There you all, all split up.

They won't merge nicely, we'll need to rebase some of them. Keep me posted

dduportal:
Thanks a lot @elthariel ! Can you update this PR to focus only on the "service configuration" (annotation, etc.) please? (ref. https://github.com/containous/traefik-helm-chart/pull/33#issuecomment-557812208) 

If you are busy, just drop a comment here and i'll update this PR and #33 for doing this.

Thanks a lot!

elthariel:
Yeah, feel free to do it

dduportal:
Thanks @ lot for your contribution @elthariel !

We're now closing this PR in favor of the other you created: #30 , #31 , #32 , #33 , #34 , #36 .

PR #27, opened 2019-11-19
Use a secret. K8s encryption at rest

ekjuanrejon:


ekjuanrejon:
@dduportal 

Thanks for the feedback. Am a bit puzzled about using environment variables. If it is not recommended then I think I would be forced to use another edge router.

My company is going through some pen test and security auditors have recommended to use secrets and load into the pod. That's the main reason I have opted to use the environment variables... I'll have to rethink Traefik 2

Am okay fixing the other issues.

dduportal:
Hi @ekjuanrejon , do you mind starting a topic on our community forum at http://community.containo.us/ and mention me?

Traefik treats Kubernetes secrets as first-class citizen and can totally work by following Kubernetes security practises with it:

- Secrets can be used for holding certificates for the IngressRoutes objects
- Secrets can be mounted inside Traefik's pod and used with the file provider to be available without copying to other namespaces

Describing your use case in the community forum could help a lot: you can still use environment variables to provide some secrets as the `AZURE*` variables from your PR, but these variables should not be here in the official helm chart as not everyone is using it. Also, Traefik's CLI based configuration is not in charge of passing any secret at all: it means that you can totally keep the CLI based configuration and still  define environment variables.

But you have to explain what is your need step by step, this is why the community forum is recommended :)

ekjuanrejon:
@dduportal 
am look at the values.yaml and actuall
`env:` is an array can be modified by the user. It just happen that I added the environment variables that work for me

dduportal:
Hi @ekjuanrejon  can you switch to the community forum please?

PR #30, opened 2019-11-23
Do not provide defaults for resources req/lim

elthariel:
Chunk 1 of #26


dduportal:
For later readers: this PR applies a good practise around Helm charts:

From https://github.com/helm/charts/blob/master/REVIEW_GUIDELINES.md#kubernetes-native-workloads:

> It is generally not advisable to provide hard resource limits to workloads and leave it configurable unless the workload requires such quantity bare minimum to function.

PR #31, opened 2019-11-23
Add an option to disable dashboard

elthariel:
Part 2 of #26

### DIFF
--- a/templates/dashboard-hook-ingressroute.yaml
+++ b/templates/dashboard-hook-ingressroute.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.dashboard.ingress.enabled }}
 apiVersion: traefik.containo.us/v1alpha1
 kind: IngressRoute
 metadata:
--- a/templates/dashboard-hook-ingressroute.yaml
+++ b/templates/dashboard-hook-ingressroute.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.dashboard.ingress.enabled }}
+{{- if .Values.dashboard.ingressRoute }}
 apiVersion: traefik.containo.us/v1alpha1
 kind: IngressRoute
 metadata:
--- a/templates/dashboard-hook-ingressroute.yaml
+++ b/templates/dashboard-hook-ingressroute.yaml
@@ -19,3 +19,4 @@ spec:
     services:
     - name: api@internal
       kind: TraefikService
+{{- end -}}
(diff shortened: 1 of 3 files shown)


ldez:
This PR need to be rebase because of  #35

elthariel:
#35 breaks my setup, so not really motivated haha

elthariel:
On a side note, I'm wondering why the ingress is installed with a hook ? iirc, it means the resource isn't deleted when removing the release and prevent re-installing it

dduportal:
If you want @elthariel , I can take care of this: I thought your PR would be merged before mine (sorry for that) 😅 

> On a side note, I'm wondering why the ingress is installed with a hook ? 
If you install on a scratch cluster, as the dashboard requires an `IngressRoute` object to be created, it means that there is a precedence with the CRDS that must be installed first. So we need a hook to be sure that Helm create the `IngressRoute` only when the Kube API is ready for it.

> iirc, it means the resource isn't deleted when removing the release and prevent re-installing it

Good point. We might want to add a hook delete policy as in https://v3.helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases .

dduportal:
@elthariel thanks a LOT for your work and feedback. I took the freedom to rebase this PR against master, and remove the `dashboard.service` part as it is not needed anymore. WDYT?

dduportal:
Latest test:

- Default install let me access the dashboard 👍 (ingress created as before)
- Install with `helm install --name=traefik --set=dashboard.ingress.enabled=false .` works (Traefik is functional), but no Ingress exposed 👍 

I'm not LGTM-ing this PR as I co-contributed :)

ldez:
seems a bit confusing with `- "--api.dashboard=true"`

dduportal:
@ldez I'm not sure to understand? Are you saying we should keep the same naming in the `values.yaml` as in Traefik alone?

ldez:
- `.Values.dashboard.ingress.enabled` add a IngressRoute, so the word `ingress` seems weird.
- `- "--api.dashboard=true"` is not configurable.

dduportal:
> Values.dashboard.ingress.enabled add a IngressRoute, so the word ingress seems weird.

Good catch, thanks!

> - "--api.dashboard=true" is not configurable.
> If I follow the PR title ("Add an option to disable dashboard") there is something missing.

Fair. WDYt about the following values?

```yml
dashboard:
  # Enable the dashboard on Traefik
  enable: true

  # Expose the dashboard and api through an ingress route at /dashboard
  # and /api This is not secure and SHOULD NOT be enabled on production
  # deployments
  ingressroute: true
```

ldez:
I agree :+1: 

`ingressroute` -> `ingressRoute`

PR #34, opened 2019-11-23
Add support for passing additional arguments

### LINKED ISSUE #7: [Feature Request] Enable --providers.kubernetesingress

elthariel:
Part 5 of #26 

Closes #7,
helps for #14 and #21

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -17,7 +17,7 @@ rollingUpdate:
 #
 # Configure Traefik entry points
 # Additional arguments to be passed at Traefik's binary
-## Use curly braces to pass values: `helm install --set="{--providers.kubernetesingress,--log.level=DEBUG}" ."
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -5,6 +5,11 @@ image:
 #
 deployment:
   replicas: 1
+
+# Additional arguments to be passed at Traefik's binary
+additionalArguments: []
```
dduportal:
```suggestion
## Use curly braces to pass values: `helm install --set="{--providers.kubernetesingress,--log.level=DEBUG}" ."
additionalArguments: []
```

elthariel: That's a good point. I never use --set={}, I always copy the values.yml file and edit it

elthariel: I wasn't aware of the commit suggestion feature of GH, that's nice.

PR #35, opened 2019-11-23
Bump Traefik v2.1.0-rc2, allowing to use native internal dashboard

dduportal:

- Traefik version bumped to 2.1.0-rc2 (ref. https://github.com/containous/traefik/releases/tag/v2.1.0-rc2)
- Add the new `TraefikService` CRD during the installation, to allow defining WRR and Mirror traefik services
- Switch dashboard to the ["secure mode"](https://docs.traefik.io/v2.1/operations/dashboard/#secure-mode) which implies removing the kubernetes service named `traefik-dashboard`


Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: traefik
-version: 2.0.1
-appVersion: 2.0.5
+version: 2.1.0
```
elthariel: This is something that could be configured in the `values.yml` as well. Allowing to upgrade the chart without upgrading traefik ?

elthariel: This comment is especially relevant provided the version below is a release candidate :)

dduportal:
The tag of the image is also defined in values.yaml. I am not sure to understand what you mean?

elthariel:
Sorry sorry @dduportal, I haven't read the code with enough attention. Please ignore the comments
#badpuppy

elthariel:
On a different level, I think we should allow the user to chose between the internal 'secure' service and the old way, which is really helpful when debugging

dduportal:
> On a different level, I think we should allow the user to chose between the internal 'secure' service and the old way, which is really helpful when debugging.

@elthariel starting with 2.1, we want the "secure" mode to be the default. Not sure if it is worth it to add more code for the "insecure" mode, since your PR #34 will enable users to tune such thing. Also, the Kubernetes service used to expose the dashboard, which was removed in the PR here, has to be managed by the end user, and it's easier to open possibilities by letting user doing this in such a edge case, my 2 cents.

rbtr:
a consolidated example of how to get to the dashboard as it's deployed now in "secure" mode would be nice - trying to piece it together from the scattered docs is challenging

elthariel:
@dduportal I think it's great to have secure being the default, but this PR is doing more than making it the default, since it removes the service ?

I've found it useful to be able to enable it temporarily to debug stuff, or because you're using another security mechanism (kubectl proxy/port-forward), or for dev/tmp clusters

It looks to me that after this PR, people would have to fork/edit the chart or add external resources to be able to have a k8s service pointing to the dashboard, while requiring additional arguments to be passed, leading to some resources being managed by charts and some not.

WDYT ?

PR #38, opened 2019-11-24
add secret for cloud provider variables when using acme

ekjuanrejon:


dduportal:
Hi @ekjuanrejon , thanks for this contribution!

Would you mind explaining the high level use case please?
I understand that you want to pass secrets as environment variables to Traefik: if this is correct,
WDYT providing the secret references to Traefik, under a `values.yaml` section named `secrets:` that would provide reference and a 2nd directive `exposeToEnv`:

```
secrets:
  - reference: a-secret-used-as-env
     exposeToEnv: true
  - reference: a-secret-not-exposed-to-env
     exposeToEnv: false
```

?
The reason is simple (if I understand correctly the use case): with Helm v2, the values of an installed chart are stored in a ConfigMap, which means that the value that you want in a secret are stored as plain text, available in the API, which you might not want :)

Instead, if you keep, as end user, responsibilities to create the secrets yourself so it's always secret, and only ask Traefik to mount these secrets, and eventually expose its, then it should solve the "end to end" secret issue, what do you think?

chrissng:
@dduportal I think what @ekjuanrejon wants to do is to make this chart work well with certificates generated and managed by cert-manager. I presume this cert is a wildcard cert of sorts.

To make this PR's feature useful, it should simple:
- allow specifying the secret name via values.yaml
- have the chart mount this secret to be used as a default cert.

Basster:
I like the idea from @dduportal, but I don't see a requirement for the `exposeToEnv` part. Is there any use-case where the referenced secret(s) should **not** be exposed to env?

My proposal:

```yaml
# values.yaml
deployment:
  # Number of pods of the deployment
  replicas: 1
  envFromSecret: <secret-name> # nil by default
```

mmatur:
Hi @ekjuanrejon,

Thanks for your contribution. 

I will close your issue because the support of secret in environment variable has been added by #101

PR #39, opened 2019-11-25
add more flexibility to dashboard ingress

ekjuanrejon:


dduportal:
Hi, thanks for this proposal! As we want to keep this helm chart as lean as possible, we thought that customizing the dashboard’s ingress route very limited, and asking end user to provide their own IngressRoute instead as soon as the « simple » reach a limit.

It looks like this is your case: you want to adapt the ingressroute to your context.

Is it ok for you if installation of the helm chart is done with dashboard enabled but dashboard.ingressroute is false, and then you can create your own ingressroute to fit your need?

The question I raise here is for discussion : the risk I see is a drift between the functionalities exposed through values.yaml, and the ingressroute features, while the proposal of « simple » let end user the freedom of whatever customization they consider necessary for them, what do you think ?

rbtr:
@dduportal not the OP, but I think the dashboard IngressRoute should have more config options and safer default values. It should at least *also* have an overridable `Host` matcher. 

I've just run in to a scenario where I had another service that's serving at `.../api`, and it has a separate IngressRoute with a unique `Host` matcher, but the Traefik dashboard has been stealing that route since the Chart does not specify a host. This has caused me some headaches until I realized this.

IMO it is better to be clear and explicit, than necessarily simple. I think the default routing should default to being too specific (like setting a default host & path matching), rather than too broad.

ldez:
The default API/dashboard IngressRoute is mainly for dev/testing.

For production, we recommend to create a more secure IngressRoute with some middlewares, TLS configuration, entrypoints, ...

rbtr:
of course, before taking this to production everything should be customized and hardened. but my example was in a dev/testing cluster (for internal-facing, vpc-private services) where i was trying to integrate traefik as a POC, and struggling with this unintuitive behavior was painful. 

hopefully that one very specific use case highlights that being a little more explicit and customizable instead of simple for the sake of being simple can result in a better experience for everyone :)

PR #45, opened 2019-12-13
Introducing a "Build workflow"

dduportal:
The goal of this PR is to introduce a "build workflow" for this Helm chart.

It provides a standard workflow, that can be run by Travis (tested here: https://travis-ci.com/dduportal/traefik-helm-chart/branches) for ensuring that any contribution on this chart in the future will be valid.

This workflow is (for now) the following

- A "lint" phase, to ensure that there is no obvious error on the charts.
  * Implementation is only a "helm lint" for today
- A "build" phase which generates an artefact of type `tgz archive` for the helm on a local directory.
- A "deploy" phase which copy the generated artefact in a dummy directory (git ignored for today: it is a placeholder) to simulate a future "helm repository" (e.g. a directory containing an `index.yaml`, a list of tgz, served by a public web server in https as github pages, netlify or s3 bucket).

Of course, there are big many steps that are not treated here to avoid a big dangling PR. The proposal is to go tiny steps by tiny steps.

Some thoughs on the next steps:

- Adding a test phase (at least a smoke test which ensures that the `helm install ./` command without any custom value is ALWAYS working)
- Use Github Pages with a `gh-pages` branch , where the helm repo would be served and deploy. 
- Adding more linting rules (my most loved one: if it is a PR, then check that the Chart version was bumped or fail the build immediately...)
- Generates and distribute the "YAML" manifest from the helm chart, by using `hem template <...>`. This would allow distributing Traefik in Kubernetes in another way, without risking a shift between hel mchart and "bare YAML". It would open the door for people who want to use Kustomize instead of Helm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 
 # Outputs directories
 dist/
+repo/
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+
+# Outputs directories
+dist/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+
+notifications:
+  email:
+    on_success: never
+    on_failure: change
+
+sudo: required
+dist: xenial
+
+services:
+  - docker
+  
+script:
+  - make
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ notifications:
 sudo: required
 dist: xenial
 
+install:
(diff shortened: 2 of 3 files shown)


amalucelli:
@dduportal I was just testing this chart and I was ready to fork the repository to serve it internally for testing when I saw this PR. The idea is to officially serve from this repository the v2 (_I remember in some thread that later this would be moved to the `stable` repository_)? If you need any kind of help I'm glad to help!

fwiw: for validating if the chart was bumped, I usually use this little trick during the build: `git diff --quiet HEAD origin/master || git diff HEAD origin/master -- Chart.yaml | grep -qE "version:.*"`

PR #49, opened 2019-12-20
expose hostnetwork option

### LINKED ISSUE #99: Expose hostNetwork customization
Opening issue for discussion as requested in https://github.com/containous/traefik-helm-chart/pull/49.

I want to be able to run Traefik in the host network namespace when I deploy it with this chart for the following potential scenarios:
- I don't want to use a cloud provider load balancer in front of Traefik due to cost or redundancy.
- I want to run Traefik on bare metal and don't have the option of fronting it with another load balancer.
- I need my Traefik pods to have the Node IP as the Pod IP (to use with, for example, https://github.com/kubernetes-sigs/external-dns which will automatically create a DNS record pointing to that IP)

For any or all of these use cases, I need to be able to run my Traefik pod in the host network because:
- binding to port 80/443 is restricted (by default k8s only allows NodePort > 3000)

As noted in the Traefik Team reply [here](https://github.com/containous/traefik-helm-chart/pull/93#issuecomment-596983077) and [included in my original PR](https://github.com/containous/traefik-helm-chart/pull/49/files#diff-a0b6b6f5d77ba10958eabde5b19477f9R30), this will also require some anti-affinity rules to be added to force Traefik pods to spread out among the nodes.

@mmatur do you have any questions about this?

rbtr:
What:
adds hostNetwork and an example anti-affinity for hostnetwork compatibility.

Why:
I want traefik to bind to the node ports 80/443, but a NodePort service can only bind to ports >3000 (unless k8s is specifically configured to allow a different NodePort range which is not recommended). The only option in this situation is to run the pod in the host network namespace...       
But if the pod is running in the host network, we can't schedule more than one per node, because the port will become in-use as soon as the first pod starts, so it is necessary to spread them out among the nodes when we have more than one pod replica.

Closes #99

### DIFF
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -46,6 +46,7 @@ spec:
     spec:
       serviceAccountName: {{ template "traefik.fullname" . }}
       terminationGracePeriodSeconds: 60
+      hostNetwork: {{ .Values.hostNetwork }}
       containers:
       - image: {{ .Values.image.name }}:{{ .Values.image.tag }}
         name: {{ template "traefik.fullname" . }}
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: traefik
-version: 5.4.0
+version: 5.5.0
 appVersion: 2.1.3
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -123,6 +123,12 @@ persistence:
   path: /data
   annotations: {}
 
+# If hostNetwork is true, runs traefik in the host network namespace
+# To prevent unschedulabel pods due to port collisions, if hostNetwork=true
+# and replicas>1, a pod anti-affinity is recommended and will be set if the
+# affinity is left as default.
+hostNetwork: false
+
 resources: {}
   # requests:
   #   cpu: "100m"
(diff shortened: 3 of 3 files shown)


mmatur:
Hi @rbtr 

Thanks for your contribution.

- Affinity has been introduced by https://github.com/containous/traefik-helm-chart/pull/57
- Env has been introduced by https://github.com/containous/traefik-helm-chart/pull/71

Regarding the hostNetwork, what is your use case?
Why you need to bind traefik to the node ports 80/443?

I am closing this PR because affinity and env have already been implemented. 

Could you please open an issue to start a discussion about the hostNetwork. We need to understand the use case behind the needs to bind traefik to node ports 80/443. 

Thanks

mmatur:
@rbtr After a discussion with the Traefik team we have decided to reopen this PR.

mmatur:
@rbtr Could you please rebase your PR

traefik/templates/deployment.yaml
```diff
@@ -27,8 +27,26 @@ spec:
...
+                operator: In
+                values:
+                - {{ template "traefik.name" . }}
+            topologyKey: failure-domain.beta.kubernetes.io/zone
+      {{- else }}
+      {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- end }}
```
mmatur:
```suggestion
      {{- with .Values.affinity }}
      affinity:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

mmatur: This condition can be replace by a comment and an example in the `values.yaml`

rbtr:
@mmatur rebased and cleaned up per review comments!

PR #50, opened 2020-01-02
Enhance security by add loadBalancerSourceRanges to lockdown ip address.

terrych0u:
Purpose:
When the type is LoadBalancer on the cloud ex: AWS, the Load Balancer ELB/NLB default will allow 0.0.0.0/0 for the public internet.

Add loadBalancerSourceRanges function in service.yaml to help lockdown IP address if needed.


sshishov:
We would like to see this in the chart. Currently, we have to not forget to apply CIDR manually when LB is created, otherwise, it will be huge security breach.

terrych0u:
I added unit tests and update PR to the new folder structure.
if there still some things I'm not done properly, please let me know
thank you for review.

PR #51, opened 2020-01-02
Disable dashboard ingressRoute by default

davidalger:
Settings explicitly not secure for production deployment should never be enabled by default. Having this enabled by default WILL result in production deployments of Traefik which have this dashboard IngressRoute enabled. Please consider accepting this change and having the dashboard ingress route require a value be set by the user of the chart.


davidalger:
@dduportal Good call on that. I've updated to bump version to 3.0.0.

PR #54, opened 2020-01-03
Allow kind override (daemonset)

Eraac:
Very basic PR here, just allowing to set affinity settings (like nodeSelector) and allow to override `kind: Deployment` (for use DaemonSet if you want)

Not sure if I need to update the version in Chart.yaml (from 2.2.2 to 2.2.3?)

Thanks!

### DIFF
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -3,7 +3,7 @@ apiVersion: apps/v1
 {{- else }}
 apiVersion: apps/v1beta1
 {{- end }}
-kind: Deployment
+kind: {{ .Values.deployment.kind }}
 metadata:
   name: {{ template "traefik.fullname" . }}
   labels:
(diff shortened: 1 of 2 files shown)


Eraac:
Job done, I've only update chart version in the second PR (from 2.2.2 to 2.3.0)

dduportal:
I'm adding another comment: I think this PR must be changed, with 2 different template files for Deployment and DaemonSet (check the nginx-ingress as a close example - https://github.com/helm/charts/tree/master/stable/nginx-ingress).

The reason is that the APIs for Deployment and DaemonSet are not compliant with each others: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17.


But this PR is still really useful term of intent: it is important to provide a way to the end user to deploy as daemonset instead of deployment!

WDYT @Eraac ?

PR #55, opened 2020-01-03
Allow affinity settings

Eraac:
PR contains also have the commits from https://github.com/containous/traefik-helm-chart/pull/54 but should be gone once is merge

### DIFF
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -3,7 +3,7 @@ apiVersion: apps/v1
 {{- else }}
 apiVersion: apps/v1beta1
 {{- end }}
-kind: Deployment
+kind: {{ .Values.deployment.kind }}
 metadata:
   name: {{ template "traefik.fullname" . }}
   labels:
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -79,6 +79,10 @@ spec:
           - {{ . | quote }}
           {{- end }}
           {{- end }}
+      {{- with .Values.affinity }}
+      affinity:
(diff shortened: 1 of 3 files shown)


templates/deployment.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apps/v1
 {{- else }}
 apiVersion: apps/v1beta1
 {{- end }}
-kind: Deployment
+kind: {{ .Values.deployment.kind }}
```
dduportal:
```suggestion
kind: Deployment
```

values.yaml
```diff
@@ -7,6 +7,8 @@ image:
 # Configure the deployment
 #
 deployment:
+  # What kind of object you want to create
```
dduportal:
```suggestion
```

templates/deployment.yaml
dduportal: This change is part of another PR (#54 ) and should not be here.

values.yaml
dduportal: This change is part of another PR (#54 ) and should not be here.

values.yaml
```diff
@@ -7,6 +7,8 @@ image:
...
+  kind: Deployment
```
dduportal:
```suggestion
```

dduportal: This change is part of another PR (#54 ) and should not be here.

dduportal:
Hi @Eraac , why did you close the PR? It's a good contribution that should be integrated :)

PR #56, opened 2020-01-03
allow kind override (daemonset)

Eraac:
https://github.com/containous/traefik-helm-chart/pull/54

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+
+# Outputs directories
+dist/
+repo/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+dist: xenial
+sudo: required
+
+notifications:
+  email:
+    on_success: never
+    on_failure: change
+
+services:
+  - docker
+
+env:
+  global:
+    - TMPDIR=/tmp
+
+install:
+  - curl -L https://git.io/get_helm.sh | bash -s -- -v v2.16.1
+
+script:
+  # validating if the chart was bumped
+  # - git diff --quiet HEAD origin/master || git diff HEAD origin/master -- Chart.yaml | grep -qE "version:.*"
+  - make
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 2.2.2
+version: 2.3.0
 appVersion: 2.1.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 2.3.0
+version: 3.1.0
 appVersion: 2.1.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 2.2.1
+version: 2.2.2
 appVersion: 2.1.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,59 @@
+
+DIST_DIR ?= $(CURDIR)/dist
+TMPDIR ?= /tmp
+SHIM_DIR ?= $(TMPDIR)/traefik
+HELM_REPO ?= $(CURDIR)/repo
+
+################################## Functionnal targets
+
+# Default Target: run all
+all: clean lint build deploy
+
+# Ensure the Helm chart and its metadata are valid
+lint: helm $(SHIM_DIR)
+	@echo "== Linting Chart..."
+	@helm lint $(SHIM_DIR)
+	@echo "== Linting Finished"
+
+# Generates an artefact containing the Helm Chart in the distribution directory
+build: helm $(DIST_DIR) $(SHIM_DIR)
+	@echo "== Building Chart..."
+	@helm package $(SHIM_DIR) --destination=$(DIST_DIR)
+	@echo "== Building Finished"
+
+# Prepare the Helm repository with the latest packaged charts
+deploy: build $(DIST_DIR) $(HELM_REPO)
+	@echo "== Deploying Chart..."
+	@cp $(DIST_DIR)/*tgz $(HELM_REPO)/
+	@helm repo index $(HELM_REPO)
+	@echo "== Deploying Finished"
+
+# Cleanup leftovers and distribution dir
+clean:
+	@echo "== Cleaning..."
+	@rm -rf $(DIST_DIR)
+	@unlink $(SHIM_DIR) >/dev/null 2>&1 || true
+	@echo "== Cleaning Finished"
+	
+################################## Technical targets
+
+$(DIST_DIR):
+	@mkdir -p $(DIST_DIR)
+
+
+## This directory is git-ignored for now, 
+## and should become a worktree on the branch gh-pages in the future
+$(HELM_REPO):
+	@mkdir -p $(HELM_REPO)
+
+helm:
+	@command -v helm >/dev/null || ( echo "ERROR: Helm binary not found. Exiting." && exit 1)
+	@helm init --client-only
+
+# This target is phony to ensure there is no conflict with other dir named "traefik"
+$(SHIM_DIR):
+	@## Helm v2 require directory to have the same name as the chart
+	@[ -L $(SHIM_DIR) ] && unlink $(SHIM_DIR) || true
+	@ln -s $(CURDIR) $(SHIM_DIR) || ( echo "ERROR: cannot link $(CURDIR) to $(SHIM_DIR). Exiting." && exit 1)
+
+.PHONY: all helm lint build deploy clean $(SHIM_DIR)
--- a/templates/_pod-template.tpl
+++ b/templates/_pod-template.tpl
@@ -75,6 +75,10 @@
           - {{ . | quote }}
           {{- end }}
           {{- end }}
+      {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- with .Values.tolerations }}
       tolerations:
         {{- toYaml . | nindent 8 }}
--- a/templates/_pod-template.tpl
+++ b/templates/_pod-template.tpl
@@ -1,0 +1,86 @@
+{{- define "metadata.labels" }}
+    app: {{ template "traefik.name" . }}
+    chart: {{ template "traefik.chart" . }}
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
+{{- end }}
+
+{{- define "spec.template.metadata.labels" }}
+        app: {{ template "traefik.name" . }}
+        chart: {{ template "traefik.chart" . }}
+        release: {{ .Release.Name | quote }}
+        heritage: {{ .Release.Service | quote }}
+{{- end }}
+
+{{- define "spec.selector" }}
+    matchLabels:
+      app: {{ template "traefik.name" . }}
+      release: {{ .Release.Name }}
+{{- end }}
+
+{{- define "spec.strategy" }}
+    type: RollingUpdate
+    rollingUpdate:
+    {{- with .Values.rollingUpdate }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+{{- end }}
+
+{{- define "spec.template.spec" }}
+      serviceAccountName: {{ template "traefik.fullname" . }}
+      terminationGracePeriodSeconds: 60
+      containers:
+      - image: {{ .Values.image.name }}:{{ .Values.image.tag }}
+        name: {{ template "traefik.fullname" . }}
+        {{- with .Values.resources }}
+        resources:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
+        readinessProbe:
+          httpGet:
+            path: /ping
+            port: {{ .Values.ports.traefik.port }}
+          failureThreshold: 1
+          initialDelaySeconds: 10
+          periodSeconds: 10
+          successThreshold: 1
+          timeoutSeconds: 2
+        livenessProbe:
+          httpGet:
+            path: /ping
+            port: {{ .Values.ports.traefik.port }}
+          failureThreshold: 3
+          initialDelaySeconds: 10
+          periodSeconds: 10
+          successThreshold: 1
+          timeoutSeconds: 2
+        ports:
+        {{- range $name, $config := .Values.ports }}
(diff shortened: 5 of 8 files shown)


Eraac:
This option exist (at least in the documentation) for 1.7 https://docs.traefik.io/v1.7/user-guide/kubernetes/#deploy-traefik-using-a-deployment-or-daemonset

And we've already experiment issue in EKS when autoscaling decide to remove old machine, the service can be down/degraded in particular case if pod are not well spread in the cluster

dduportal:
> This option exist (at least in the documentation) for 1.7 https://docs.traefik.io/v1.7/user-guide/kubernetes/#deploy-traefik-using-a-deployment-or-daemonset

Yes, you are totally right on this and we agree: the functionality "allow Traefik to be deployed as a DaemonSet" should be provided to end user.

But my concern is about the implementation: as the Deployment and DaemonSet APIs are different, then the YAML files for creating objects through the `kubectl apply` command are different. Changing "only" the spec attribute "kind" won't work as soon as the end user provides an attributes specific to `Deployment` or `DaemonSet`.

This is why my proposal to you is to update the PR, to provide 2 differents template files: 1 for Deployment and one for Daemonset.

You can keep the values' attribute `kind`, but move it on top level, and add a comment to explain the possible values it can take.

Then this attribute can be used to enable/disable one of the 2 YAML files.

Eraac:
> But my concern is about the implementation: as the Deployment and DaemonSet APIs are different

Indeed, I've create new file specific for DaemonSet and replace `.Values.deployment.kind` with `.Values.useDaemonSet` (by default false)

That's a lot of duplication, and I'm afraid someone update deployment but forget daemonset (maybe we can add a comment in top of each file?)

davidalger:
> And we've already experiment issue in EKS when autoscaling decide to remove old machine, the service can be down/degraded in particular case if pod are not well spread in the cluster

Wouldn't this be better addressed by [specifying a PodDisruptionBudget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) to prevent unavailability due to an evicted pod? Since Traefik is stateless, this really seems like the better and more recommended solution to resolving availability concerns.

That said, I certainly see value in supporting a DaemonSet as there are other possible advantages such as DaemonSets automatically replicating to new nodes when they join the cluster or for example ensuring only one replica of a pod runs on any single node (although this is I believe possible with affinity settings as well).

Eraac:
> although this is I believe possible with affinity settings as well

Yes, that why initially I commit both in the same PR

> Wouldn't this be better addressed by specifying a PodDisruptionBudget 

I can make a PR for this too

dduportal:
Hello, about the code duplication between Deployment and DaemonSet,
Helm provides a feature of "partials files": https://helm.sh/docs/chart_template_guide/named_templates/#partials-and-files .

It could be used to hold the "common definition" (e.g. the Pod template), and then could be included in both template manifests, WDYT?


About adding the PodDistruption budget, unless it's crucial for you in short time, could you wait a little bit? I'm going to propose something to avoid requiring bazillions of values for tuning each attribute, where the idea is to pass "raw YAML" as a value to customize based on your contextual needs.

davidalger:
@dduportal Personally I would prefer seeing configuration to enable resources like `PodDisruptionBudget` exist as supported flags in the values file along with ability to customize as needed.

I could see this being as simple as this in the values file:
```
podDisruptionBudget:
  enabled: false
  minAvailable
  maxUnavailable: 1
```

The idea of passing raw yaml works for things like specifying resource requests and limits, simple things, but the deeper the yaml becomes the more difficult it is to specify said yaml in context of a `--set` argument, which is a common pattern when deploying helm charts via Terraform. For example, the following is what I have setup in Terraform to deploy this chart into a test cluster I'm working on getting built to run demo sites:

```

resource "helm_release" "traefik" {
  name       = "traefik"
  namespace  = "kube-system"
  chart      = "${path.module}/charts/traefik"
  depends_on = [module.eks]

  ## This is required for Traefik access logs to record correct IPs (when logging is enabled) and for Magento
  ## to respond correctly to an https request (without this, infinite redirection will occur)
  set {
    name  = "additionalArguments"
    value = "{${join(",", [
      "--entryPoints.web.forwardedHeaders.trustedIPs=${join(",",
        data.aws_vpc.vpc.cidr_block_associations.*.cidr_block
      )}"
    ])}}"
  }

  set {
    name  = "service.annotations.magento-demo\\.base-domain"
    value = var.base_domain
  }

  set {
    name  = "service.annotations.service\\.beta\\.kubernetes\\.io/aws-load-balancer-ssl-cert"
    value = var.certificate_arn
  }

  set {
    name  = "service.annotations.service\\.beta\\.kubernetes\\.io/aws-load-balancer-backend-protocol"
    value = "http"
  }

  set {
    name  = "ports.web.exposedPort"
    value = 443
  }

  ## Traefik is not handling SSL termination
  set {
    name  = "ports.websecure.expose"
    value = false
  }

  set {
    name  = "resources.requests.cpu"
    value = "100m"
  }

  set {
    name  = "resources.requests.memory"
    value = "20Mi"
  }

  set {
    name  = "resources.limits.cpu"
    value = "300m"
  }

  set {
    name  = "resources.limits.memory"
    value = "60Mi"
  }
}
```

Eraac:
@dduportal I've implement template file, not very satisfied because you can't `{{- template "..." . | nindent X }}`, so you should put tabs into the define block, it's a bit confusing.

I let you decide what's the best for this helm chart

dtomcej:
The issue I have with allowing this configuration is that it implies that Traefik supports running in Daemonset mode with the full configurations. Note that LetsEncrypt is not supported for more than one pod, so it could be misleading.

The deployment defaults to 1 replica, which has the full feature support. This was one of the main reasons to use the deployment in the first place.

Please see:
https://github.com/helm/charts/pull/11813#issuecomment-530001942

for further discussion.

Eraac:
> The deployment defaults to 1 replica, which has the full feature support. This was one of the main reasons to use the deployment in the first place.

Yes of course, but I guess no all users of Traefik use Lets Encrypt and set 1 to replica isn't for production.

mmatur:
Hi @Eraac ,

Thanks for your contribution.

We have been discussing a lot within the Traefik team and we have decided to close this PR for reasons explain [here](https://github.com/helm/charts/pull/11813#issuecomment-530001942)

> Hello All,
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> Thanks,
> 
> The Traefik Team.

ShakataGaNai:
This is extremely disappointing to see the Traefik team refusing to support a valid deployment strategy that was used previously to great effect. If you're going make the helm chart highly opinionated then it should only be provided along with the standard YAML as part of the install documentation https://docs.traefik.io/getting-started/install-traefik/.

In my use case I ran Traefik 1.x in DaemonSet because, gasp, there are some people who DON'T run in the cloud. I don't have load balancers in front of my servers. Instead I provide load balancing via DNS RR. I need those hosts to all be running Traefik with 100% guarantee. The *correct* way to assure one and exactly one copy of a pod is running on every host is a DaemonSet. This is not my opinion, it's in the official docs. Anti-Affinity hacks are not the recommended solution, DaemonSet is. Oh and since I'm aware I can't have LetsEncrypt on multiple nodes, I run CertManager. As it was previously recommended. It's not rocket science.

It used to be easier to get all this running in Traefik 1.x, now 2.x has obscured setups behind helm charts and made it a huge effort in order to even figure out what the correct "default" configuration is, let alone customize it.

mamiu:
@mmatur @dtomcej 

We have exactly the same setup as described in the previous comment (by @ShakataGaNai) except that we're using traefik for automatic certificate generation (with Let's Encrypt).

[You wrote](https://github.com/traefik/traefik-helm-chart/pull/56#issuecomment-596983096) that:

> Traefik v2 will not support LetsEncrypt in multi-node mode, so the option of a daemonset will not work if users want to use LE.

Why won't Traefik support *multi-node mode*? Our cluster supports [ReadWriteMany](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) (the volume can be mounted as read-write by many nodes) and since Traefik doesn't read/write much to the Volume it shouldn't be a bottleneck.

And Daniel [mentioned](https://github.com/traefik/traefik-helm-chart/pull/56#issuecomment-571225078) that:

> The issue I have with allowing this configuration is that it implies that Traefik supports running in Daemonset mode with the full configurations.

Is that really an issue? A small note in the documentation or a comment in the `values.yaml` file of the helm chart would be absolutely sufficient to indicate this.

There number of Github issues (#64, [#3190](https://github.com/helm/charts/issues/3190)) and PRs (#141, #93, [11813](https://github.com/helm/charts/pull/11813), [19306](https://github.com/helm/charts/pull/19306), [12335](https://github.com/helm/charts/pull/12335), [9797](https://github.com/helm/charts/pull/9797)) concerning this topic indicates the interest of many users for this feature. And nobody has asked to make this configuration the default.

Or is there another configuration you recommend in multi-node on-prem clusters which do load balancing via DNS RR?
Anti-affinity is no option for us because of [this behavior](https://github.com/traefik/traefik-helm-chart/pull/141#issuecomment-610229457).

SantoDE:
Hello @mamiu,

the problem is not only about mounting a volume, but also about who is answering the challenge in case of TLS or HTTP.

As only the traefik instance who initiated the challenge knows the proper answer, dispatching to other traefik nodes through a round robin algorithm will cause the challenge to fail.

As this is a bit hidden or might not be obvious to understand immediately, we will not support a deamon set like deployment by default.

mamiu:
@SantoDE Thanks for your fast response!

I fully understand the difficulty you described, but I don't find it hard (or "not obvious") to understand at all. On the contrary, even that doesn't really sound like a problem to me. Because the different nodes have shared access to the same volume anyway, the first Traefik instance that is up and running can create a lock file so that all other Traefik instances don't initiate another challenge. This lock file should either contain the network data of the initiated Traefik instance so that in case another Traefik instance receives the response from the ACME provider, the response can be forwarded to the appropriate pod (the Traefik instance that initiated the challenge) or the lock file should contain all the information which are necessary for other Traefik instances to be able to solve the DNS challenge.

Or do I still miss something that makes this scenario impossible to implement?

PS: Schöne Grüße aus Australien 😉

davidalger:
> As this is a bit hidden or might not be obvious to understand immediately, we will not support a deamon set like deployment by default.

@SantoDE I really fail to see (as I think could be said of many others here) how this is in any way different than running a Deployment with deployment.replicas set to any value greater than 1 (something this chart currently supports doing, and the anti-affinity recommendations do in fact make no sense apart from running a multi-node deployment of Traefik)

Most of the technical "reasons" I've seen provided come across as copouts. There are caveats to be sure, hence the following note being made above:

> Traefik v2 will not support LetsEncrypt in multi-node mode, so the option of a daemonset will not work if users want to use LE.

Document the _caveat_ that LE support will not function when a DaemonSet is used. Unless I'm missing something, this chart already supports multi-node deployments using X replicas on the Deployment, where the same caveat would apply, limiting the ability to LE support to function.

Long-term, as others have inferred, there really is no reason LE support couldn't be made to work where ReadWriteMany volumes are used, but short of that happening, whey not let the overwhelming community support from the user-base Traefik serves to guide this chart's direction and add DaemonSet support with a technical caveat/limitation when DaemonSet is in-use?

notsureifkevin:
> Or do I still miss something that makes this scenario impossible to implement?

@mamiu You're not missing anything, however, Traefik is a project which works with many other platforms outside of kubernetes. While this could be achieved with something like a lock file, and shared storage, this also presents a potential race condition. What happens if the response comes in before the file is released by the owner (we have no control over the latency of the underlying storage subsystem). As a result, a whole bunch of contingencies and concurrency problems are suddenly something we would be responsible for -- this is why the team decided to abandon support for distributed LE in Traefik v2, it caused a lot of headache.

That being said, K8S has a project already which solves these concerns, https://cert-manager.io/docs/installation/kubernetes/ so it doesn't really make sense to bring this capability into Traefik when it's already solved elsewhere, and with that you can use Traefik as a Daemon/ReplicaSet without the limitation of a single instance for LE support.

@SantoDE @davidalger 

> why not let the overwhelming community support from the user-base Traefik serves to guide this chart's direction and add DaemonSet support with a technical caveat/limitation when DaemonSet is in-use?

Would it not be possible to limit the install of a DaemonSet using similar logic that blocks more than 1 replica being deployed when LE is enabled? ex: https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/deployment.yaml#L1-L10

ShakataGaNai:
Should be simple enough. Just tell people "Do not deploy LE and multiple copies (or in DaemonSet)". I'm fairly sure that used to be in the documentation even. 

Traefik v2 works perfectly well in a DaemonSet, using CertManager if you're into TLS. So all we need is for merge requests to actually be allowed for that feature.

davidalger:
@notsureifkevin 

> Would it not be possible to limit the install of a DaemonSet using similar logic that blocks more than 1 replica being deployed when LE is enabled? ex: https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/deployment.yaml#L1-L10

That's actually a really good point, it's totally possible to have that same sort of guard rail in place, preventing installation using DaemonSet in combination with ACME. Personally I would see no issue with that.

mamiu:
@notsureifkevin Although I think that the problem you described could be solved relatively easily, the overall distributed LE support is a not-so-trivial solution and I definitely agree that Traefik doesn't need to implement and maintain such a feature if there's already a working alternative!
Does the Cert-Manager work together with Traefik v2.x CRDs and if so, are there any code samples available?

And yes, I absolutely agree with you and the previous comments that an error message (like [this one](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/deployment.yaml#L6)) is absolutely sufficient to include the DaemonSet feature.

notsureifkevin:
> > unfortunately Cert-Manager cannot interface directly with the CRDs _yet_, but this is being worked on by our team
> 
> Do you know the current state of the development of this feature?

I do not, but perhaps @SantoDE can provide some additional feedback on that. It would certainly be nice to have.

SantoDE:
@davidalger yes that should be possible. Maybe we should try & poc on this?

@mamiu : Currently, we have technical difficulties to provide a technical integration (clashing dependencies). At least, last time we tried. Maybe, I`ll give it another shot soon!

mamiu:
Thanks to all!
I'm really keen on updates regarding the DaemonSet option and the interface between Cert Manager and CRDs.

mamiu:
@SantoDE 

Just wanted to check back regarding the Cert-Manager integration with the Traefik IngressRoute CRD.

Here my question from above:

> > unfortunately Cert-Manager cannot interface directly with the CRDs _yet_, but this is being worked on by our team
> 
> Do you know the current state of the development of this feature?

And that was your answer:

> @mamiu : Currently, we have technical difficulties to provide a technical integration (clashing dependencies). At least, last time we tried. Maybe, I`ll give it another shot soon!

Is the issue with the clashing dependencies resolved by now? We're still keen to see this feature coming.

There's also a ticket about this in the Cert-Manager repo: https://github.com/jetstack/cert-manager/issues/3808

PR #59, opened 2020-01-03
Switch Chart linting to ct

dduportal:

- Move chart related content to the directory `./traefik` to fulfill Helm Chart convention (where "chart name" == "name of directory containing the chart).
  - Remove Makefile's code related to `SHIM_DIR` (which was a trick to remove the folder constraint)
- Add "testing/linting" related elements into a new `./test/` directory
- Switch the `lint` makefile phase to using `ct` (the official helm chart testing tool - https://github.com/helm/chart-testing). 
  - This `ct` command does the `helm lint ...` for us, but also add plenty more checks as the version increment bump 💪 
  - The `ct` command is run using the docker container, to allow better portability, as there are a lot of dependencies to install, and we want contributors to have an easy life


dduportal:
To do: add a documentation in `test/README.md` which describes everything (please don't merge before this)

dduportal:
Ping @ldez @SantoDE @dtomcej , this PR is ready to be reviewed :)

Makefile
```diff
@@ -10,15 +10,18 @@ HELM_REPO ?= $(CURDIR)/repo
 all: clean lint build deploy
 
 # Ensure the Helm chart and its metadata are valid
-lint: helm $(SHIM_DIR)
+lint: helm
 	@echo "== Linting Chart..."
-	@helm lint $(SHIM_DIR)
+	@git remote add traefik https://github.com/containous/traefik-helm-chart >/dev/null 2>&1 || true
+	@git fetch traefik master >/dev/null 2>&1 || true
+	@docker run --rm -t -v $(CURDIR):/charts -w /charts quay.io/helmpack/chart-testing:v3.0.0-beta.1 \
```
dtomcej: This puts a docker requirement for `make lint`... is that something that we feel is acceptable?

Kind of nice to be able to run a quick lint without docker running.

dduportal: Good point. 

=> The requirements carried by `ct`'s docker image are listed here: https://github.com/helm/chart-testing#prerequisites . It looks acceptable for me to require it, and fail fast (e.g. the Makefile fails with a simple error message if the requirement is missing).


=> We could also use a variable to use or not docker, likewise the traefik build makefile also.

dduportal: Oh, I realize that I've put the `docker` make target, which ensure that it is a requirement, is in #60 while it should be here. Let me transplant it here, so the change will be atomic.

Makefile
```diff
@@ -1,24 +1,36 @@
...
+# Ensure the Helm chart, YAMLs and metadatas are valid
+
+ifeq ($(LINT_USE_DOCKER),true)
+lint: docker
+else
+lint: helm ct
+endif
 	@echo "== Linting Chart..."
-	@helm lint $(SHIM_DIR)
+	@git remote add traefik https://github.com/containous/traefik-helm-chart >/dev/null 2>&1 || true
```
ldez: Could you explain this line?

Makefile
dduportal: @dtomcej running `make lint` without Docker is now possible, with requirements check, and documented.

Makefile
dduportal: Version increment check is done by `ct`, and we have to provide the name of a remote for this in `test/ct.yaml` (or to the `ct` CLI with a flag).

We cannot be sure that the original remote always have the same name for contributors (example: they don't use the pattern `upstream`, or fork of a fork, etc.).

This code ensure there is always a remote named `traefik` to compare to, which is the original one. Please note that it's expected to be the case on contributors repository, as we expect the version check to be run when they execute a `make lint` locally.

I took this trick from the `helm/charts` repo for information.

Maybe there is another way (or a better way) to do this?

PR #60, opened 2020-01-04
Add Unit Tests for the chart

dduportal:
**Important:** This PR is blocked by #59 and requires it.

- Add "Unit test" phase on the chart:
  - Use the make goal `unit-test` for this
  - Implemented with <https://github.com/rancher/helm-unittest>, fixed version. It's a fork of a repository which does not have activity since 8 month, while rancher maintains their fork. I tried with https://github.com/xchapter7x/hcunit but it does not provide "test suite runner": you must script commands OR write a single big OPA rego files 😨 
- Update directory structure:
  - Former directory ".test/" (from #59) renamed to "./lint/" the unit tests MUST be inside the chart directory. For this reason, the test's documentation has been moved to the root of the repository
  - Unit tests are stored in the default path of `helm-unittest`, aka. `./traefik/tests`, which is ignored when packaging the chart.
- Removing support for Kubernetes cluster under 1.9.0:
  - The "if case" was failing all the unit tests systems
  - The other objects (service, CRD, etc.) are using features which won't work with Kubernetes 1.9.x. I felt this "if case" was a copy and paste from the Traefik v1 helm chart.
  - Helm v3 changes the ".Capabilities" object structure: this "if case" is not working with helm v3
  - Kubernetes 1.9 is not updated since februrary 2019 (ref. https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md).


dduportal:
Thanks @SantoDE ! Link fixed.

Btw, #59 had been rebased to master, and this one as well. Don't forget to review #59 first :)

ldez:
@dduportal could you rebase on master?

dduportal:
Ready to roll!

PR #67, opened 2020-01-21
Make NodePort configurable

pamri:
This PR makes the NodePort configurable.


ldez:
Could you update the chart version, thanks.

pamri:
> Could you update the chart version, thanks.

Done.

PR #76, opened 2020-01-31
Allow exposing dashboard via configurable entrypoints

### LINKED ISSUE #79: EntryPoints limited to "web"

ncmans:
It's common practice to access privileged interfaces via `kubectl` proxy and port-forwarding tools. This PR exposes the dashboard via the unexposed default `traefik` entrypoint which is available on port 9000 on the traefik pods. This change makes the helm chart more plug and play without negatively affecting security.

Fixes #79

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: traefik
-version: 4.1.2
+version: 4.1.3
 appVersion: 2.1.3
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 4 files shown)


mmatur:
Hi @ncmans 

Thanks for you contribution.

Could you please rebase and bump chart version.

ncmans:
@mmatur done :+1:

mmatur:
@ncmans Thanks for the quick answer. 

After a discussion with the Traefik team we have decided to not give the possibility to customise the dashboard entry-point and to serve the dashboard on `traefik` entry-point that is not exposed by default.

I am closing this PR in favor of #100

PR #82, opened 2020-02-18
feature: make dashboard entryPoints configurable

### LINKED ISSUE #79: EntryPoints limited to "web"

bernot-dev:
As mentioned in #79, this PR allows dashboard entryPoints to be configurable.

Fixes #79

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 3.5.0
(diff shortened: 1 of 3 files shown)


ldez:
Same goal as #76 but not exactly the same approach.

---
You have increment the chart version.

bernot-dev:
#76 is better approach, and complies with TDD guidelines. Closing this PR in favor of other approach.

PR #86, opened 2020-02-22
Added persistent volume support.

trankchung:
This PR is to support persistent data so certificates are intact on pod restart.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: traefik
-version: 4.1.1
+version: 4.1.2
 appVersion: 2.1.3
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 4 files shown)


portswigger-tim:
This clearly restricts you to one replica when persistence is enabled 😬 - why did you not use StatefulSet with volumeClaimTemplate? 🤦

zek:
> Why would you need more than one replica? The saved acme.json file is exactly the same on each persistent volume.

High Availability

mloiseleur:
@portswigger-tim @ckt114 @zek Feel free to submit a PR if you think you can improve the current options on storage.

zek:
> @portswigger-tim @ckt114 @zek Feel free to submit a PR if you think you can improve the current options on storage.

I've decided that I don't need volumes to store the certificates. I don't see any benefits. Therefore I stick with cert-manager.

PR #87, opened 2020-02-23
Support hostPort

cameronbraid:
I wish to scale traefik to one replica per host (origin),  and use a hostPort of 443/80 then use a cloudflare load balancer to balance between my hosts (origins)

### DIFF
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -59,6 +59,9 @@ spec:
         - name: {{ $name | quote }}
           containerPort: {{ $config.port }}
           protocol: TCP
+          {{- if $config.hostPort }}
+          hostPort: {{ $config.hostPort }}
+          {{- end }}
         {{- end }}
         args:
           - "--global.checknewversion={{ .Values.additional.checkNewVersion }}"


mmatur:
Hi @cameronbraid,

Thanks for you contribution.

You should be able to solve this use case by using nodePort in the`values.yaml` file.
 
- HostPort: 
> Don’t specify a hostPort for a Pod unless it is absolutely necessary. When you bind a Pod to a hostPort, it limits the number of places the Pod can be scheduled, because each combination must be unique. If you don’t specify the hostIP and protocol explicitly, Kubernetes will use 0.0.0.0 as the default hostIP and TCP as the default protocol. [kubernetes.io](https://kubernetes.io/docs/concepts/configuration/overview/#services)

- NodePort:
> If you explicitly need to expose a Pod’s port on the node, consider using a NodePort Service before resorting to hostPort. [kubernetes.io](https://kubernetes.io/docs/concepts/configuration/overview/#services)

cameronbraid:
A node port is not a satisfactory solution.

With a 3 node cluster and with 3 traefik pods -  one on each node.    Having a node port configured can lead to http connection to one node being proxied to traefik pod on another node.  

What a host port allows you to do is have the traefik pod listen on the port on the actual host, thereby preventing the need to proxy the port, and ensuring the connection remains on that host.

PR #90, opened 2020-02-27
Add Service Monitor

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

This PR adds the Service Monitor useful when Traefik is deployed in a cluster where Prometheus Operator is deployed.

Prometheus metrics flags have been added to Traefik command.

The approach follows the one from the stable Helm chart.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 3.5.0
+version: 3.6.0
 appVersion: 2.1.3
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -70,6 +70,11 @@ spec:
           - "--ping=true"
           - "--providers.kubernetescrd"
           - "--log.level={{ .Values.logs.loglevel }}"
+          {{- if .Values.metrics.prometheus.enabled -}}
+          - "--metrics.prometheus=true"
+          - "--entryPoints.metrics.address=:9100"
+          - "--metrics.prometheus.entryPoint=metrics"
+          {{- end -}}
           {{- with .Values.additionalArguments }}
           {{- range . }}
           - {{ . | quote }}
--- a/traefik/templates/service-prometheus.yaml
+++ b/traefik/templates/service-prometheus.yaml
@@ -1,0 +1,20 @@
+{{- if .Values.metrics.prometheus.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "traefik.fullname" . }}-prometheus
+  labels:
+    app: {{ template "traefik.name" . }}
+    chart: {{ template "traefik.chart" . }}
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
+spec:
+  type: ClusterIP
+  selector:
+    app: {{ template "traefik.name" . }}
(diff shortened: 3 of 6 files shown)


mazzy89:
I add to rename from `service.yaml` to `service-prometheus.yaml` the file because was colliding under unit test with the `service.yaml` in the parent dir.

Wondering if there is a better way to handling this naming collision.

mazzy89:
There is an opened issue in the upstream repo https://github.com/lrills/helm-unittest/issues/44

At the current stage seems not possible to have subdirs having files with the sane name.

mazzy89:
I'm going to add some unit tests later today.

ldez:
Hello,

could you test locally?

FYI we receive a mail for each commits.

mazzy89:
@ldez Sure. I will. Sorry for spamming here.

mazzy89:
Unit tests green here. Ready for review.

myco:
@mazzy89 
Is it possible to enable this feautre by a CLI parameter to helm install?

mazzy89:
Traefik flags are defined . Service monitor is not a feature of Traefik but it is a CR belonging to prometheus operator

mazzy89:
@ldez do you see here any blockers?

mmatur:
Hi @mazzy89,

Thanks for your contribution. 

We should probably only enable Prometheus in this PR and keep the ServiceMonitor outside of the helm chart. 
Because adding option like that will increase the complexity of the helm chart. And I think we should keep the helm chart as simple as possible. This option is specific to people that have Prometheus Operator installed and want to use ServiceMonitor. So we will probable have other people that would want to have PodMonitor and maybe other people that would like to use annotation to scrape metrics.

So we should keep thing most generic as possible and as simple as possible.

So could you please remove the service monitor part and only keep the prometheus configuration?

mazzy89:
I see. No problem. I'm going to remove ServiceMonitor.

mmatur:
Thanks 

FYI I worked on an example available here https://github.com/mmatur/prometheus-traefik

mmatur:
@mazzy89 

Metrics on traefik can be enable with the following `values.yaml`:

```yaml
additionalArguments: 
  - "--metrics.prometheus=true"
  - "--entryPoints.metrics.address=:9100"
  - "--metrics.prometheus.entryPoint=metrics"
```

For this reason I will close your PR. 

Thanks a lot for your contribution

mazzy89:
True but it would have been nice to put the configuration in the `values.yaml` rather than to play with flags that can be more error prone. But as you say

n4nn31355:
It's very strange decision not to include a feature flag with ServiceMonitor. There is no convenient and simple way to add custom templates to existing charts, so you just kill the value of this chart when decline this widely used feature.

mazzy89:
@n4nn31355 thanks for this. I completely agree with you. The maintainer here claims to keep the helm chart simple without including it but I really have hard time to get how including such CR like service monitor can increase complexity. We are talking here of a bunch of YAML manifests baked even by unit tests.  I would agree if there was some lesson learned with the helm chart for traefik 1.x but I don't think there is here, that helm chart is there for pretty long time and it is is stable and more important not hard at all.

DirtyCajunRice:
@mazzy89 @n4nn31355 They are just sticking with the "if its k8s native or traefik native it can be added, otherwise it has conflicts of interest". That said if you add the incubator/raw chart as a subchart you can accomplish adding a servicemonitor extremely simply!

PR #91, opened 2020-02-27
Add Compress middleware

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

This PR adds the `Compress` middleware.

I haven't found a way to configure this middleware via command flags. The recommended way to configure `Compress` in Kubernetes is to add an object of Kind `Middleware`.

Please advice here whether there is a different direction recommended.

### DIFF
--- a/traefik/templates/middlewares/compress.yaml
+++ b/traefik/templates/middlewares/compress.yaml
@@ -1,0 +1,15 @@
+{{- if .Values.middlewares.compress.enabled }}
+apiVersion: traefik.containo.us/v1alpha1
+kind: Middleware
+metadata:
+  name: {{ template "traefik.fullname" . }}
+  labels:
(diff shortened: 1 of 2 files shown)


mazzy89:
Failing because version must be bumped. I will fix it after first review here.

mmatur:
Hi @mazzy89,

Thanks for your contribution. 

I'm not sure it is the correct way to create middlewares.

We would like to keep the helm chart as simple as possible. 

Like it is explained in the [documentation](https://docs.traefik.io/getting-started/configuration-overview/#the-dynamic-configuration), middlewares are part of the dynamic configuration.

So we don't want to update the helm chart to be able to apply an update on middlewares.

mazzy89:
I see. fair enough.

PR #93, opened 2020-03-05
Daemonset support

pamri:
Based on #56, this PR adds support for daemonset. It has the following changes:
- Separate template for Deployment / DaemonSet
- Added test cases for Daemonset
- There is one breaking change, which is, rollingUpdate is moved under deployment and daemonset, since the specs are slightly different.

As a side note, aware of https://github.com/helm/charts/pull/11813#issuecomment-530001942 , but needed it for our use case, which is quite similar to https://github.com/helm/charts/pull/11813#issuecomment-530790527

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 3.5.0
+version: 3.5.1
 appVersion: 2.1.3
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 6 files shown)


mmatur:
Hi @pamri 

Thanks for your contribution.

We have been discussing a lot within the Traefik team and we have decided to close this PR for reasons explain [here](https://github.com/helm/charts/pull/11813#issuecomment-530001942)

> Hello All,
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> Thanks,
> 
> The Traefik Team.

PR #97, opened 2020-03-10
Not authorise acme if replica > 1

mmatur:
### Description

This PR adds a security to avoid to enable acme and have multiple replicas of Traefik


Mokto:
Hi, what if you use a ReadWriteMany volume?

mmatur:
I am introducing this change to not allowed user to have multiple  traefik instances when they have acme enable. 

I think it is not the best approach. So I am working to improve that.

PR #100, opened 2020-03-10
Expose dashboard by default but only on traefik entrypoint

### LINKED ISSUE #79: EntryPoints limited to "web"

mmatur:
### Description

This PR exposes the dashboard by default but only on the entry-point `traefik`. 

This entry-point is not exposed. So it give the possibility to users to access to this dashboard with a port-forward:

```
kubectl port-forward traefik-xxx-xxx 9000:9000
# http://localhost:9000/dashboard
```

If people want to expose the dashboard in another entry-point they can easily creates an IngressRoute to do it.

Closes #79 #76


yidongw:
You can either use port-forward or create an IngressRoute to expose the dashboard, but how is `traefik/templates/dashboard-hook-ingressroute.yaml` useful at all?

I spent hours trying to figure out how to use the `dashboard-hook-ingressroute.yaml` to expose the dashboard. 
My intuition is that if there is an `IngressRoute` that is kinda doing the exposing thing, I must be able to mutate it a little in order for me to access it from the Service.

The answer is I can't. `dashboard-hook-ingressroute.yaml` is now become useless.

I would suggest either removing `dashboard-hook-ingressroute.yaml` from the `templates` because it is causing confusion for people like me or making `entryPoints` changeable so we can actually use it.

PR #112, opened 2020-03-13
hpa, dashboard via standard ingress

### LINKED ISSUE #48: Horizontal Pod Autoscaler
In the current specification, the HPA isn't enabled, allowing only the value of replicas to be changed.

Is there a plan to support this kind of scaling based on CPU and/or memory as is done in the previous version?

https://github.com/helm/charts/blob/master/stable/traefik/values.yaml#L482-L493

I can port this to this chart if that is the plan.

travisghansen:
This allow for exposing the dashboard via a standard `Ingress` object *without exposing the service to the `LoadBalancer` service for actual inbound traffic.

Additionally introduces an `hpa` for automatic scaling.

Fixes #48

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: traefik
-version: 5.6.0
+version: 5.7.0
 appVersion: 2.1.3
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/templates/dashboard-ingress.yaml
+++ b/traefik/templates/dashboard-ingress.yaml
@@ -1,0 +1,34 @@
+{{- if .Values.dashboard.ingress.enabled }}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: {{ template "traefik.fullname" . }}-dashboard
+  labels:
+    app: {{ template "traefik.name" . }}
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
+  {{- if .Values.dashboard.ingress }}
+  {{- range $key, $value := .Values.dashboard.ingress.labels }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
+  {{- end }}
+  annotations:
+  {{- if .Values.dashboard.ingress }}
+  {{- range $key, $value := .Values.dashboard.ingress.annotations }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
+  {{- end }}
+spec:
+  rules:
+  - host: {{ .Values.dashboard.ingress.domain }}
+    http:
+      paths:
+      - backend:
+          serviceName: {{ template "traefik.fullname" . }}-dashboard
+          servicePort: traefik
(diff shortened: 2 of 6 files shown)


travisghansen:
@mmatur yeah, I realized there's possibly still some security considerations that need to be dealt with after sending this over. I'm going to run through a couple use-cases to see what happens. Ultimately (this may already exist, was just getting ready to dig into it) I think the `kubernetesingress` provider should be able to set default entrypoints (ie: just `web` and `websecure`) so the `traefik` entry point is not added to every `router` created from an `Ingress`. If that's possible, then I *think* this should be secure (or as secure as the operator makes the `Ingress`). The dashboard `Ingress` would simply have `traefik.ingress.kubernetes.io/router.entrypoints: traefik` using the new annotation features from `v2.2` and all other `Ingress` resources would have the defaults (or specifically **not** `traefik`).

My motivations for using `Ingress` vs `CRD` are quite lengthy. I recently shared via a private email with @geraldcroes which he can send over if you'd like. I can get you more info if he cannot.

travisghansen:
My previous comment about setting the annotation would not work..sorry still wrapping my brain around 2.x a bit along with the intersection of options needed to make this all work.

After a bit more analysis I *think* the combination of options I have are secure (or again, as secure as I configure auth on the `Ingress` to be). Relevant options:

```
ports:
  traefik:
    expose: false
    # https://github.com/containous/traefik/issues/6474
    websecure:
      port: 443

additionalArguments:
- "--providers.kubernetesingress=true"
- "--api.insecure=true"

dashboard:
  ...
```

Because the non-dashboard `Service` does not expose the `traefik` entrypoint directly, traffic will never reach the `@internal` traefik services via the `LoadBalancer`. Additionally, the newly introduced `Service` specifically for use with the dashboard defaults to `ClusterIP` preventing access to the service from undesirable locations (someone could still access it via the `NodePort` if they had access to the node network..but that's not the use-case).

Let me know if I've got something off in my analysis.

travisghansen:
@mmatur any comments on the above? With 2.2 out of like to see this merge to expand broader testing.

mmatur:
@travisghansen 

I'm agree with your proposition about hpa.

But I'm absolutely not convince of the benefit to manage the exposition of the dashboard with the helm chart. To me it adds a lot of complexity of the helm chart usage and give absolutely not benefit. 

The exposition and security of the dashboard need to be done outside of the helm chart because there are so many way to handle it.

travisghansen:
The approach is exactly the same as the community chart. I'm certainly not alone in not wanting to use `CRD`. You're suggesting I create these resources outside the chart and not have all the ingress resources managed as a single unit (helm release)?

mmatur:
Today the helm chart Traefik v1 is very complex to maintain and we don't want to reproduce that on this chart.

I can totally understand that you don't want to use the `CRD`.  Yes I'm suggesting you to create these resources outside of this chart.

Many solution exist to manage this type of thing like creating your own chart with [dependencies](https://helm.sh/docs/topics/charts/#chart-dependencies).

travisghansen:
Ugh. Complex but useful.

Ok, I'm happy with the changes proposed for the hpa..would you like to simply cherry pick that feature and close this down?

mmatur:
If you can close this one and create a PR for HPA please it would be great.

Thanks

travisghansen:
Moved to #126

PR #116, opened 2020-03-16
Add ability to add annotations to deployment

### LINKED ISSUE #113: Easy integration with jaeger-operator
In [values.yaml](https://github.com/containous/traefik-helm-chart/blob/master/traefik/values.yaml#L13) we have access to pod annotations, however sidecar injection for jaeger operator requires annotations on the deployment. Having a seperate entry in values for that would be useful.

arlyon:
The jaeger operator automatically injects sidecars to any deployment annotated with "sidecar.jaegertracing.io/inject" (described [here](https://www.jaegertracing.io/docs/1.17/operator/#auto-injecting-jaeger-agent-sidecars)). It would be beneficial to allow the helm chart to support this.

Not sure if we need a minor or patch version bump, I'll leave that to your discretion.

Closes #113

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: traefik
-version: 6.0.2
+version: 6.1.0
 appVersion: 2.1.3
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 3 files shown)


dtomcej:
@arlyon I would just do a patch bump.

dtomcej:
But other than that, it looks good to me 🎉

ldez:
could you bump to version to 6.1.0

PR #117, opened 2020-03-16
Change default values of arrays from objects to actual arrays

jasperkuperus:
Some default values in `values.yaml` are initialized with `{}`, while they should be `[]` in my opinion. When this is not the case, everything works fine when you override them, but you'll get a warning like:

```
coalesce.go:196: warning: cannot overwrite table with non table for env (map[])
```

Fixes https://github.com/containous/traefik-helm-chart/issues/74

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: traefik
-version: 6.0.0
+version: 6.0.1
 appVersion: 2.1.3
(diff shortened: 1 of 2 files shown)


jasperkuperus:
Not sure why the build fails. Locally I get this on linting:

```
$ helm lint -f values.yaml 
==> Linting .

1 chart(s) linted, 0 chart(s) failed
```

PR #120, opened 2020-03-16
Correct storage class key name

ThetaSinner:
Updated the key name for the storage class to allow the storage class to be set.

Documentation for the pvc resource [here](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistent-volumes)

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: traefik
-version: 6.0.1
+version: 6.0.2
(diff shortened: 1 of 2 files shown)


ThetaSinner:
@trankchung @mmatur I'd appreciate your input on this one, it looks like you're the owners of this file :)

trankchung:
Good catch. Looks like I did a fat finger. Your PR is correct. This needs to be merged to fix the issue. I had a default storageClass in my cluster so it worked for me without needing to specify a storage class.

PR #127, opened 2020-03-26
add servicemonitor for prometheus operator  

Vad1mo:
the pr allows enabling prometheus metrics and adds a Service Monitor CR.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: traefik
-version: 6.2.0
+version: 6.3.0
 appVersion: 2.2.0
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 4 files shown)


mmatur:
Hi @Vad1mo,

Thanks a lot for your contribution. Unfortunately we will not be able to accept your PR for few reasons:

- Metrics can already be enable with `additionalArguments` properties:

```yaml
additionalArguments: 
  - "--metrics.prometheus=true"
```

- `ServiceMonitor` is a CRD specific to the prometheus operator. As you probably know traefik is compatible with [multiple metrics systems](https://docs.traefik.io/observability/metrics/overview/). We don't want to introduce CRD from third party mostly to avoid to add complexity to the helm chart.

- An option `prometheus.additionalServiceMonitors` exist in the [prometheus operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator#prometheus) to creates additional `SeviceMonitors`.

You can find a full example of the configuration of traefik with the prometheus operator [here](https://github.com/mmatur/prometheus-traefik)

Thanks again for your contribution.

PR #129, opened 2020-03-27
Add ability to set serviceAccount annotations

relu:
This comes in useful when dealing with the Route53 dns challenge provider on AWS.


traefik/templates/rbac/serviceaccount.yaml
```diff
@@ -2,3 +2,8 @@ kind: ServiceAccount
 apiVersion: v1
 metadata:
   name: {{ template "traefik.fullname" . }}
+  annotations:
+  {{- range $key, $value := .Values.serviceAccountAnnotations }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
+  {{- end }}
```
mmatur:
```suggestion

```

Issue #64, opened 2020-01-17
How to use daemonset type install traefik?


fengfuyun52:
Hello there,
I want to know that how can we use daemonset type install traefik?
Thanks very much!

prima101112:
why you want to install traefik as daemonset?
it could be done by manually modify the templates but by defaukt traefik will run as standard deployments i think.

travisghansen:
Several reasons potentially, as I see it generally `DaemonSet` would likely be used in environments that:

- Do not have support for `LoadBalancer` services

and/or

- Are fronting the cluster with an external/existing load balancer

It's certainly not uncommon for `Ingress` controllers to be installed as a `DaemonSet` using host networking.

Fank:
Another reason, installing a microk8s on a single vm and wanted to use traefik as default ingress controller.

acim:
Yes, this would be really nice to have for self-managed clusters with no LoadBalancer service by cloud provider, for example Hetzner Cloud. By the way, nginx-ingress chart has DaemonSet possibility.

toabi:
The documentation until v2 explicitly mentioned a DaemonSet install: https://docs.traefik.io/v1.7/user-guide/kubernetes/#deploy-traefik-using-a-deployment-or-daemonset

> If you are unsure which to choose, start with the Daemonset.

Our use case for DaemonSet-Install:

Cluster in EKS with two Node Pools:

Pool 1: Workers - Not put in any target group
Pool 2: Ingress Nodes - Automatically spawn into a "ingress" target group and get node label `app=ingress`

We then would install traefik as DaemonSet on app=ingress Nodes and provision an terraform-managed ALB with forwards to that target group.

At least that was the way to go with ingress-nginx and it was nice. But now we need more features from our ingress controller.

davidalger:
Fixed by #298

jakubhajek:
Thank you @davidalger for the #298, which allows using Deamonset with Traefik.

PR #141, opened 2020-04-07
Add DaemonSet template

visagey:
A simple DaemonSet. See #64 for the "why".
It is pretty much a copy of `deployment.yaml`.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 7.2.1
+version: 7.3.0
 appVersion: 2.2.0
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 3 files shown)


mmatur:
Hi @visagey 

Thanks for your contribution.

For the same reason as #54 #56 #93  we have decided to close your PR (cf [here](https://github.com/helm/charts/pull/11813#issuecomment-530001942))

> Hello All,
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> Thanks,
> 
> The Traefik Team.

visagey:
> For this reason, it makes more sense to use a deployment with anti-affinity to push pods to all nodes instead of using a daemonset.

That is too bad, deployment with anti-affinity does not replace a daemonset: when a node is down, a new pod is created on a node where there is already a pod. And then, when the node is up again, no pod is created on this node.

the-pinky-man:
@mmatur  Deployment out the box for general use and the reasons listed, DaemonSet with some extra configuration for people that want it - why cant it support both?

PR #144, opened 2020-04-07
Make helm chart work with UDP routes.

### LINKED ISSUE #125: Change deployment & service for udp support
There should be support to add `/udp` to `--entrypoint` in the deployment, this should result in multiple changes to container ports and service port, i suggest to make it simple and add `protocol` to `ports:`

bergmannf:
This is a first attempt to fix #125 

The ugly part is that it is not possible to run the `TCP` and `UDP` routes in the same `LoadBalancer`, so instead I have to loop over the ports and will create a service for each `port`.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.0.0
+version: 8.1.0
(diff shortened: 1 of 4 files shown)


bergmannf:
Superseded by #145

PR #145, opened 2020-04-07
UDP support

mmatur:
### Description

This PR adds UPD support with another approach than #144

### DIFF
--- a/traefik/templates/service.yaml
+++ b/traefik/templates/service.yaml
@@ -11,6 +11,7 @@
 {{- end }}
 
 apiVersion: v1
+kind: List
 items:
 {{- if  $tcpPorts }}
   - apiVersion: v1
@@ -88,7 +89,7 @@ items:
       - port: {{ default $config.port $config.exposedPort }}
         name: {{ $name }}
         targetPort: {{ $name | quote }}
-        protocol: {{ default "TCP" $config.protocol | quote }}
+        protocol: {{ default "UDP" $config.protocol | quote }}
         {{- if $config.nodePort }}
         nodePort: {{ $config.nodePort }}
         {{- end }}
@@ -105,6 +106,4 @@ items:
       {{- toYaml . | nindent 6 }}
       {{- end -}}
 {{- end }}
-
-
 {{- end -}}
--- a/traefik/templates/service.yaml
+++ b/traefik/templates/service.yaml
@@ -27,6 +27,7 @@ spec:
   - port: {{ default $config.port $config.exposedPort }}
     name: {{ $name }}
     targetPort: {{ $name | quote }}
+    protocol: {{ default "TCP" $config.protocol | quote }}
     {{- if $config.nodePort }}
     nodePort: {{ $config.nodePort }}
     {{- end }}
--- a/traefik/templates/service.yaml
+++ b/traefik/templates/service.yaml
@@ -1,46 +1,110 @@
 {{- if .Values.service.enabled -}}
+
+{{ $tcpPorts := dict }}
+{{ $udpPorts := dict }}
+{{- range $name, $config := .Values.ports }}
+  {{- if eq ($config.protocol | toString) "UDP" }}
+    {{ $_ := set $udpPorts $name $config }}
+  {{- else }}
+    {{ $_ := set $tcpPorts $name $config }}
+  {{- end }}
+{{- end }}
+
 apiVersion: v1
-kind: Service
-metadata:
-  name: {{ template "traefik.fullname" . }}
-  labels:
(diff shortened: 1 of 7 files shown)


bergmannf:
I think just adding all ports to a single service will not work - if this PR is applied using `helm` to a cluster:

```
Error: Service "traefik" is invalid: spec.ports: Invalid value: []core.ServicePort{core.ServicePort{Name:"udp", Protocol:"UDP", Port:8099, TargetPort:intstr.IntOrString{Type:1, IntVal:0, StrVal:"udp"}, NodePort:0}, core.ServicePort{Name:"web", Protocol:"TCP", Port:80, TargetPort:intstr.IntOrString{Type:1, IntVal:0, StrVal:"web"}, NodePort:0}, core.ServicePort{Name:"websecure", Protocol:"TCP", Port:443, TargetPort:intstr.IntOrString{Type:1, IntVal:0, StrVal:"websecure"}, NodePort:0}}: cannot create an external load balancer with mix protocols
```

The cluster was running `1.17.2`:

```
❯ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:22:30Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
```

To tests this I just added a `UDP` port to the `values.yaml`:
```yaml
  udp:
    port: 8099
    expose: true
    exposedPort: 8099
    protocol: UDP
```

mmatur:
Thanks @bergmannf for the input.

So I think the best solution in this case is to create a second LB only for UDP connexions.

I will try to think to the best approach to solve this issue

bergmannf:
@mmatur Yes I think having only 2 LoadBalancers is the best way - I couldn't figure out a nice way to get this to work in my PR (but I'm also quite new to the `go template` language).

I will close my PR, as this one already has all the testing and looks generally nicer.

diegombeltran:
Hi! Any chance to get this merged?

demon:
If memory serves, AWS only supports UDP traffic on network load balancers--not the ELB we spin up by default. So using this would require one to set `service.beta.kubernetes.io/aws-load-balancer-type: "nlb"` on the Service (which I'm fine with, I much prefer NLBs anyway).

Some docs to this effect would be great, if not configuring the Service outright.

ryantiger658:
Any news on merging this in? I am anxious for UDP support!

traefik/templates/service.yaml
```diff
@@ -1,45 +1,110 @@
...
+      selector:
+        app.kubernetes.io/name: {{ template "traefik.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+      ports:
+      {{- range $name, $config := $udpPorts }}
+      {{- if $config.expose }}
+      - port: {{ default $config.port $config.exposedPort }}
+        name: {{ $name }}
+        targetPort: {{ $name | quote }}
+        protocol: {{ default "TCP" $config.protocol | quote }}
```
RouxAntoine, line 91: Default "TCP" even if we are in udp condition ?

https://github.com/containous/traefik-helm-chart/blob/ee94b9f10fed728d53fa7dcc576330ca6d241e70/traefik/templates/service.yaml#L62

traefik/templates/service.yaml
```diff
@@ -1,45 +1,110 @@
...
-  {{- toYaml . | nindent 2 }}
-  {{- end -}}
-  {{- end -}}
-  {{- with .Values.service.externalIPs }}
-  externalIPs:
-  {{- toYaml . | nindent 2 }}
-  {{- end -}}
+{{- end }}
+
+apiVersion: v1
```
RouxAntoine, line 13: Here it seems missing `kind: List`
helm complain : `Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: kind not set`

RouxAntoine:
Into this two set `protocol: TCP` is missing I think :

https://github.com/containous/traefik-helm-chart/pull/145/files#diff-76ff81f79f6efcd76b91e037d8126589R94

https://github.com/containous/traefik-helm-chart/pull/145/files#diff-76ff81f79f6efcd76b91e037d8126589R68

It seems the reason of build fail :)

PR #147, opened 2020-04-08
Add Support for subPath in data volumeMount

Jawastew:
If the PVC is requesting a shared PV it would be nice to not give traefik full root level access to the volume. (Useful if the PVC is accessing a nfs volume, for example)  

This would add the option to limit traefik to an arbitrary ```subPath``` on a volume.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.0.2
+version: 8.0.3
 appVersion: 2.2.0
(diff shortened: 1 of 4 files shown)


mmatur:
@Jawastew Thanks for your contribution

Could you please rebase and bump chart version.

Jawastew:
Done. I hope I did it correctly, this is my first open source pull request :)

ldez:
As the PR comes from an organization, our bot, due to a weird limitation from GitHub, is not able to merge this pull request.

PR #152, opened 2020-04-10
configurable match rule for dashboard

den-is:
as suggested in this comment https://github.com/containous/traefik-helm-chart/issues/143#issuecomment-610795188

If the dashboard tends to be exposed insecurely using usual Service chances high that it will be hosted using own URL.
So IngressRoute, if needed, should be able to filter traffic by Host header too.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.0.4
+version: 8.0.5
 appVersion: 2.2.0
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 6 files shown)


den-is:
@mmatur any specific reason why this is not getting merged?

mmatur:
Hi @den-is,

Thanks for your contribution and sorry for the late answer. 

Unfortunately we will not be able to merge your PR. 

We have documented [how to expose the dashboard](https://github.com/containous/traefik-helm-chart#exposing-the-traefik-dashboard) and we don't want to allow user to customized the dashboard exposition through the helm chart. 

Thanks again for your contribution.

Issue #138, opened 2020-04-04
Add possibility to bind environment variables via envFrom


sbaeurle:
Right now it is possible to deploy the helm chart and use ConfigMaps and Secrets outside of the Helm release to pass ENV values to traefik. (e.g. Secrets for ACME-DNS01 Challenges).

Yet you cannot pass complete configmaps via `envFrom` as seen in [Usage of ConfigMaps](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables). (also possible for secrets)

I suggest to add an extra configuration option in the `values.yml` to enable such scenarios, similar to the binding of individual env variables. (this would be very useful to e.g. dynamically enable/disable logging without the need to redeploy the chart)

Proposal:
`values.yml`:
```
envFrom:
- listOfConfigMapAndSecretRefs
```
`deployment.yml`:
```
spec:
  ...
  template:
    ...
    containers:
      ...
      {{- with .Values.envFrom }}
      envFrom:
        {{- toYaml . | nindent 10 }}
```

mmatur:
Hi @sbaeurle,

Seems a great improvement of the helm chart. 

Could you please open a PR?

sbaeurle:
Hi @mmatur,

will do so ;)

sbaeurle:
Hi @mmatur,

I finally applied the changes made here to my own deployment and encountered a problem rendering the feature practically useless for configuration.

According to the [Documentation](https://docs.traefik.io/getting-started/configuration-overview/#the-static-configuration) of Traefik, one can only use either a configuration file, cli arguments or env variables for configuration.
The options are mutually exclusive and since this chart automatically uses arguments (even after disabling the globalArguments) the env config is not loaded at all.

The feature is still perfectly usable for DNS challenge secrets, but utterly useless for configuration.

Maybe you could clarify 
1. If Traefik will support multiple static configuration sources in the future?
2. If you see any possibility to circumvent the current usage of cli arguments in the configuration and enable ENV configuration without multiple configurations sources?

Thanks in advance

sbaeurle:
@mmatur wanted to check if you had the possibility to clarify the env usage of traefik and the helm chart?

mmatur:
Hi @sbaeurle,

Sorry for my later answer.

> 1. If Traefik will support multiple static configuration sources in the future?

From what I know for now there is no plan to have multiple static configuration sources.

[...]

PR #154, opened 2020-04-17
Possibility to bind environment variables via envFrom

sbaeurle:
Pull Request to address enhancement as described in #138


sbaeurle:
@mmatur from my understanding i need to bump the version in order to fix the CI build.

If I'm right I would go with 8.0.4

mmatur:
@sbaeurle yes exactly

PR #155, opened 2020-04-19
Add priorityClassName & securityContext

steven-sheehy:
Adds 2 features:

- `priorityClassName`
- `securityContext` at both container and pod level
  - Defaults to run as non-root and read only filesystem, which is a change from previous. Default security context works with other defaults in values.yaml. If users are running with ports below 1024 they will need to adjust to run as root again.


mmatur:
@steven-sheehy Thanks for your contribution

Unfortunately we will not be able to accept the feature regarding the `ServiceMonitor` (cf https://github.com/containous/traefik-helm-chart/pull/127#issuecomment-604935789).

So could you rebase your PR and keep only feature on `priorityClassName ` and `securityContext `

steven-sheehy:
@mmatur Thanks for the review. I've removed the ServiceMonitor functionality if you'd like to take another look.

mmatur:
Thanks @steven-sheehy.

Could you please add tests on these two new features?

steven-sheehy:
@mmatur Tests added

PR #157, opened 2020-04-22
Add OpenAPI v3 schemas to CRDs

jrostand:
This allows upgrading to `apiextensions/v1` and I believe should solve containous/traefik#5473 and containous/traefik#6109. These were lovingly built by hand from the source code of traefik.

Things I'm unsure of:
1. Are these schemas correct/comprehensive? I'm particularly interested in any enum values that could be encoded.
2. Should they be behind a config option for k8s < 1.16?
3. Extra printer columns?
4. Can/should this schema generation be automated?

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.1.0
+version: 8.2.0
 appVersion: 2.2.0
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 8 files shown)


ldez:
Hello,

Thank you for that work.

We plan to handle this automatically in the Traefik repository.

Because of that, I'm sorry, but we will not able to accept your PR.

FYI, we are already working on it.

thank you anyway.

sponte:
Hello @ldez 

Hope you're well. Just curious if Traefik now provides these CRDs for Kubernetes v1.17? I couldn't find them in containous/traefik repository.

Thanks,
Stanislaw Wozniak

PR #168, opened 2020-04-29
Clarify additionalArguments with logs.level

emilevauge:
This PR adds an example with `logs.level` used in the [Traefik documentation](https://docs.traefik.io/getting-started/install-traefik/#use-the-helm-chart).

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -50,9 +50,9 @@ globalArguments:
 # Configure Traefik static configuration
 # Additional arguments to be passed at Traefik's binary
 # All available options available on https://docs.traefik.io/reference/static-configuration/cli/
-## Use curly braces to pass values: `helm install --set="additionalArguments={--providers.kubernetesingress,--global.checknewversion=true}"`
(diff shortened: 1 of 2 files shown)


traefik/values.yaml
```diff
@@ -50,9 +50,9 @@ globalArguments:
 # Configure Traefik static configuration
 # Additional arguments to be passed at Traefik's binary
 # All available options available on https://docs.traefik.io/reference/static-configuration/cli/
-## Use curly braces to pass values: `helm install --set="additionalArguments={--providers.kubernetesingress,--global.checknewversion=true}"`
+## Use curly braces to pass values: `helm install --set="additionalArguments={--providers.kubernetesingress,--logs.level=DEBUG}"`
 additionalArguments: []
-#  - "--providers.kubernetesingress"
+#  - "--providers.kubernetesingress, --logs.level=DEBUG"
```
mmatur, line 55:
```suggestion
#  - "--providers.kubernetesingress"
# - "--logs.level=DEBUG"
```

emilevauge: 😅my bad

PR #179, opened 2020-05-14
Add option to disable RBAC and ServiceAccount

steven-sheehy:
- Adds `rbac.enabled` to control whether ClusterRole and ClusterRoleBinding are created or not
- Adds `rbac.namespaced` to control whether Role/Rolebinding or ClusterRole/ClusterRoleBinding are created or not
- Adds `serviceAccount.name` to control whether ServiceAccount is created or not
- Adds tests for all combinations of new settings

Besides the `namespaced` property, these options are all the standard options that `helm create` provides out of the box.

The use case for disabling rbac and serviceAccount is some cloud providers don't allow RBAC related components being created. Specifically in our case, Google Cloud Marketplace [does not allow](https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/master/docs/building-deployer-helm.md#declare-rbac-requirements-and-disable-rbac-in-the-chart) any RBAC be created and will create them via another mechanism.

The use case behind the `rbac.namespaced` setting is that we want to deploy a different Traefik per namespace so our chart is more self-contained and secured. Also, our operations team locks down the cluster so we can't deploy cluster level components. Setting `rbac.namespaced=true` and setting `providers.kubernetesIngress.namespaces=mynamespace` in the Traefik config allows us to reduce the permission of Traefik and have it watch only our specific namespace.

To reduce duplication, I put the if/else for namespaced inside the clusterrole and clusterrolebinding. I can duplicate them to separate role/rolebinding files if you prefer that instead.

### DIFF
--- a/traefik/templates/rbac/clusterrole.yaml
+++ b/traefik/templates/rbac/clusterrole.yaml
@@ -1,4 +1,9 @@
+{{- if .Values.rbac.enabled -}}
+{{- if .Values.rbac.clusterScope }}
 kind: ClusterRole
+{{- else }}
+kind: Role
+{{- end }}
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ template "traefik.fullname" . }}
@@ -46,3 +51,4 @@ rules:
       - get
       - list
       - watch
+{{- end -}}
--- a/traefik/templates/rbac/clusterrole.yaml
+++ b/traefik/templates/rbac/clusterrole.yaml
@@ -1,8 +1,8 @@
 {{- if .Values.rbac.enabled -}}
-{{- if .Values.rbac.clusterScope }}
-kind: ClusterRole
-{{- else }}
+{{- if .Values.rbac.namespaced }}
 kind: Role
+{{- else }}
+kind: ClusterRole
 {{- end }}
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
--- a/traefik/templates/rbac/clusterrolebinding.yaml
+++ b/traefik/templates/rbac/clusterrolebinding.yaml
@@ -1,4 +1,9 @@
+{{- if .Values.rbac.enabled -}}
+{{- if .Values.rbac.clusterScope }}
 kind: ClusterRoleBinding
+{{- else }}
+kind: RoleBinding
+{{- end }}
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ template "traefik.fullname" . }}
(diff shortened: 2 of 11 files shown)


traefik/templates/rbac/clusterrole.yaml
```diff
@@ -1,4 +1,9 @@
+{{- if .Values.rbac.enabled -}}
+{{- if .Values.rbac.namespaced }}
```
mmatur, line 2: Could you please split this into 2 file. 
- clusterrole.yaml
- role.yaml

traefik/templates/rbac/clusterrolebinding.yaml
```diff
@@ -1,4 +1,9 @@
+{{- if .Values.rbac.enabled -}}
+{{- if .Values.rbac.namespaced }}
```
mmatur, line 2: Could you please split that into 2 files:
- clusterrolebindin.yaml
- rolebindin.yaml

traefik/values.yaml
```diff
@@ -165,6 +165,19 @@ persistence:
...
+# Whether Role Based Access Control objects like roles and rolebindings should be created
+rbac:
+  # If set to false, installs ClusterRole and ClusterRoleBinding so Traefik can be used across namespaces.
+  # If set to true, installs namespace-specific Role and RoleBinding and requires provider configuration be set to that same namespace
+  namespaced: false
+  enabled: true
+
+# Whether to create the ServiceAccount
+serviceAccount:
+  create: true
```
mmatur, line 177: To me this property is not necessary.

You can handle the use case only with the name. 

If the property name is not defined or empty create the service account. otherwise reference an existing service account. 

WDYT?

steven-sheehy: `serviceAccount.create` is a standard option supplied by `helm create`, hence why I kept it as most people are familiar with it and how to use it. Removing `serviceAccount.create` would disallow the chart creating a service account with a custom name. 

That's probably not a common use case though, so I can remove it. I'll keep the name as `serviceAccount.name` instead of `serviceAccountName` to allow the possibility to add create or other options in the future as the need arises.

traefik/templates/rbac/clusterrolebinding.yaml
steven-sheehy: Done

traefik/templates/rbac/clusterrole.yaml
steven-sheehy: Done

steven-sheehy:
Any update @mmatur ?

PR #192, opened 2020-06-03
Add pod disruption budget

ncmans:


traefik/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.podDisruptionBudget -}}
```
steven-sheehy, line 1: Would recommend an explicit `podDisruptionBudget.enabled` field so users can turn it on/off for different environments without removing default config for min/max

ncmans: Done

SantoDE:
Hi @ncmans,

we currently can't merge your PR, because you did not allow modifactions by the maintainers team. Could you please change that?

ncmans:
> Hi @ncmans,
> 
> we currently can't merge your PR, because you did not allow modifactions by the maintainers team. Could you please change that?

As the fork is owned by an organization, I don't have the option to allow upstream access to PRs.

mmatur:
@ncmans could you please bump the chart version? I will merge the PR manually.

PR #193, opened 2020-06-09
Fix read-only /tmp

### LINKED ISSUE #178: Buffering does not work with default security context
Hi,

I recently updated to the 8.1.3 chart version but keep Traefik image version at 2.1.2.
After that, I started to get errors like these:
```
time="2020-05-11T05:43:37Z" level=error msg="open /tmp/temp-multibuf-419673259: read-only file system"                                                                                         
time="2020-05-11T05:48:25Z" level=error msg="vulcand/oxy/buffer: error when reading request body, err: open /tmp/temp-multibuf-063517198: read-only file system" 
```

It's because the `/tmp` directory is read-only by default:
```
securityContext:
  capabilities:
    drop: [ALL]
  readOnlyRootFilesystem: true
  runAsGroup: 65532
  runAsNonRoot: true
  runAsUser: 65532
```

I opened a related issue here: https://github.com/containous/traefik-library-image/issues/79, but I'm not sure that would solve the problem, don't know that a docker volume count as part of the root filesystem or not.
I think the helm chart should have an option to mount the tmp dir as an emptydir at least.
Or add the ability to mount any volume from values, not just Secrets and Configmaps.

steven-sheehy:
Fixes #178 by adding a `/tmp` as an emptyDir to allow read/write

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.6.0
+version: 8.6.1
 appVersion: 2.2.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -96,6 +96,8 @@ spec:
             {{- if .Values.persistence.subPath }}
             subPath: {{ .Values.persistence.subPath }}
             {{- end }}
+          - name: tmp
+            mountPath: /tmp
           {{- range .Values.volumes }}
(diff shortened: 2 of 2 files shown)


traefik/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.4.1
+version: 8.4.2
```
SantoDE, line 3: Can you set a proper version? We had some PR's merged before so we should bump that version as well

steven-sheehy: Done

PR #196, opened 2020-06-10
Template custom volume name

steven-sheehy:
This passes `volumes[].name` through tpl so it can be templated. This is necessary because the extra volumes name is either a ConfigMap or Secret and those usually contain the dynamic Helm release name. For example:
```yaml
  volumes:
    - name: '{{ printf "%s-traefik" .Release.Name }}'
      mountPath: /etc/traefik/
      type: configMap
```

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.6.0
+version: 8.6.1
 appVersion: 2.2.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -96,8 +96,9 @@ spec:
             {{- if .Values.persistence.subPath }}
             subPath: {{ .Values.persistence.subPath }}
             {{- end }}
+          {{- $root := . }}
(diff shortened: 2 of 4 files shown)


steven-sheehy:
It is required if you have more than one Traefik deployment, especially in Helm 2 where releases are global. It is best practice to include the release name in all Kubernetes objects to avoid conflicts on install. Right now the chart requires a static value like `traefik` for volume name so you can't install multiple Traefiks. We had to revert from using a ConfigMap volume to configure Traefik to using the CLI arguments until this could be fixed.

SantoDE:
Hello @steven-sheehy,

sorry for the long no reply. I agree, it makes sense to have it. Can you rebase to the current state?

steven-sheehy:
Thanks @SantoDE. I've fixed the conflicts.

steven-sheehy:
Fixed conflict

SantoDE:
Thanks! Sorry for the hiatus.

One question: Is that a potential breaking change? Like, will old volume names still work? It will, right? Just wondering about the helm-chart version.

steven-sheehy:
Yes, it is backwards compatible with non-templated volumes. I've added a test for that scenario.

SantoDE:
Thank you @steven-sheehy

PR #199, opened 2020-06-13
Add existing PVC support(#158)

### LINKED ISSUE #158: Support persistence.existingClaim
Hi, it would be nice to support `existingClaim` for persistence, so we can rebind any existing volume to the helm deployment.

The current workaround is to pre-provision a PV with a `claimRef` matching the one that the helm release will generate (and the size).

walbertus:
Adding support to use existing PVC by specifying persistence.existingClaim value.

This PR will closes #158

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.5.0
+version: 8.6.0
 appVersion: 2.2.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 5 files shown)


traefik/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.4.1
+version: 8.5.0
```
SantoDE, line 3: There's another PR merged now, so I guess that would need to be 8.6.0?

walbertus: Updated to `8.6.0`

PR #202, opened 2020-06-17
Add extra containers and emptydir volume

djbulsink:
This PR adds:
* the ability to use extraContainers and extraInitContainers to traefik
* the ability to use emptyDir volumes. 

The purpose of the emptyDir volume is to share a volume with an extraContainer. 

## extraContainers and extraInitContainers

Example:
```yaml
extraContainers:
  - name: echo
     image: busybox
     imagePullPolicy: Always
     args:
       - echo
       - hello
extraInitContainers:
  - name: echo
    image: busybox
    args:
      - sleep
      - 1 
```

## EmptyDir volume

Example:
```yaml
volumes: 
- name: configs
   mountPath: "/config"
   type: emptyDir
```

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.7.0
+version: 8.8.0
 appVersion: 2.2.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.5.0
+version: 8.7.0
 appVersion: 2.2.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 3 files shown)


SantoDE:
Hello @djbulsink,

thanks for your interest in the project and your contribution. 

Prior to a review: What is your exact use-case here? For what sort of things you need an extra sidecar container? I'm not sure about that for the Traefik Helm chart.

Thanks :-)

djbulsink:
Hi @SantoDE ,

We like to add access logs to an emptyDir volume and send the access logs to a central logging server by the sidecar. It would be possible to use the stdout for this, but we like to separate the access logs stream from the application logs.

djbulsink:
@SantoDE ,

Can you  please have a look at this PR?

Regards. djbulsink

SantoDE:
Hello @djbulsink,

yes of course. Sorry, things were a little wild.

To be honest, I'm not too sure about your request. While I can see your use-case, we try to keep the helm-chart simple for the majority of use-cases for our user-base.

Your request seems to be "not that general", at least in terms of feedback we received so far.  Therefore, we will decline that PR.

If you want modifications like that, you can either fork our chart or use helm dependencies to enhance your version of the chart.

Thanks for your interest in the project :-)

Issue #186, opened 2020-05-27
Expose our dashboard publicly (with basic authentication)


SantoDE:
Currently, by doing a basic helm installation, the dashboard is reachable only via a port forward, as for example explained [here](https://containo.us/blog/traefik-v2-helm/).

We want to introduce a second option, to have the Dashboard exposed publicly (without the need of a port forward).

While that could introduce a security whole, we want to protect the exposed dashboard by default with basic authentication. The credentials can either be provided prior to the installation, or otherwise a random pair will be generated.

fireba11:
Would be nice to have that in the helm chart already.

travisghansen:
I propose we create specific services for the dashboard along with an *real* ingress (which you could annotate for auth or whatever). This would allow you to bind the dashboard to a different ingress controller altogether (ie: something that isn't publicly accessible for example etc).

I proposed this previously and it got shot down but it may be time to revisit.

I have it implemented in my fork here: https://github.com/travisghansen/traefik-helm-chart

jbergler:
A less breaking change might be to allow adding middleware to the ingressroute via values.yaml

PR #203, opened 2020-06-19
Add basic auth for the exposed dashboard

walbertus:
Users will be able to specify the username and password used to login to the dashboard.
Setting the username and password is mandatory when the user chooses to expose the dashboard.

This PR will help #186

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,32 +57,24 @@ kubectl apply -f traefik/crds
 
 ### Exposing the Traefik dashboard
 
-This HelmChart does not expose the Traefik dashboard by default, for security concerns.
-Thus, there are multiple ways to expose the dashboard.
-For instance, the dashboard access could be achieved through a port-forward :
-
-```
-kubectl port-forward $(kubectl get pods --selector "app.kubernetes.io/name=traefik" --output=name) 9000:9000
-```
-
-Another way would be to apply your own configuration, for instance,
-by defining and applying an IngressRoute CRD (`kubectl apply -f dashboard.yaml`):
+This Helm Chart does not expose the Traefik dashboard by default, for security concerns.
+To expose the Traefik dashboard you need to configure it on `values.yaml`:
 
 ```yaml
-# dashboard.yaml
-apiVersion: traefik.containo.us/v1alpha1
-kind: IngressRoute
-metadata:
-  name: dashboard
-spec:
-  entryPoints:
-    - web
-  routes:
-    - match: Host(`traefik.localhost`) && (PathPrefix(`/dashboard`) || PathPrefix(`/api`))
-      kind: Rule
-      services:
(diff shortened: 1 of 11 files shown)


SantoDE:
Hello @walbertus,

Thanks for you work so far :-)

Let me know once you have a PR state we can test. Should I test the current version already?

walbertus:
Hi @SantoDE 

You can test this PR now but you need to specify the password, as there's no mechanism to print generated password yet.

walbertus:
Updates:
 - Dashboard password is now mandatory

Issues:
 - Basic auth uses MD5, SHA1, or BCrypt for encryption ([here](https://docs.traefik.io/middlewares/basicauth/#general)). Helm support some methods related to cryptography ([here](https://helm.sh/docs/chart_template_guide/function_list/#cryptographic-and-security-functions)).
These methods only intersect at `htpasswd`, which failed on pipeline with error `function "htpasswd" not defined`.

Do you have any suggestion about this @SantoDE ?

SantoDE:
Hello @walbertus,

sorry for answering a little late :)

Hmm, not sure about your issue tough. Maybe it would be an option to use a sha1sum on the provided password and run this through a b64 enc before writing to the secret? Does it make sense?

walbertus:
Hello @SantoDE,

That's a nice idea, user can generate their own credentials using whatever encryption they choose.

Will update this

SantoDE:
@walbertus can I give it another test? Did you found a way for the credential "delivery?

SantoDE:
Hey @walbertus,

any updates?

walbertus:
Hi @SantoDE sorry I miss your comment
User-generated credentials are the only way I think, do you have any other suggestion?

Will resolve the conflict after work

CuBiC3D:
Hey I kinda think, that `userData` sounds a little bit misleading as it suggests to provide some data by the user and not user auth related settings. And generated secrets by Helm should also provide an option to be self provided by secret name, since people use GitOps nowdays with sealed secrets or SOPS and really don't like to have those things be gitted as values.

SantoDE:
I agree with you @CuBiC3D, at least if I understood correctly. Any ideas how we wanna move that forward?

CuBiC3D:
I would just suggest to rename `userData` to something like `usersString` or `users` (one could probably expect an array here) and create `useExistingSecret` and `secretName` values to define those (first one could be omitted since an empty secret name can also be used as a condition).

walbertus:
I agree `userData` is misleading. `users` seems more appropriate since Traefik also uses [users](https://docs.traefik.io/middlewares/basicauth/#users).

We can also use the `useExistingSecret` filled with the secret name

WDYT?

SantoDE:
Sounds good in theory. Eager to see it :)

PR #209, opened 2020-06-25
Update values.yaml

tehmoon:
Fix typo


SantoDE:
Sorry, I did overlook that this requires a version bump. Can you please do a patch version so we can merge? :)

PR #212, opened 2020-06-28
run init container to set proper permissions on volume

### LINKED ISSUE #217: Missing documentation / notifications about ENV parameters required for `make unit-test`
`make unit-test` requires a couple of ENV variables to be set which don't appear to be documented anywhere.

if these aren't set, the helm plugin install doesn't populate the values required to download the binary for the given platform:

```
plugin_install.go:73: [debug] loading plugin from /home/kcrawley/.local/share/helm/plugins/helm-unittest
plugin.go:60: [debug] running install hook: /bin/sh -c $HELM_PLUGIN_DIR/install-binary.sh
Downloading https://github.com/rancher/helm-unittest/releases/download//helm-unittest-linux-.tgz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     9  100     9    0     0     32      0 --:--:-- --:--:-- --:--:--    32
tar: This does not look like a tar archive
```

by setting the ENV variables:

```
$export CATTLE_HELM_UNITTEST_VERSION=v0.1.6-rancher1
$export ARCH=amd64
```
and running the `make unit-test`, the install seems to work and the tests are run:

```
➜  traefik-helm-chart git:(master) make unit-test
== Checking global requirements...
== Checking that plugin helm-unittest is available...
[debug] updating https://github.com/rancher/helm-unittest
[debug] symlinking /home/kcrawley/.cache/helm/plugins/https-github.com-rancher-helm-unittest to /home/kcrawley/.local/share/helm/plugins/helm-unittest
plugin_install.go:73: [debug] loading plugin from /home/kcrawley/.local/share/helm/plugins/helm-unittest
plugin.go:60: [debug] running install hook: /bin/sh -c $HELM_PLUGIN_DIR/install-binary.sh
Downloading https://github.com/rancher/helm-unittest/releases/download/v0.1.6-rancher1/helm-unittest-linux-amd64.tgz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   642  100   642    0     0   1725      0 --:--:-- --:--:-- --:--:--  1725
100 4799k  100 4799k    0     0  4195k      0  0:00:01  0:00:01 --:--:-- 8052k
Preparing to install into /home/kcrawley/.local/share/helm/plugins/helm-unittest
helm-unittest installed into /home/kcrawley/.local/share/helm/plugins/helm-unittest
 ....
Installed plugin: unittest
== plugin helm-unittest is ready
== Unit Testing Chart...

### Chart [ traefik ] ./traefik

 PASS  Pod configuration        traefik/tests/pod-config_test.yaml
 PASS  Traefik configuration    traefik/tests/ports-config_test.yaml
 PASS  RBAC configuration       traefik/tests/rbac-config_test.yaml
 PASS  Service configuration    traefik/tests/service-config_test.yaml
 PASS  Main Container configuration     traefik/tests/container-config_test.yaml
 PASS  default install  traefik/tests/default-install_test.yaml
 PASS  Deployment configuration traefik/tests/deployment-config_test.yaml
 PASS  PodDisruptionBudget configuration        traefik/tests/poddisruptionbudget-config_test.yaml
 PASS  Traefik configuration    traefik/tests/traefik-config_test.yaml

Charts:      1 passed, 1 total
Test Suites: 9 passed, 9 total
Tests:       43 passed, 43 total
Snapshot:    0 passed, 0 total
Time:        67.2957ms

== Unit Tests Finished...
```

StarpTech:
I'm not 100% sure what's the root issue but this change will fix issues with persistent volume when the `acme.json` file can't be open due to too open rights [here](https://github.com/containous/traefik/blob/cb1d0441e925ddc10c0071df33c9c5f35bb4c58b/pkg/provider/acme/local_store_unix.go#L31).

Details:

- The traefik process starts with a umask of 002
- The cloud volume is mounted with `drwxrws---    2 root     65532         4096 Jun 28 10:13 data`

I observed the behavior that all files under `/data` are mounted with umask `0022`. This happens after every traefik container restart.

Fix https://github.com/containous/traefik/issues/6972
Fix #217 

Related:
- https://stackoverflow.com/questions/57566773/how-to-change-the-umask-when-mounting-volumes-in-kubernetes-pods
- https://medium.com/faun/digitalocean-kubernetes-and-volume-permissions-820f46598965


chris13524:
This PR solves the issue for me.

notsureifkevin:
Hardcoding this init container is somewhat presumptuous, perhaps the operator would like to know if the permissions on that file have changed, or perhaps they'd like to introduce another init-container to check for dependencies prior to starting up Traefik?

I recommend we allow consumers of the helm chart, at a minimum, to implement their own init containers -- if the community feels the permission reconciliation belongs here, we should allow the behavior to be toggled on/off, and certain parameters are configurable (image, command).

WDYT?

StarpTech:
Great suggestion but in this issue I'd like to address only the specific issue because it seems to affect everybody with custom volumes. There is no workaround available.

notsureifkevin:
Would introducing the ability to define `initContainers` in values.yaml be sufficient to provide the functionality needed yet offer the flexibility for users to define their own implementations for this (and other concerns)?

eg:

```
      {{- with .Values.initContainers }}
      initContainers:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

StarpTech:
> Would introducing the ability to define initContainers in values.yaml be sufficient to provide the functionality needed yet offer the flexibility for users to define their own implementations for this (and other concerns)?

I agree and we should add the proposed `initContainer` as default as long as no other solution exists.

StarpTech:
@notsureifkevin please recheck. The CI configuration seems broken.

notsureifkevin:
@StarpTech Thanks for the change. I believe the reason why you're failing the build is that the chart version needs a bump. https://github.com/containous/traefik-helm-chart/blob/master/traefik/Chart.yaml#L3 will probably need to go up a minor for the rest of the checks to run/pass.

WRT to you change, I think this is the right direction. I'd be interested in hearing the maintainers thouhgs on whether `initContainers` value should be under `deployment`, along with the other configuration options for deployments. In addition, whether or not we should include the permission reconcillation enabled / commented. Eg:

```
deployment:
  enabled: true
  # Number of pods of the deployment
  replicas: 1
  # Additional deployment annotations (e.g. for jaeger-operator sidecar injection)
  annotations: {}
  # Additional pod annotations (e.g. for mesh injection or prometheus scraping)
  podAnnotations: {}
  # Operator defined initContainers to run prior to Traefik startup (e.g. acme permission modification)
  initContainers: {}
  #- name: acme-permissions
  #  image: busybox:1.31.1
  #  command: ["sh", "-c", "chmod -Rv 600 data/acme.json"]
  #  volumeMounts:
  #  - name: data
  #    mountPath: /data
```

The reason why I commented out the example is since this is a workaround (providers should probably leave file permissions alone, or at least provide a way to configure them), operators should be aware of this behavior prior to opt-ing into the behavior.

StarpTech:
> I'd be interested in hearing the maintainers thouhgs on whether initContainers value should be under deployment, along with the other configuration options for deployments

When `Deployment` is mapped 1to1 to the `Deployment` Kubernetes resource I agree but there are already sections like `rollingUpdate`, `volumes` which are handled separately. I think for the sake of simplicity I prefer a flat structure.

SantoDE:
Can you provide an example of what you think of @StarpTech?

notsureifkevin:
> When `Deployment` is mapped 1to1 to the `Deployment` Kubernetes resource I agree but there are already sections like `rollingUpdate`, `volumes` which are handled separately. I think for the sake of simplicity I'd prefer a flat structure.

It's possible those made it past review, but since this is 1:1 mapped to the `Deployment` manifest, does it not makes sense to put it there? What other reason would it make sense to share that configuration with another template?

Also, I feel this should be commented out by default, since persistence is `false` by default and as such no reconciliation would be necessary until the operator enables persistence and uses a provider which overwrites permissions on mount.

StarpTech:
I think this discussion goes a little bit away from the initial issue. I'd like to land this fix so others won't run into it. Let's discuss it afterward. I think a discussion of why that permission fix is necessary is much more important.

traefik/templates/deployment.yaml
```diff
@@ -52,6 +52,10 @@ spec:
       serviceAccountName: {{ include "traefik.serviceAccountName" . }}
       terminationGracePeriodSeconds: 60
       hostNetwork: {{ .Values.hostNetwork }}
+      initContainers:
```
SantoDE, line 55: Shoudln't line 55 and 56 be swapped? Like, the initContainer should only be there if "configured"?

StarpTech: Good catch.

StarpTech:
@SantoDE done.

traefik/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.7.2
+version: 8.8.1
```
SantoDE, line 3: it should be 8.9.0, no? It's adding new functionality so for me, its a minor and not a patch

notsureifkevin:
Ideally we'll have some simple tests for this, which can be achieved similarly to the ones I just added for `additionalContainers`, for reference: https://github.com/containous/traefik-helm-chart/pull/216/files

If you have any problems running those tests locally, be sure to set the ENV params as noted in https://github.com/containous/traefik-helm-chart/issues/217

Thank you for the contribution and for working with us on getting this merged.

~kc

traefik/Chart.yaml
StarpTech: Minor was already increased in this PR from 8.7.0 to 8.8.0 due to this new feature. I increase the patch just to fix the ci :smile:

StarpTech:
@notsureifkevin next please make a review :smile:

traefik/Chart.yaml
StarpTech: Updated due to master

StarpTech:
What's missing?

SantoDE:
Can you fix your conflicts and do a quick rebase? then I'll take a look :)

StarpTech:
Done.

notsureifkevin:
LGTM

PR #215, opened 2020-06-30
added additionalContainers option to chart

### LINKED ISSUE #213: Support for additional containers
Hi! First off, great work on the Helm chart, I've used it and it works really well!

It would be great to allow to add additional containers to the Traefik deployments. For my use-case, I want to add a side-car that collects the Prometheus metrics from Traefik and sends them somewhere else.

To do this, I think a simple solution is to add an `additionalContainers` key to the `values.yaml` file, and just append its content to the `containers` of the pods spec in the Deployment.

notsureifkevin:
Fixes #213 
Addresses "additionalContainer" for https://github.com/containous/traefik-helm-chart/pull/202

This follows a fairly common pattern across helm charts exposing the ability for operators to define sidecars and init containers (which is being worked on in https://github.com/containous/traefik-helm-chart/pull/212)

Re: PR#202, emptyDir can be handled shared between `/tmp`, or possibly through `additionalVolumes`, which is another common approach for exposing common administrative tasks to chart consumers.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.7.3
+version: 8.8.0
 appVersion: 2.2.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.7.2
+version: 8.7.3
 appVersion: 2.2.1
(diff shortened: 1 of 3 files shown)


traefik/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.7.2
+version: 8.7.3
```
SantoDE, line 3: Should be a minor bump and not a patch bump, no?

notsureifkevin: yup

PR #216, opened 2020-07-02
Additional container fix

notsureifkevin:
* added tests to cover additional container spec
* fixed using a table instead of a map for `additionalContainers`

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -15,7 +15,7 @@ deployment:
   # Additional pod annotations (e.g. for mesh injection or prometheus scraping)
   podAnnotations: {}
   # Additional containers (e.g. for metric offloading sidecars)
-  additionalContainers: {}
+  additionalContainers: []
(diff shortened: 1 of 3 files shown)


traefik/values.yaml
```diff
@@ -14,6 +14,8 @@ deployment:
   annotations: {}
   # Additional pod annotations (e.g. for mesh injection or prometheus scraping)
   podAnnotations: {}
+  # Additional containers (e.g. for metric offloading sidecars)
+  additionalContainers: []
```
mmatur, line 18: An example would help the end user.

mmatur: And you are breaking with this change. You can't bump a minor version

notsureifkevin: populating this as a tables would have never generated a sucessful install, only `maps` would be sucessful. as such, this may qualify as a patch.

Issue #188, opened 2020-05-27
Redirect from HTTP to HTTPS globally


SantoDE:
Currently, the only way to define a global HTTP to HTTPS redirect with the helm chart is to use the `additionalArguments` like that `--entrypoints.web.http.redirections.entryPoint.scheme=https`

We can hide that behind a helm configuration to enforce that for the user automatically

Partially fixes #185

PR #219, opened 2020-07-10
Fix additional /tcp path on redirect https globally using additionalArguments

### LINKED ISSUE #208: Not able to redirect from http to https
is there any trick i can do to make http to https redirect work?

added  --entrypoints.web.http.redirections.entryPoint.scheme=https in additionalArguments but it didn't work.

but when i add these two arguments while deployment from helm chart:
--entrypoints.web.http.redirections.entryPoint.to=websecure --entrypoints.websecure.http.tls.certResolver=default

but when i browsed the URL it redirected me to example.com:8443/tpc.

is this a bug in traefik ingress controller or am i m doing it wrong?

akhfa:
Hi

This should fix #208 

I'm pretty sure that this also will be related with #188 

Additional details explained in https://github.com/containous/traefik-helm-chart/issues/208#issuecomment-656673347

This PR expected to be failed when the test include `/tcp` in these lines
https://github.com/containous/traefik-helm-chart/blob/master/traefik/tests/ports-config_test.yaml#L22
https://github.com/containous/traefik-helm-chart/blob/master/traefik/tests/ports-config_test.yaml#L43
https://github.com/containous/traefik-helm-chart/blob/master/traefik/tests/ports-config_test.yaml#L63
https://github.com/containous/traefik-helm-chart/blob/master/traefik/tests/ports-config_test.yaml#L89
https://github.com/containous/traefik-helm-chart/blob/master/traefik/tests/ports-config_test.yaml#L117

Should I remove the `/tcp` part in the test, or maybe it has some functionalities?
Because on my case (and maybe some other people case), this `/tcp` add additional path when we need redirect 🤔 
And this `/tcp` thing also doesn't exist on this [Kubernetes and Let's Encrypt guides](https://docs.traefik.io/user-guides/crd-acme/#deployments)

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.9.0
+version: 8.10.0
 appVersion: 2.2.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -114,7 +114,7 @@ spec:
           {{- end }}
           {{- end }}
           {{- range $name, $config := .Values.ports }}
-          - "--entryPoints.{{$name}}.address=:{{ $config.port }}/{{ default "tcp" $config.protocol | lower }}"
+          - "--entryPoints.{{$name}}.address=:{{ $config.port }}"
           {{- end }}
           - "--api.dashboard=true"
           - "--ping=true"


SantoDE:
The `/tcp` is just on the entrypoint and thus, required. Traefik can have entrypoints of either `TCP` or `UDP` and that is to indicate, that this should be a TCP port, so I'm not sure if this is a valid fix to be honest. 

Therefore, I think we should close the PR and keep on discussing in the issue to see (hopefully) see another cause for your issue. I'll take a look today.

akhfa:
So, do you find how to fix this issue?

Does the traefik's code itself that need to be fixed? I'm pretty sure that this issue happened to people that install traefik using helm, at least for traefik version 2.2.1 🤔 

My question is why traefik add additional `/tcp` to the host address when we just need it for the TCP UDP identifier? 🤔 
And how about the [Kubernetes and Let's Encrypt guides](https://docs.traefik.io/user-guides/crd-acme/#deployments) docs that doesn't include neither `/tcp` or `/udp` ? Is the default value set to tcp? 🤔

SantoDE:
👋 

See my comment on #208

PR #223, opened 2020-07-22
feat: added lifecycle hook to container

amalucelli:
This change adds the lifecycle to the container, that can be useful with the `preStop` hook, used to avoid aborting ongoing requests ([here is why](https://blog.sebastian-daschner.com/entries/zero-downtime-updates-kubernetes)). This is also useful as `traefik` supports custom values for [graceful timeouts](https://docs.traefik.io/routing/entrypoints/#lifecycle).

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.8.0
+version: 8.10.0
 appVersion: 2.2.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.7.2
+version: 8.8.0
 appVersion: 2.2.1
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -90,6 +90,10 @@ spec:
         securityContext:
           {{- toYaml . | nindent 10 }}
         {{- end }}
+        {{- with .Values.lifecycle }}
+        lifecycle:
+          {{- toYaml .Values.lifecycle | nindent 10 }}
+        {{- end }}
         volumeMounts:
           - name: data
             mountPath: {{ .Values.persistence.path }}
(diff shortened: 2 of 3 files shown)


SantoDE:
Hello @amalucelli,

thanks for your PR.  Currently, I'm just a bit unsure how this would look in practice? Like what part of `yaml` would you put? I can see that beeing used for graceful timeouts, but that's not what this PR is aiming for, right? Can you explain me more?

Thank you very much. :-)

amalucelli:
Hi @SantoDE, thank you for your response.

I was reviewing this implementation, and I think that maybe I was using it wrongly.

My problem is with very long-lived connections, but I just realized that `SIGTERM` is only triggered after the `preStop` hook, so any kind of hook in that phase is not helping at all.

I think the right way would be to increase the `lifeCycle.graceTimeOut` on `traefik` and increasing the `terminationGracePeriodSeconds` on `Pod` specification also.

Currently I see a `60` seconds for the `terminationGracePeriodSeconds` in this chart, do you think you could accept a custom value here? I could change this implementation.

https://github.com/containous/traefik-helm-chart/blob/37d8ea6c40158f51d3d087c82ecaf7c31b546673/traefik/templates/deployment.yaml#L53

SantoDE:
Yes. Sounds like the better approach. We should close that PR tough and do it another PR. Do you agree?

SantoDE:
Hey @amalucelli,

I'll close that PR for now, as it looks inactive. If you want, you can implement the `terminationGracePeriodSeconds` change in a seperate PR.

Thanks for your work!

mars64:
Stumbled here researching ways to run traefik behind an nlb-ip in eks via the new [aws-load-balancer-controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/), in a way that doesn't drop traffic when traefik rotates faster than the hideously slow target group healthcheck endpoints update (given that [the in-tree provider is broken until 1.20 ](https://github.com/kubernetes/kubernetes/issues/82595), aws-load-balancer-controller doesn't currently provide for `nlb` and I need more than `80` and `443` so ALB is right out). At one point I sought to implement this same `lifecycle` option to artificially delay terminating traefik pods.

Turns out, I needed to look into [aws-load-balancer-controller's pod readiness gate](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/pod_readiness_gate/). This effectively causes the pod rotation to wait based on load balancer target health (in my testing this is something around 4mins per pod). It still seems a bit of a wonky hack to me, but the tradeoff works for my situation. I look forward to a legit fix in `1.20`.

Issue #185, opened 2020-05-27
Implement methods for simple use cases


SantoDE:
Currently, the helm chart is very flexible and can be seen as a scaffolding base for further configuration and or enhancements by our users.

However, we will try to implement simple wrappers inside that helm chart for the following simple use cases:

- [Expose our dashboard publicly (with basic authentication)](https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#publish-and-protect-traefik-dashboard-with-basic-auth)
- ~~Provide a "default" Certificate for TLS~~
- ~~Redirect from HTTP to HTTPS globally~~
- ~~Set a log Level~~
- ~~Setup Lets Encrypt (ACME) with either HTTP or TLS challenge~~

This will not be a replacement for the current flexibility but rather an addition to it, to ease the installation process for the use-cases mentioned above.

mloiseleur:
@SantoDE I guess we're good now. 
Please let me know if you think we missed something.

PR #224, opened 2020-07-31
Add logging, and http redirect config

### LINKED ISSUE #188: Redirect from HTTP to HTTPS globally

### LINKED ISSUE #189: Set a log Level
Currently, the only way to define a `logLevel` the helm chart is to use the additionalArguments like that `--logs.level=DEBUG`

We can hide that behind a helm configuration to enforce that for the user automatically

Partially fixes #185

DirtyCajunRice:
Allow configuration of metrics and http redirects. 
http redirects fixes #188 and works on #185
logging fixes #189 and works on #185
~~metrics is added as a new feature without prior issue~~

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.11.0
+version: 8.11.1
 appVersion: 2.2.8
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.12.1
+version: 8.13.0
 appVersion: 2.2.8
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.9.1
+version: 8.10.0
 appVersion: 2.2.5
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -1,7 +1,7 @@
 # Default values for Traefik
 image:
   name: traefik
-  tag: 2.2.5
+  tag: 2.2.8
 
 #
 # Configure the deployment
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -88,9 +88,6 @@ redirections:
 logs:
   # Traefik logs concern everything that happens to Traefik itself (startup, configuration, events, shutdown, and so on).
   general:
-    # By default, the logs are written to the standard output.
-    # You can configure a file path instead using the filePath option.
-    # filePath: "/path/to/traefik.log"
     # By default, the logs use a text format (common), but you can
     # also ask for the json format in the format option
     # format: json
@@ -99,9 +96,6 @@ logs:
   access:
     # To enable access logs
     enabled: false
-    # By default access logs are written to the standard output.
-    # To write the logs into a log file, use the filePath option.
-    # filePath: "/path/to/traefik.access.log"
     # By default, logs are written using the Common Log Format (CLF).
     # To write logs in JSON, use json in the format option.
     # If the given format is unsupported, the default (CLF) is used instead.
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -78,10 +78,7 @@ volumes: []
 redirections:
   web:
     enabled: false
-    to: websecure
-    scheme: https
-    permanent: true
-    priority: 1
+    port: 443
 
 # Logs
 # https://docs.traefik.io/observability/logs/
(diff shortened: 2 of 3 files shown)


SantoDE:
Hello @DirtyCajunRice,

thanks for your PR and your hard work.

This PR is basically a PR for 3 different topics at once, which will need some sorting. 

I really like the part about `redirects, this looks good.

For the logs, I'm a bit unsure. It offers a lot of different configuration options (which in itself is great), but I'm a bit hesitating as it could make stuff more complex at the beginning. We were only thinking about the level. Maybe the format. But I'm unsure about all the "edge case" stuff like filtering, buffersize and so on?

I'm also not sure about the metrics part. We try to keep that helm-chart to solve most base requirements, that looks (to me) more advanced as well. WDYT?

DirtyCajunRice:
So I can absolutely remove the extra config options, but I strongly believe that if you are going to offer an option with option, and that option has subset options, you are doing a "disservice" by only offering "some" of those options. Basically a continuity thing i guess? 

As for the metrics specifically, it is part of traefik so- in my optinion- it should be here. If i was adding- for example- a ServiceMonitor which, although is prometheus, is not directly tied to traefik it would/should be separate. 

There are lots more traefik 2.0 config options i would like to add but this was already a pretty large mr so i stopped here haha.

Yall are the final decision on all things traefik-helm-chart so just tell me what you will allow and ill modify.

SantoDE:
Hello @DirtyCajunRice,

thanks for your answer.

For logging, I think we can remove the `filePath` (in access and general log) option as it doesn't make lots of sense in a k8s environment, right? I'm still a bit afraid by all that options for filters and stuff. So I might need a second opinion on that :-)

For metrics: I'm a bit afraid by the "complexity" added to the templates. Maybe (for consistency) we could do a bit like in the Maesh helm chart? https://github.com/containous/maesh/blob/master/helm/chart/maesh/templates/mesh/mesh-daemonset.yaml#L208 WDYT?

DirtyCajunRice:
Hey @SantoDE . Ive removed filepath. In my head i was preparing for the argument from some users that would scrape logs via filebeat or similar  that dont scrape pod stdout/err but instead scrape a file in the pod. In the same breath if no one ever uses a feature is it a feature at all? haha. 

As for the metrics section, this is just a bit fancier way to do it. The maesh chart is a bit redundant/verbose. They both accomplish the same thing. If you would like me to switch this to the longer version I can, or if you would like me to make a PR to the maesh chart to update the format i can do that as well. I would just like to get this in sooner rather than later! 

Thank you

traefik/values.yaml
```diff
@@ -72,6 +72,127 @@ volumes: []
 #   mountPath: "/config"
 #   type: configMap
 
+# Port Redirections
+# Added in 2.2, you can make permanent redirects via entrypoints.
+# https://docs.traefik.io/routing/entrypoints/#redirection
+redirections:
+  web:
```
SantoDE, line 79: after some further testing and discussion, we should make that block easier. Just `entryPoint`based redirects (e.g. the `to`) won't work in all cases, as the traefik port is another one thatn the port of the outside service and thus, redirections wont work out.

Can you simplify this block to something, that only `activates` either a boolean to redirect to TLS or to redirect to a given port?

traefik/values.yaml
```diff
@@ -72,6 +72,127 @@ volumes: []
...
+        defaultmode: drop
+        names: {}
+          # Examples:
+          # User-Agent: redact
+          # Authorization: drop
+          # Content-Type: keep
+
+# Metrics
+# https://docs.traefik.io/v2.2/observability/metrics/overview/
+metrics:
```
SantoDE, line 130: Can we have a seperate PR for the metrics block? Looks like we have different approaches of the implementation.

DirtyCajunRice: Ive removed it in this PR and will open that one and a few others once this initial finishes up :)

traefik/values.yaml
DirtyCajunRice: Ive simplified to ignore adding priority/scheme through the chart as it makes sense that those options are probably almost ALWAYS default. Instead of "to" it shows "port" now and explicitly prefixes with a `:` in the templating to enforce.

SantoDE: Thank you! One thing I know realize as you take the entryPoint name in the template.

We could move that part down to the actual port declaration because otherwise, we would have an implicit
relation between the two parts.

Then, we could fallback to the `to` value beeing the name of another port in that block. It could
look like this:

```
ports:
  web:
    port: 8000
    redirectTo: websecure
  websecure:
    port: 8443
    exposedPort: 443
```

With that beeing the case, we could do something like that in the template of the deployment (not tested ;-)) :

```yaml
{{ $toPort := index .Values.Ports $config.redirectTo }}
- "--entrypoints.{{ $name }}.http.redirections.entryPoint.to=:{{ $toPort.exposedPort }}"
- "--entrypoints.{{ $name }}.http.redirections.entryPoint.scheme=https"
```

Once that is resolved (and the comment from above), I think we're good to go :) Do you want me to do the change or can you do it?

DirtyCajunRice: Makes sense. Ill make the changes this afternoon.

traefik/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.12.0
+version: 8.12.1
```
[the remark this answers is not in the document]
DirtyCajunRice: Done!

DirtyCajunRice:
Both done :)

PR #225, opened 2020-08-03
Add dns policy option

vanhtuan0409:
Add `dnsPolicy` option as it is recommended to use `ClusterFirstWithHostNet` when enable `hostNetwork: true`

https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.9.1
+version: 8.9.2
 appVersion: 2.2.5
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.9.2
+version: 8.9.3
 appVersion: 2.2.8
 description: A Traefik based Kubernetes ingress controller
 keywords:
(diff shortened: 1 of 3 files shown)


traefik/templates/deployment.yaml
```diff
@@ -52,6 +52,9 @@ spec:
       serviceAccountName: {{ include "traefik.serviceAccountName" . }}
       terminationGracePeriodSeconds: 60
       hostNetwork: {{ .Values.hostNetwork }}
+      {{- with .Values.deployment.dnsPolicy }}
```
SantoDE, line 55: no values for that?

vanhtuan0409: Hi, default is we don't need to input any config for `dnsPolicy`

Adding `dnsPolicy: ClusterFirst` might break old deployments

That's why I made it as optional configurable

SantoDE: which is fine, but I think we should have at least an example in the `values.yml` even if its commented out. Don't you think so?

vanhtuan0409: @SantoDE Absolutely agree. Added example on `values.yml`

traefik/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.9.2
+version: 8.9.3
```
SantoDE, line 3: Thanks. Now, last step (sorry) - its adds a new functionality and thus needs to be a minor bump on the version :)

vanhtuan0409: @SantoDE no prob. Bumped to 8.11.0.

But shouldn't it be 8.10?

PR #229, opened 2020-08-10
Add hostIp to port configuration

bokysan:
This change is necessary to support configurations where traefik is deployed multiple times on a single node with multiple IP addresses.

This way one ingress controller can listen on public and the other on the private network.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.9.2
+version: 8.10.0
 appVersion: 2.2.8
 description: A Traefik based Kubernetes ingress controller
 keywords:
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -88,6 +88,9 @@ spec:
           {{- if $config.hostPort }}
           hostPort: {{ $config.hostPort }}
           {{- end }}
+          {{- if $config.hostIP }}
+          hostIP: {{ $config.hostIP }}
+          {{- end }}
           protocol: {{ default "TCP" $config.protocol | quote }}
         {{- end }}
         {{- with .Values.securityContext }}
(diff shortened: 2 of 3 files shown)


traefik/templates/deployment.yaml
```diff
@@ -88,6 +88,9 @@ spec:
           {{- if $config.hostPort }}
           hostPort: {{ $config.hostPort }}
           {{- end }}
+          {{- if $config.hostIP }}
+          hostIP: {{ $config.hostIP }}
```
SantoDE, line 92: no exmaple in `values.yaml`?

bokysan: > To be honest, I'm not too sure whether this a use-case mainly considered or not

It might be a bit of an edge use case, but it's still an use case. Here it is:
- I have a server with a public-facing IP/network and a private-facing one
- I would like to have a private ingress controller and a public ingress controller

Without specifiying the `hostIP`, the first controller to start up "sits" on the port and makes it impossible to run the second controller without much magic and fiddling with the iptables manually.

> I don't see that value in the `values.yaml` tough, can you add it please.

Example added to `values.yaml`: https://github.com/containous/traefik-helm-chart/pull/229/files#diff-76ff6debe949af52b1679bce6b97cac0R115

> Also, please rebase
Sorry, onto which branch? My branch is exactly one commit ahead of `containous:master`. Thanks.

traefik/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.9.2
+version: 8.9.3
```
SantoDE, line 3: it needs to be a minor bump, as it adds new functionality.

bokysan:
> Only a version bump left ;-) Thanks so far!

There you go :-)

https://github.com/containous/traefik-helm-chart/pull/229/files#diff-3f1f0a7250b3125519a27749c079b90eR3

PR #242, opened 2020-08-20
feat: Addressing (#220, #233, #234, #235, #236)

### LINKED ISSUE #220: web and websecure entrypoints shouldn't be present in default values.yaml
Hi,

We use this helm chart as a dependency:
```yaml
dependencies:
- name: traefik
  version: 8.8.1
  repository: https://containous.github.io/traefik-helm-chart
```
It means the default `values.yaml` which is present in this repository is applied and `web` and `websecure` entrypoints are created whatever we set in our own `values.yaml`. See [here](https://github.com/containous/traefik-helm-chart/blob/master/traefik/values.yaml#L127).

This is annoying because we want to use port 8000 for another purpose. We can override the port for `web` (and `websecure`) but these entrypoints are there, no matter what and we do not want useless and confusant config.

I understand the aim behind, which is offering a valid and functionnal chart, but it leads to a chart helm which is rigid and not 100% modular.

We can keep those parts as commented for example, but not force them to be set.

### LINKED ISSUE #233: Migrate Chart.yaml to apiVersion: v2
Helm 3 has apiVersion as 2 by default. There are also many automations that decide which binary to use based on the apiVersion.

### LINKED ISSUE #234: Add IngressClass for upcoming 2.3 Release
Lock behind a kube version check of 1.18.x

### LINKED ISSUE #235: Migrate label/selector definitions to new standard
Defined in _helper.tpl and can be added to.

### LINKED ISSUE #236: Remove disabling deployment
As its the entire design of the chart, it seems fruitless

DirtyCajunRice:
Fixes #220 (Allow removal of port config)
Fixes #233 (Migrate chart to v2)
Fixes #234 (Add IngressClass)
Fixes #235 (Migrate labels/selectors to newest helm standard)
Fixes #236 (Removal of disabling deployment)

Also Some cleanup and indent autofixes


DirtyCajunRice:
Commit history is all messed up. re-opening

SantoDE:
And please in multiple PR‘s. That’s a little too much ;-) But thanks for the work ! 👍

Issue #68, opened 2020-01-21
[Feature Request] Add enabled fields


masterada:
Please add service.enabled, deployment.enabled, rbac.enabled and crd.enabled to traefik chart so we can install parts of the chart.

## Use case
We are using pulumi to automatize the install of traefik. We install crd-s in a different project before installing the rest of the chart, because installing crd and the custom resources (eg: an IngressRoute) together causes issues. We have been using the prometheus chart similarly without issues.

mmatur:
@masterada If you want to skip crds installation you can use `--skip-crds` from helm.

```console
--skip-crds                    if set, no CRDs will be installed. By default, CRDs are installed if not already present
```

PR #244, opened 2020-08-20
feat: Remove disabling deployment

### LINKED ISSUE #236: Remove disabling deployment

DirtyCajunRice:
Fixes #236

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -8,7 +8,6 @@ image:
 # Configure the deployment
 #
 deployment:
-  enabled: true
   # Number of pods of the deployment
   replicas: 1
   # Additional deployment annotations (e.g. for jaeger-operator sidecar injection)
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: traefik
-version: 8.13.2
+version: 8.14.0
 appVersion: 2.2.8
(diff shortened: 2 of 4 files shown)


DirtyCajunRice:
Seems like a lot but its just a couple spaces fixed with indenting + the deployment removal.

traefik/values.yaml
```diff
@@ -8,7 +8,6 @@ image:
 # Configure the deployment
 #
 deployment:
-  enabled: true
```
mmatur, line 11: Regarding this PR https://github.com/containous/traefik-helm-chart/pull/110 and this issue https://github.com/containous/traefik-helm-chart/issues/68, could you please keep the `enable` field.

mmatur (changes requested):

traefik/values.yaml
SantoDE: I agree. `enabled` field should be kept.

DirtyCajunRice: The entire premise of #68 is based on a lack of proper helm usage and hook definition. Helm will install CRDs prior to the rest and IngressRoutes only after set properly _even when using helm template_. I strongly disagree with this flag existing at all.

DirtyCajunRice: Ill leave it for now because there are so many things that need to be fixed but i would like to circle back to this later

PR #245, opened 2020-08-20
feat: Move Chart apiVersion: v2

### LINKED ISSUE #233: Migrate Chart.yaml to apiVersion: v2

DirtyCajunRice:
fixes #233

### DIFF
--- a/traefik/templates/service.yaml
+++ b/traefik/templates/service.yaml
@@ -3,7 +3,7 @@
 {{ $tcpPorts := dict }}
 {{ $udpPorts := dict }}
 {{- range $name, $config := .Values.ports }}
-  {{- if eq ($config.protocol | toString) "UDP" }}
+  {{- if eq (toString $config.protocol) "UDP" }}
     {{ $_ := set $udpPorts $name $config }}
   {{- else }}
     {{ $_ := set $tcpPorts $name $config }}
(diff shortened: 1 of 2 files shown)


traefik/templates/service.yaml
```diff
@@ -3,7 +3,7 @@
 {{ $tcpPorts := dict }}
 {{ $udpPorts := dict }}
 {{- range $name, $config := .Values.ports }}
-  {{- if eq ($config.protocol | toString) "UDP" }}
+  {{- if eq (toString $config.protocol) "UDP" }}
```
SantoDE, line 6: what is that change for?

DirtyCajunRice: Once you move to v2 the lint does not like toString being used as a "Piped to" option and wants to you use it with an arg. Error you get is
> Traefik configuration	traefik/tests/ports-config_test.yaml
	- should have a custom port when specified via values
		Error: render error in "traefik/templates/service.yaml": template: traefik/templates/service.yaml:6:32: executing "traefik/templates/service.yaml" at <toString>: wrong number of args for toString: want 1 got 0

They both do the same thiing but this makes the linter happy haha

PR #247, opened 2020-08-20
bug: Check for port config (#220)

### LINKED ISSUE #220: web and websecure entrypoints shouldn't be present in default values.yaml

DirtyCajunRice:
fixes #220

### DIFF
--- a/traefik/templates/service.yaml
+++ b/traefik/templates/service.yaml
@@ -3,7 +3,7 @@
 {{ $tcpPorts := dict }}
 {{ $udpPorts := dict }}
 {{- range $name, $config := .Values.ports }}
-  {{- if eq ($config.protocol | toString) "UDP" }}
+  {{- if eq (toString $config.protocol) "UDP" }}
     {{ $_ := set $udpPorts $name $config }}
   {{- else }}
     {{ $_ := set $tcpPorts $name $config }}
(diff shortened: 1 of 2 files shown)


traefik/templates/service.yaml
```diff
@@ -3,7 +3,7 @@
 {{ $tcpPorts := dict }}
 {{ $udpPorts := dict }}
 {{- range $name, $config := .Values.ports }}
-  {{- if eq ($config.protocol | toString) "UDP" }}
+  {{- if eq (toString $config.protocol) "UDP" }}
```
SantoDE, line 6: also here - what's that for?

DirtyCajunRice: https://github.com/containous/traefik-helm-chart/pull/245#discussion_r474518059

PR #249, opened 2020-08-21
feat: Enable entrypoint tls config

### LINKED ISSUE #248: Enable tls for websecure in arguments
From traefik 2.2,  [the TLS section in a Ingress don't control the activation of the TLS](https://github.com/containous/traefik/pull/6504).

To enable TLS:
* use the TLS configuration on the entry point
* use annotation traefik.ingress.kubernetes.io/router.tls: "true"

I suggest enable tls on the entryp point.
That will be eazier to migrate from other ingress controller to treafik.

For example:
```
additionalArguments:                                                                                                                                                                                               
  - "--entryPoints.websecure.http.tls=true"
```
Or an additional tls config in values.yaml:
```
  websecure:                                                                                                                                                                                                       
    port: 8443                                                                                                                                                                                                     
    # hostPort: 8443                                                                                                                                                                                               
    expose: true                                                                                                                                                                                                   
    exposedPort: 443                                                                                                                                                                                               
    # The port protocol (TCP/UDP)                                                                                                                                                                                  
    protocol: TCP                                                                                                                                                                                                  
    # nodePort: 32443
    tls: true 
```

DirtyCajunRice:
Fixes #248

### DIFF
--- a/traefik/tests/traefik-config_test.yaml
+++ b/traefik/tests/traefik-config_test.yaml
@@ -47,3 +47,32 @@ tests:
       - contains:
           path: spec.template.spec.containers[0].args
           content: "--providers.kubernetesingress.ingressendpoint.publishedservice=foo/bar"
+  - it: should have tls configuration args when specified in values.yaml for port
+    set:
+      ports:
+        websecure:
+          tls:
+            enabled: true
+            options: "foo"
+            certResolver: "certResolverFoo"
+            domains:
+              - main: "sub.domain.tld"
+                sans:
+                  - "alt1.domain.tld"
+                  - "alt2.domain.tld"
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--entrypoints.websecure.http.tls=true"
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--entrypoints.websecure.http.tls.options=foo"
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--entrypoints.websecure.http.tls.certResolver=certResolverFoo"
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--entrypoints.websecure.http.tls.domains[0].main=sub.domain.tld"
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--entrypoints.websecure.http.tls.domains[0].sans=alt1.domain.tld,alt2.domain.tld"
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -197,6 +197,18 @@ ports:
     # The port protocol (TCP/UDP)
     protocol: TCP
     # nodePort: 32443
+    # Set TLS at the entrypoint
+    # https://docs.traefik.io/v2.2/routing/entrypoints/#tls
+    tls:
+      enabled: false
+      options: ""
+      certResolver: ""
+      domains: []
+        # - main: example.com
+        #   sans:
+        #     - foo.example.com
+        #     - bar.example.com
+
 
 # Options for the main traefik service, where the entrypoints traffic comes
 # from.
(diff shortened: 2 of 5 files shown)


SantoDE (changes requested):
hey @DirtyCajunRice,

thanks for your contribution. Can you please try to add some tests?

SantoDE:
Ping @DirtyCajunRice. Any updates? :)

DirtyCajunRice:
@SantoDE  Guess i lied when i said soon haha. Ive caught back up. Here is your requested tests :)

gregberns:
@SantoDE bump - looks like requested changes were made

traefik/values.yaml
```diff
@@ -213,6 +213,18 @@ ports:
     # The port protocol (TCP/UDP)
     protocol: TCP
     # nodePort: 32443
+    # Set TLS at the entrypoint
+    # https://docs.traefik.io/v2.2/routing/entrypoints/#tls
+    tls:
+      enabled: false
+      options: ""
```
SantoDE, line 220: Sorry for the long time no see. Thanks for requesting my review :-)

The options could be a yaml block (https://doc.traefik.io/traefik/https/tls/#tls-options). Can we at least mention it in the example and show it that way?

traefik/tests/traefik-config_test.yaml
```diff
@@ -47,3 +47,32 @@ tests:
...
+              - main: "sub.domain.tld"
+                sans:
+                  - "alt1.domain.tld"
+                  - "alt2.domain.tld"
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--entrypoints.websecure.http.tls=true"
+      - contains:
+          path: spec.template.spec.containers[0].args
```
SantoDE, line 68: Maybe its better to have a "yaml" block in the test?

SantoDE (changes requested):
Just some minor questions :-)

traefik/tests/traefik-config_test.yaml
DirtyCajunRice: If i am not mistaken, this is to define the name of a tls options configuration, not the options themselves. If that is true this is exactly what it should be

traefik/values.yaml
DirtyCajunRice: Per other comment
> If i am not mistaken, this is to define the name of a tls options configuration, not the options themselves. If that is true this is exactly what it should be

That said, it did bring up a good point that there is no way to define a TLSOption CRD. Because of that i have added such a way.

DirtyCajunRice (commented):
After the comment that made it seem as if tls options could be defined as a block in entrypoint, i added the ability to create TLSOption CRDs.
The example in the file shows as:
```yaml
# TLS Options are created as TLSOption CRDs
# https://doc.traefik.io/traefik/https/tls/#tls-options
# Example:
# tlsOptions:
#   default:
#     sniStrict: true
#     preferServerCipherSuites: true
#   foobar:
#     curvePreferences:
#       - CurveP521
#       - CurveP384
tlsOptions: {}
```
and if the example was actually rendered it would show
```yaml
---
# Source: traefik/templates/tlsoption.yaml
apiVersion: traefik.containo.us/v1alpha1
kind: TLSOption
metadata:
  name: default
  labels:
    helm.sh/chart: traefik-9.8.0
    app.kubernetes.io/instance: traefik
spec:
  preferServerCipherSuites: true
  sniStrict: true
---
# Source: traefik/templates/tlsoption.yaml
apiVersion: traefik.containo.us/v1alpha1
kind: TLSOption
metadata:
  name: foobar
  labels:
    helm.sh/chart: traefik-9.8.0
    app.kubernetes.io/instance: traefik
spec:
  curvePreferences:
  - CurveP521
  - CurveP384

DirtyCajunRice:
@SantoDE  new pr with unbroken base repo at #285

PR #253, opened 2020-09-01
Update reference to using kubectl proxy to kubectl port-forward

davidalger:
With the traefik port not-exposed, the `kubectl proxy` command won't work to access it. Rather port-forward should be used to forward port 9000 for access at http://localhost:9000/dashboard/

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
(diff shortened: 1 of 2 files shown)


SantoDE:
Yeah I agree :-) You will need to do a patch bump of the version, so the CI doesn't fail.

davidalger:
@SantoDE So that's why it failed… :D I've updated the PR with a version bump

SantoDE (approved):
LGTM 👼 Thanks!

PR #246, opened 2020-08-20
feat: Add ingressclass

DirtyCajunRice:
Fixes #234

PR #256, opened 2020-09-02
Add missing RBAC permissions

myoung34:
Resolves https://github.com/containous/traefik-helm-chart/issues/254

### DIFF
--- a/traefik/templates/rbac/clusterrole.yaml
+++ b/traefik/templates/rbac/clusterrole.yaml
@@ -47,4 +47,14 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
+      - extensions
+      - networking.k8s.io
+    resources:
+      - ingresses
+      - ingressclasses
+    verbs:
+      - get
+      - list
+      - watch
 {{- end -}}
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.1.0
+version: 9.1.1
 appVersion: 2.2.8
 keywords:
   - traefik


SantoDE:
We should merge that after we officialy bumped the chart with 2.3. Don't you think so?

myoung34:
AFAIK this is a problem in 2.2 as well

SantoDE:
But it can't for the `ingressClass`, as 2.2 is not loading it. Maybe for Ingress with a changed apiGroup tough. Maybe, we could limit the PR to this then? Because, RBAC should be in #254 then I guess. Right?

myoung34:
I'm not sure to be honest, I'm still on 2.3, but I can try to find out for sure this weekend and scope this to #246

SantoDE:
Hey @myoung34,

thanks! Any updates?

traefik/templates/rbac/clusterrole.yaml
```diff
@@ -47,4 +47,14 @@ rules:
...
+  - apiGroups:
+      - extensions
+      - networking.k8s.io
+    resources:
+      - ingresses
+      - ingressclasses
+    verbs:
+      - get
+      - list
+      - watch
```
DirtyCajunRice, line 59: This creates duplication. The addition should just be to the existing apigroups

DirtyCajunRice (changes requested):
This should be removed in favor of augmenting the existing apigroups

https://github.com/traefik/traefik-helm-chart/blob/92517a02f5f07ff89cb1129dd93cb3387cd1d031/traefik/templates/rbac/clusterrole.yaml#L22-L29

would have the `networking.k8s.io` apiGroup added and the `ingressclasses` resource added.

as for 2.3 conversation: If you are running the RC of traefik you need to manually add the apiGroups. The chart should match the release of 2.3 in permissions and would happen at the same time as #246

traefik/templates/rbac/clusterrole.yaml
notsureifkevin: `extensions` -> `ingresses` -> `get`, `watch`, `list` appear to be covered already, however, `networking.k8s.io` is not, and could be added to the existing definition at `apiGroups[1]`

SantoDE:
As #267 is merged, I'll close that one.

Thanks 👍

PR #259, opened 2020-09-10
Add additional volumes for init and additional containers

ch4nn0n:
This increases the flexibility of the init and additional containers, I ran into the issue where I needed to mount a volume that was not a configMap or a secret to a sidecar.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.5.2
+version: 9.6.0
 appVersion: 2.3.1
 keywords:
   - traefik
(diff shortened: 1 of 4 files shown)


SantoDE:
hey hey,

thanks for your contribution. Not sure I understand the use-case tough. You have a traefik sidecar container, which needs to mount an additional volume which is not a config or a secret? Do I understand that correctly? How would a configuration into `values.yaml` be looking?

ch4nn0n:
Yes, I would like to mount a volume of type `hostPath` to a container running as a sidecar to Traefik.

Here is an example of what would be in a `values.yaml`:
```
deployment:
  additionalContainers:
    # https://docs.datadoghq.com/developers/dogstatsd/unix_socket/?tab=host
    - name: socat-proxy
      image: alpine/socat:1.0.5
      args: ["-s", "-u", "udp-recv:8125", "unix-sendto:/socket/socket"]
      volumeMounts:
        - name: dsdsocket
          mountPath: /socket
  additionalVolumes:
      - name: dsdsocket
        hostPath:
          path: /var/run/statsd-exporter
```

SantoDE (approved):
LGTM 👼

PR #267, opened 2020-09-24
Release Traefik 2.3

SantoDE:
This PR updates Trafik to 2.3, including the required changes for the new IngressClass feature.

Superseeds: #246 / #256


SantoDE:
Ping @DirtyCajunRice @myoung34 - I merged the PR's together to this one. PTAL if you have time :)

traefik/templates/rbac/clusterrole.yaml
```diff
@@ -31,6 +33,7 @@ rules:
       - extensions
     resources:
       - ingresses/status
+      - networking.k8s.io
```
DirtyCajunRice, line 36: This is an ApiGroup not a resource and needs to be removed

traefik/templates/ingressclass.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels:
+    app.kubernetes.io/name: {{ template "traefik.name" . }}
+    helm.sh/chart: {{ template "traefik.chart" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  name: {{ template "traefik.fullname" . }}
+spec:
+  controller: traefik.io/ingress-controller
+    {{- end }}
+    {{- end }}
```
DirtyCajunRice, line 17: newline at the end of the file

SantoDE:
Maybe @notsureifkevin as well :-)

DirtyCajunRice (changes requested):
Other than these requested changes, if you want this to be == in feature to the 2.3 release you should add a top level 
```yaml
pilot:
  token: “”
```
and an if statement enabling it as an argument. Other than that it looks good!

traefik/templates/ingressclass.yaml
SantoDE: Done

traefik/templates/rbac/clusterrole.yaml
SantoDE: Done

SantoDE:
Hmm, good catch @DirtyCajunRice. I think we should do it :)

SantoDE:
Done @DirtyCajunRice

traefik/templates/rbac/clusterrole.yaml
```diff
@@ -21,14 +21,17 @@ rules:
       - watch
   - apiGroups:
       - extensions
+      - networking.k8s.io
     resources:
       - ingresses
+      - ingressclasses
```
notsureifkevin, line 27: would this break if defined on a cluster that is not `v1.18`?

SantoDE: I think it would just not work :)

ldez (approved):
LGTM

tobylo:
What and where is the name of the controller specified? How would a setup look like when running multiple instances of traefik with the IngressClass resource?

PR #273, opened 2020-09-30
Create PodSecurityPolicy and RBAC when needed.

### LINKED ISSUE #272: helm Chart does not support PodSecurityPolicies
When installing traefik with this helm chart in a cluster with enabled PodSecurityPolicies, the PSP, Role and Binding have to be created additionally.

It would be nice, when the helm chart would also create the needed objects.

czunker:
Fixes #272

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.4.3
+version: 9.5.0
 appVersion: 2.3.1
 keywords:
   - traefik
--- a/traefik/templates/rbac/clusterrole.yaml
+++ b/traefik/templates/rbac/clusterrole.yaml
@@ -50,4 +50,14 @@ rules:
       - get
       - list
       - watch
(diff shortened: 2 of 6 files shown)


SantoDE:
Thanks for your PR @czunker.

Can you explain a bit more on the benefit, please? I've never used PSP in production yet (as its new with 1.19) so I'm interested in some insights.

Apart from that: the PR would need a version bump to `9.5.0`, please.

Thanks!

czunker:
Hi @SantoDE ,

I've bumped the chart version.

I don't know when PSPs where introduced, but the docs range back to at least 1.15:
https://kubernetes.io/docs/concepts/policy/pod-security-policy/

Pod Security Policies allow to limit the possibilities of PODs, e.g.:
- which volume types to mount
- whether a container might run privileged or not
- whether a container might run as root or not
- whether a container might access host ports or not
- ...

As the policies are controlled by an admission controller, the caveat is, PODs will not start, when the Admission Controller is enabled and no matching PSP is found.

PSPs in general enhance security and when integrated into helm charts, free users from having to write their own.

SantoDE (approved):
LGTM. Thanks :-)

PR #281, opened 2020-10-13
fix semver compare

DirtyCajunRice:
* tag was forced causing chart appversion to be unused and redundant
* version didnt always return a sane semver as compare expected whereas gitversion does

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.5.0
+version: 9.5.1
 appVersion: 2.3.1
 keywords:
   - traefik
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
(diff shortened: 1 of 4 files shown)


SantoDE:
Hey @DirtyCajunRice,

can you describe what's currently failing please? Just a PR without a description is a little hard ;-)

DirtyCajunRice:
> Hey @DirtyCajunRice,
> 
> 
> 
> can you describe what's currently failing please? Just a PR without a description is a little hard ;-)

@SantoDE updated description

SantoDE:
Thanks @DirtyCajunRice. Sadly, I still can't follow really :(

Why it is "better" or more sane in the GitVersion rather than in the Kube Version? And what is more convenient with the default App version? I'm sorry, but I need more explanation. I feel lost in why this is "better"

DirtyCajunRice:
@SantoDE  I actually did not realize that kube had slipped GitVersion into deprecated state. I have changed this to remove the semvercompare altogether and instead checked for the proper apiversion. This sanely decouples a hard number being placed in the yaml and instead better ensures that the apiversion exists. Doing this will remove the possible edge case where the version is correct but the apiversion is missing.

DirtyCajunRice:
@SantoDE I have resolved merge conflicts post TLS merge :)

SantoDE (approved):
LGTM 👼

PR #284, opened 2020-10-15
Replace extensions with policy because of deprecation

czunker:
Deprecation Info:
https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

The current docs also show the policy apiGroup:
https://kubernetes.io/docs/concepts/policy/pod-security-policy/#via-rbac


SantoDE (approved):
LGTM. Thanks 👼

PR #289, opened 2020-10-27
Add list of watched provider namespaces

samling:
If `rbac.namespaced` is `true`, the kubernetesIngress and kubernetesCRD providers need to be provided with a list of namespaces to watch, which right now defaults to the namespace the chart is installed in. Additional namespaces can be provided in `additionalArguments`, but this change allows them to be specified in the Values file the same way they would be in config.yaml.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -4,6 +4,7 @@ description: A Traefik based Kubernetes ingress controller
 type: application
 <<<<<<< HEAD
 <<<<<<< HEAD
+<<<<<<< HEAD
 version: 9.9.0
 =======
 version: 9.8.3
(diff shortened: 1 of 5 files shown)


SantoDE:
Hey @samling,

thanks for the contribution. I like it :)

One nitpick: can we have a unit test for that?

traefik/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.8.3
+version: 9.8.4
```
SantoDE, line 5: This needs to be 9.9.0, as it adds new values

SantoDE (changes requested):
Just one quick change :)

SantoDE (approved):
LGTM 👼

Issue #290, opened 2020-10-28
Add ImagePullSecrets


ctschubel:
I want to use this chart on an environment where I don't have access to public repositories, so it would be nice if I could pass custom `ImagePullSecrets` with a values.yaml.

PR #294, opened 2020-10-29
Add imagePullSecrets

ctschubel:
Add the possibility to pass custom imagePullSecrets with values.yaml
Issue: #290

### DIFF
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -51,7 +51,7 @@ spec:
     spec:
       {{- if .Values.deployment.imagePullSecrets }}
       imagePullSecrets:
-{{ toYaml .Values.deployment.imagePullSecrets | indent 6 }}
+        {{ toYaml .Values.deployment.imagePullSecrets }}
       {{- end }}
       serviceAccountName: {{ include "traefik.serviceAccountName" . }}
       terminationGracePeriodSeconds: 60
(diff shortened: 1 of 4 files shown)


traefik/templates/deployment.yaml
```diff
@@ -49,6 +49,10 @@ spec:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
         app.kubernetes.io/instance: {{ .Release.Name }}
     spec:
+      {{- if .Values.deployment.imagePullSecrets }}
+      imagePullSecrets:
+{{ toYaml .Values.deployment.imagePullSecrets | indent 6 }}
```
SantoDE, line 54: That indentation looks off?

SantoDE (changes requested):
Only 1 minor nitpick :)

Imroath:
In fact since 02.11 docker hub will introduce pull rate limiting, and everybody (especially on AKS/GKE clusters etc) will have to login to be able to pull.
This is a must have ASAP :)

traefik/templates/deployment.yaml
ctschubel: @SantoDE Thanks for the review!

The indent function of helm starts at the first character of the value so it has to be either:
```
    spec:
      {{- if .Values.deployment.imagePullSecrets }}
      imagePullSecrets:
{{ toYaml .Values.deployment.imagePullSecrets | indent 6 }}
      {{- end }}
```
or without indent function:
```
    spec:
      {{- if .Values.deployment.imagePullSecrets }}
      imagePullSecrets:
      {{ toYaml .Values.deployment.imagePullSecrets }}
      {{- end }}
```
To get this result:
```
    spec:
      imagePullSecrets:
      - name: myRegistryKeySecretName
```

SantoDE (approved):
LGTM 👍

SantoDE:
Thanks @Imroath. I know, and its going to be merged soon.

PR #296, opened 2020-10-30
Add ServersTransport crd

Lucaber:
Traefik introduced a new CRD `ServersTransport`.
Traefik does not start without this new crd deployed in the cluster and throws rbac permission errors.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.12.3
+version: 9.13.0
 appVersion: 2.3.6
 keywords:
   - traefik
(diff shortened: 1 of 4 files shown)


SantoDE:
Hello @Lucaber,

thanks for your PR.

The `ServersTransport` CRD is currently part of the master branch and not released with a Traefik Version. Id rather like to stale that up until we fully release the next minor versions.

Sounds good to you?

jorge07:
I would prefer put this out so I can test master in production for some time, there're features we want to use even before the release and we can act as beta testers if needed

SantoDE:
@jorge07 Right.

However, I don't think we should install components with the Chart that are not yet usable by the released version. Maybe, we could hide it by a special feature toggle?

jorge07:
Experimental flags sounds like an option for you?

Lucaber:
The crd is now behind an experimental feature flag and disabled by default.
It also moved into the `templates` directory as files in the crds directory are not rendered by helm.

jorge07:
I really like this implementation, you can promote that feature just moving it outside `experimental` when considered stable 👍

SantoDE:
Hello @Lucaber,

as we are on the way to release 2.4, can you move it outside of the experimental block again please?

PR #300, opened 2020-11-02
Specify IngressClass resource when checking for cluster capability

### LINKED ISSUE #287: IngressClass not working on AWS EKS 1.18 Cluster - no matches for kind "IngressClass" in version "networking.k8s.io/v1"
The way this Helm chart tries to evaluate the API version for `IngressClass` doesn't work for AWS EKS 1.18 cluster.

``` 
  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
apiVersion: networking.k8s.io/v1
  {{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
apiVersion: networking.k8s.io/v1beta1
``` 

You will get...
``` 
error: unable to recognize "ingressclass.yaml": no matches for kind "IngressClass" in version "networking.k8s.io/v1"
``` 

...because `IngressClass` runs with `networking.k8s.io/v1beta1`.

```
$ k explain IngressClass                              
KIND:     IngressClass
VERSION:  networking.k8s.io/v1beta1
```

Because `NetworkPolicy` runs with `networking.k8s.io/v1` as an API version `.Capabilities.APIVersions.Has "networking.k8s.io/v1"` will IMHO evaluate to `true`.

```
$ for kind in `kubectl api-resources | tail +2 | awk '{ print $1 }'`; do kubectl explain $kind; done | grep -e "KIND:" -e "VERSION:"
...
KIND:     IngressClass
VERSION:  networking.k8s.io/v1beta1
KIND:     Ingress
VERSION:  extensions/v1beta1
KIND:     NetworkPolicy
VERSION:  networking.k8s.io/v1
```

tobylo:
When installing the latest chart version on a 1.18.9 cluster, the capability check passes on having networking.k8s.io/v1, although in 1.18, the IngressClass resource was still in networking.k8s.io/v1beta1.

Now checking specifically for the IngressClass capability in the two api versions.

Resolves #287

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.10.0
+version: 9.10.1
 appVersion: 2.3.1
 keywords:
   - traefik
--- a/traefik/templates/ingressclass.yaml
+++ b/traefik/templates/ingressclass.yaml
@@ -1,7 +1,7 @@
 {{- if and .Values.ingressClass.enabled (semverCompare ">=2.3.0" (default .Chart.AppVersion .Values.image.tag)) -}}
-  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/IngressClass" }}
 apiVersion: networking.k8s.io/v1
-  {{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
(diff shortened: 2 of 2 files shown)


SantoDE (changes requested):
Good catch. Thanks!

Lint is failing, once resolved we can merge.

tobylo:
Does it still? think it was just because another PR was merged so had to bump the version one more time.

ldez (approved):

tobylo:
> 
> 
> Good catch. Thanks!
> 
> Lint is failing, once resolved we can merge.

I'm unable to resolve your change request and it seems blocking the automatic merge.

SantoDE (approved):

PR #304, opened 2020-11-11
add rbac aggregation cluster role

joschi36:
This PR adds two new ClusterRoles view and admin which aggregate to ClusterRole admin and view. This enables users to work with the CRDs provided.


joschi36:
@SantoDE can you have a look?

SantoDE:
Hello @joschi36,

I just took a look, but Im honest with you, I dont see the benefit as we have the other Cluster Role. Can you elaborate a bit more?

joschi36:
@SantoDE With an aggregated reader and edit/admin ClusterRole we have the possibility to authorize users to use Traefik without the need to create seperate ClusteRoleBinding.

The idea is when you grant the admin ClusterRole with a RoleBinding to a namespace, you also get access to work with Traefik within that namespace. You can aggregate to the existing view, edit and admin ClusterRole by setting labels to the ClusterRole, like I've done [here for the edit/admin](https://github.com/traefik/traefik-helm-chart/pull/304/files#diff-e5ffadfa7f1d31e8ad1d9be7fd46d661313e0f7c8a6ebc7af7ec7a549a36560dR11-R12) and [here for the view](https://github.com/traefik/traefik-helm-chart/pull/304/files#diff-e5ffadfa7f1d31e8ad1d9be7fd46d661313e0f7c8a6ebc7af7ec7a549a36560dR43) ClusterRole.

You can find more about ClusterRole aggregation here: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles

Thanks for looking at it :smiley:

SantoDE:
Hello @joschi36,

thanks for keeping the PR up2date.

One quick question because I cant find it: How is the aggregation flow? Am I correct to understand, that the one that receives the `rbac.authorization.k8s.io/aggregate-to-` annotations will be aggregated with the target, right?

Like what would the roles have effectively? Im super sorry, but its the first time I read up on that feature :)

joschi36:
Hi @SantoDE 

No problem thanks for your time.
Yes, as far as I understand you, that's correct.

Just to precise with an example of the admin ClusterRole.

This is the default `admin` ClusterRole in Kubernetes
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    rbac.authorization.kubernetes.io/autoupdate: "true"
  labels:
    kubernetes.io/bootstrapping: rbac-defaults
  name: admin
aggregationRule:
  clusterRoleSelectors:
  - matchLabels:
      rbac.authorization.k8s.io/aggregate-to-admin: "true"
rules:
# Aggregated rules from ClusteRoles with rbac.authorization.k8s.io/aggregate-to-admin: "true"
```

When you now create the traefik-aggregated-admin ClusterRole, the ClusterRole rules get written automatically into the admin ClusterRole by the kube-controller-manager (clusterrole-aggregation controller)

Does that help?

SantoDE:
Yes, that does help and does make me sure. Thanks a lot!

SantoDE:
Hello @joschi36,

sorry for the late replay. I was off for some time ;-)

One question left for me: Does that require some specific kube version we would need to be aware of or so? Do we need to hide the activation of the aggregation behind a specific flag?

joschi36:
Hey @SantoDE 
Aggregation was added a long time ago. I even think it was added RBAC, I quickly searched but couldn't find the exact version.


I think it is a good default, as everyone would want use traefik with the normal permissions provided by K8s.
Maybe we could hide it behind a flag and enable it by default.

SantoDE:
Im fine with that @joschi36.

Can you do the change? :)

SantoDE:
Hello @joschi36,

any update here?

joschi36:
Hey @SantoDE I've missed you message.
I will update the branch some when in the next days. The problem is I'm not sure how to do proper testing in Helm, but I will look into it. 
Happy New Year :smiley: :tada:

joschi36:
@SantoDE @ldez I finally found time to update the chart. I also tried to implement testing. Can you look at it?
I would be happy to see this merged :smile: 
FYI: I found that cert-manager is doing it similar in their Helm Chart: https://github.com/jetstack/cert-manager/blob/02c165a734a42efeac5e94d836de9eeb464e3940/deploy/charts/cert-manager/templates/rbac.yaml#L411-L455
However, without an option to control it.

joschi36:
I have updated the chart version to 9.15.0

joschi36:
@SantoDE Friendly Bump :)

joschi36:
Hi @SantoDE. I wanted if you have time to merge this.

PR #310, opened 2020-11-19
Fix dashboard ingress route

yix:
This PR fixes default dashboard ingressroute. In current configuration Traefik is trying to talk to dashboard back-end via TLS which doesn't work as it is listening on plain HTTP.

### DIFF
--- a/traefik/templates/dashboard-hook-ingressroute.yaml
+++ b/traefik/templates/dashboard-hook-ingressroute.yaml
@@ -17,6 +17,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
+  tls: {}
   entryPoints:
     - traefik
   routes:


SantoDE:
Hello @yix,

im sorry, but I can`t really follow the reason of that PR. 

The `IngressRoute` TLS Definition is used for incoming Traffic to Traefik and not for the backend connection part. Can you try to explain more, because I don`t think that is working the way you described in your PR?

yix:
I might not getting the purpose of this ingressRoute right. I though it's used to access Træfik dashboard from the outside. I was getting 404 until this empty tls section was added.

P.S. 404 response is a good reason to review and improve error reporting. 404 is returned in almost any situation when Træfic having trouble figuring out or reaching the backend. It should return something other than 404 if it knows about the backend but cannot talk to it for some reason. Current behaviour makes troubleshooting a pain.

PR #316, opened 2020-11-25
Adding support for metric collection through Prometheus Operator

langecode:
Adding support for enabling the Prometheus metrics in Traefik and also deploying ServiceMonitor resource for Prometheus Operator to pick up.


SantoDE:
Hello @langecode,

thanks for your contribution.

As you are probably aware, we had this PR a couple of times already (#90 #112 #116 #12) and thus, this is rather controversial topic ;-)

Please be aware, I will discuss this at great length with the others maintainers, however, that might take some time.

langecode:
Ahh, sorry, I actually did not search through the old PRs just thought this would be a nice feature first of all because the defacto metrics collector of Kubernetes workloads are currently Prometheus and second the Prometheus Operator approach makes it possible for deployments to signal the metrics they expose by deploying a ServiceMonitor manifest. I see more and more public charts supporting the rendering of an optional ServiceMonitor manifest as this is really a neat way to handle this.

But no rush we have cloned your repository and created a release of the chart internally with the metrics support included - so this was just to contribute back as we already did the change.

dmerrick:
This is awesome and looks super simple to me!

kitos9112:
A big thumb up for pushing this PR!

anthr76:
It would be great to have this! Maintainers you have my vote :)

faust64:
Changing the pod template should not be necessary.

Exposing Prometheus metrics can be done with:

```
additionalArguments:
- --metrics.prometheus=true
- --metrics.prometheus.entrypoint=metrics
- --metrics.prometheus.buckets=0.1,0.3,1.2,5.0
[...]

ports:
  metrics:
    port: 9113
    expose: false
    exposedPort: 9113
    protocol: TCP
[...]
```

Unless using a LoadBalancer Service exposing your ingresses (cloud integration), we could even add:

```
service:
  enabled: true
  labels:
    yourprometheuslabelselector: value
  type: ClusterIP
```

Using cloud integration, there could be an argument to add a separate Service, that would not expose metrics outside of your SDN.

Using the Prometheus Operator, we might want to automate creating such ServiceMonitor. Though that operator is still in beta, after quite a while, some bugs, ... Bearing in mind it used to be a CoreOS project, ... I would usually recommend deploying Prometheus from scratch, or using Helm.
ServiceMonitors implies namespace-scoped monitoring, as opposed to Prometheus native Kubernetes service discovery. The latter makes more sense, IMHO. Besides, in the few cases where namespace scoped monitoring was actually requested, I would still end up using Prometheus Kubernetes service discovery, while RBAC would filter in which namespaces to discover Services from. ServiceMonitors are basically useless, introducing more complications with the sole purpose of adding some granularity none of my customers had a need for so far.
But sure, that's up for debate.

Also, changing CRD api version should be done carefully, i think the `version` goes away and should be replaced by a `versions` array. Which I've been testing over there: https://github.com/Worteks/traefik-helm-chart/tree/update-crd . Also could be a little premature, v1beta1 won't be deprecated until 1.22, while v1 won't be available prior 1.16. And sure, anything below 1.18 should be updated right away. But my bet is there still are some out there.
In doubt, I can confirm such yaml won't apply (1.18 here), resulting in the following message:

```
 [ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): missing required field "versions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec]
```

langecode:
Yes, indeed the CRDs in this PR are not valid - we recently also discovered this. I have generated correct `v1` CRDs from the Traefik go code an added those to our fork of the Helm chart: https://github.com/neticdk/traefik-helm-chart/tree/master/traefik/crds

I will close this PR now as there does not seem to be an interest in natively support the Prometheus Operator in the Helm chart. If anyone finds this PR we currently maintain a fork of the Helm chart including the Prometheus Operator support available as Helm repository here: https://neticdk.github.io/traefik-helm-chart

PR #318, opened 2020-12-06
Adding urls that the dashboard will be accessible via to the readme

Bwvolleyball:
For folks who are attempting to follow this helm chart, it's useful to have which urls the traefik dashboard will be accessible via under each scenario, especially since the trailing `/` is necessary, and it may be unintuitive that the localhost ip address must be used when port-forwarding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ For instance, the dashboard access could be achieved through a port-forward :
 kubectl port-forward $(kubectl get pods --selector "app.kubernetes.io/name=traefik" --output=name) 9000:9000
 ```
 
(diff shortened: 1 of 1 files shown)


SantoDE (changes requested):
Thanks @Bwvolleyball.

That PR needs a minor version bump to pass. Once done, its ready to get merged! :)

Bwvolleyball:
Hello @SantoDE - can you elaborate? These are just documentation changes, so I don't understand what version needs to be  bumped. Sorry if this is a silly question..

SantoDE (approved):
LGTM 👼

Issue #339, opened 2021-01-20
Support for overriding the liveness/readiness probes port


ttauveron:
Hello,

I spent a bit of time recently to deploy traefik behind a Google L7 HTTP/S Load balancer.
To create a HTTPS Load balancer in GKE, a service must first have a NodePort type and an Ingress of type "gce" must point to this service. Google will then automatically create healthchecks using by default the configuration specified in the pod for liveness/readiness probes. However, by default, this helm chart uses the "traefik" entrypoint port (that shouldn't be exposed) and the Google Load balancer cannot succeed its healthchecks because it can't reach this entrypoint. 

I managed to solve this by modifying the ping configuration in traefik : 

```yaml
additionalArguments:
  - "--ping"
  - "--ping.entrypoint=web"
```

and change the liveness/readiness ports to the web entrypoint one (by default, 8000 instead of 9000).

Would that make sense in this chart to support overriding those liveness/readiness ports in the values?

Something like that?

```diff
diff --git a/traefik/templates/_podtemplate.tpl b/traefik/templates/_podtemplate.tpl
index f401d26..3d5ddc5 100644
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -38,7 +38,7 @@
         readinessProbe:
           httpGet:
             path: /ping
-            port: {{ .Values.ports.traefik.port }}
+            port: {{ default .Values.ports.traefik.port .Values.ports.traefik.healthchecksPort }}
           failureThreshold: 1
           initialDelaySeconds: 10
           periodSeconds: 10
@@ -47,7 +47,7 @@
         livenessProbe:
           httpGet:
             path: /ping
-            port: {{ .Values.ports.traefik.port }}
+            port: {{ default .Values.ports.traefik.port .Values.ports.traefik.healthchecksPort }}
           failureThreshold: 3
           initialDelaySeconds: 10
           periodSeconds: 10
diff --git a/traefik/values.yaml b/traefik/values.yaml
index 50cab94..8dfe19c 100644
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -207,6 +207,11 @@ ports:
   # liveness probes, but you can adjust its config to your liking
   traefik:
     port: 9000
+
+    # Override the liveness/readiness port. This is useful to integrate traefik
+    # with an external Load Balancer that performs healthchecks.
+    # healthchecksPort: 9000
+
     # Use hostPort if set.
[...]

PR #353, opened 2021-02-04
feat(healthchecksPort): Support for overriding the liveness/readiness probes port

faust64:
Allows to change liveness & readiness probes port, facilitating integration with GKE.

Fixes #339

PR #354, opened 2021-02-04
Customizable liveness/readiness probes

Tillerino:
This addresses #339 but instead of customizing which of the named ports is used for the probes, the entire probes are exposed. While this is not as elegant, it looks to me like this is the standard way of doing this - albeit my experience here is surely limited.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.14.2
+version: 9.15.0
 appVersion: 2.4.2
 keywords:
   - traefik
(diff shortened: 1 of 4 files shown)


Tillerino:
Looks like the other approach was implemented simultaneously over in #353 :D

Either way works for me. Right now I'm stuck with `kubectl apply` 😅

SantoDE:
Hey 👋 

First of, Im sorry for the late reaction. Last days were pretty busy.

I would look to go with #353 first as its simpler and might already solve the need. If we figure out, its not enough we can still iterate and potentially work on that here. 

Ill close that PR here for that reason. Please let me know if you think, that is not suitable enough.

Tillerino:
Yeah, works for me!

PR #356, opened 2021-02-08
allow `spec.strategy` to be `Recreate` in deployment

### LINKED ISSUE #355: allow `spec.strategy` to be `Recreate` in deployment
At the moment, `RollingUpdate` is hard-coded as the update strategy for Traefik deployment. Please create a value to change the strategy to `Recreate` since this is required for RWO PVCs.

faust64:
Allows to configure a deployment strategy from chart values.
Using LetsEncrypt integration with ReadWriteOnce PersistentVolumeClaims requires the deployment strategy to be Recreate, which is not doable with the current hard-coded RollingUpdate.

fixes #355

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.14.2
+version: 9.14.3
 appVersion: 2.4.2
 keywords:
   - traefik
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.14.3
+version: 9.14.4
 appVersion: 2.4.5
 keywords:
   - traefik
--- a/traefik/templates/deployment.yaml
+++ b/traefik/templates/deployment.yaml
@@ -32,10 +32,12 @@ spec:
       app.kubernetes.io/name: {{ template "traefik.name" . }}
       app.kubernetes.io/instance: {{ .Release.Name }}
   strategy:
-    type: RollingUpdate
+    type: {{ default "RollingUpdate" .Values.deployment.strategy }}
+    {{- if eq (default "RollingUpdate" .Values.deployment.strategy) "RollingUpdate" }}
     rollingUpdate:
       {{- with .Values.rollingUpdate }}
         {{- toYaml . | nindent 6 }}
       {{- end }}
+    {{- end }}
   template: {{ template "traefik.podTemplate" . }}
 {{- end -}}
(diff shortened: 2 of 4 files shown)


SantoDE:
Hey @faust64,

thanks for the contribution.

Sorry to ask, but I can't quite follow. 

> Using LetsEncrypt integration with ReadWriteOnce PersistentVolumeClaims requires the deployment strategy to be Recreate

Why is that? WDYM?

faust64:
I'm not sure what was the reason for #355, though I did have a customer generating certificates with LetsEncrypt yet not using a persistent volume storing them.

Reconfiguring their deployment, I eventually ended up with LetsEncrypt refusing to issue new certificates, as I did hit a limit. And outside of Traefik/Kubernetes context, I've also been hitting with that limit: I'm scheduling certificates renewal such as I don't request more too many witthin a 7 days period. (not sure to remember what's that limit exactly, some mentions about this: https://community.letsencrypt.org/t/too-many-certificates-already-issued-for-exact-set-of-domains/110521 , https://community.letsencrypt.org/t/too-many-certificates-already-issued/110571 , https://community.letsencrypt.org/t/too-many-certificates-already-issued-for-exact-set-of-domains/103003 ).

Anyway, one way to work around this would then be to store certificates in a persistent volume. Although ReadWriteMany accessMode isn't possible with any StorageClass, most would just allow for exclusive accesses (ReadWriteOnce).

Now, when you have a Deployment, with a ReadWriteOnce volume attached, you should know the RollingUpdate deployment strategy won't work.
In Kubernetes, the RollingUpdate imples that when changing your Deployment definition, a new Pod would start first, while the previous copy would not be stopped until the new one's liveness/readiness probes return OK.
Using ReadWriteOnce volume, your new Pod can't be started, as Kubernetes scheduler would fail to attach the corresponding PVC. Pod would be stuck in ContainerCreating.
You need a Recreate strategy, for the scheduler to first stop the previous Pod, and only then create the new one.

faust64:
So, after a few tests with @jakubhajek, it turns out the default values would install a Deployment whose rollingUpdate params include a `maxUnavailable: 1` ( https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml#L91-L92 ).

As such, when applying a new revision for a one-replica deployment, the scheduler would actually create a new Pod and terminate the previous one pretty much simultaneously. Thus, when using a RWO PVC, the multi-attach errors might show for a short while, though would not prevent the new revision from eventually starting up. Whereas my previous remarks relates to a deployment that was not created by Helm, and did use Kubernetes default maxUnavailable value of 25% - in which case, we would reproduce the issue.

Then, we probably don't need to allow for changing the deployment strategy.
If you have more than one replicas, you can't use a RWO PVC anyway.
If you use a RWO PVC, as long as you don't change the default `rollingUpdate` params, you'ld be OK.

Closing. Unless there's a better argument I would have missed, rollingUpdate should be fine.
If anything, we could raise a warning, should we see a combination of `persistence.enabled: true` with `rollingUpdate.maxUnavailable` being less than 1 or 100% and `persistence.accessMode` != ReadWriteMany.

PR #360, opened 2021-02-16
fix(tags): handle non-semvers tag

### LINKED ISSUE #321: ingressclass template causes chart install to fail when using non-semver values in `image.tag`
I encountered this issue when trying to force the helm chart to use the `latest` image tag when testing a new feature. This bug cannot be worked around because the logic is always going to evaluate the SemVer in the following code (even if `.Values.ingressClass.enabled` evaluates to `false`):

https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/ingressclass.yaml#L1

Do we need this logic check here? Is it not enough to assume the operator is enabling `ingressClass` when they understand their cluster has the required APIs and support? See related issue - https://github.com/traefik/traefik-helm-chart/issues/309

values.yaml:

```yaml
        image:
          name: traefik/traefik
          tag: latest
          pullPolicy: Never

```

Result:

```
can't install chart `traefik`: template: traefik/templates/ingressclass.yaml:1:41: executing \"traefik/templates/ingressclass.yaml\" at <semverCompare \">=2.3.0\" (default .Chart.AppVersion .Values.image.tag)>: error calling semverCompare: Invalid Semantic Version"
```

faust64:
Ensures IngressClass installation check would properly handle deployment for traefik `latest` tag.

Fixes #321

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.15.2
+version: 9.15.3
 appVersion: 2.4.6
 keywords:
   - traefik
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,8 +2,13 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
+<<<<<<< HEAD
 version: 9.15.3
 appVersion: 2.4.6
+=======
+version: 9.14.4
+appVersion: 2.4.5
+>>>>>>> fix(version)
 keywords:
   - traefik
   - ingress
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,13 +2,8 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-<<<<<<< HEAD
 version: 9.15.3
 appVersion: 2.4.6
-=======
-version: 9.14.4
-appVersion: 2.4.5
->>>>>>> fix(version)
 keywords:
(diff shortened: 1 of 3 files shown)


SantoDE:
Hey @faust64,

can you bump the version again please? I would do so, but looks like you did not allow edits from maintainers :)

traefik/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.14.3
+version: 9.14.4
```
SantoDE, line 5: Needs to be 9.15.5

SantoDE (changes requested):
See above

SantoDE (commented):
LGTM. Thanks!

SantoDE (approved):
LGTM

SantoDE:
Hey @faust64,

can you please provide maintainers with permissions to do edits on that PR? That would help a lot.

Thanks!

faust64:
I'm not sure I can. AFAIU, there should be a button on under the Unsubscribe one, in that pull request ( https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork )

I can't find it, not sure why:
unsubscribe-btn

let me know if there's anything I can do.

SantoDE:
Somehow, I cant push to your branch. No matter what I try.

Can you open the PR (check the box) to allow edits from maintainer? :)

faust64:
Sorry, as stated earlier, I don't see that checkbox. This seems to be due to my forking to my organization instead of my own account. I can re-open my PRs, forking again, if there's no way around this.

faust64:
see https://github.com/traefik/traefik-helm-chart/pull/384

PR #361, opened 2021-02-16
feat(labelSelector): option matching Ingresses based on labelSelectors

### LINKED ISSUE #264: labelSelector field for kuberneteingress provider
labelSelector allows to decide wheter an ingress should be managed by traefik or not (ingress controller selection). Right now, the only to achieve this is through the additionalArguments, but should rather be available directly under the provider section, such as:

```
providers:
  kubernetesIngress:
    enabled: true
    labelSelector: A and not B
```

faust64:
Fixes #264

traefik can pick Ingress objects based on a labelSelector.
This can be useful in larger clusters, when ingress are dedicated to an environment (prod, stage, ...), when SDN is broken into security zones with dedicated ingresses, ...

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.14.2
+version: 9.14.3
 appVersion: 2.4.2
 keywords:
   - traefik
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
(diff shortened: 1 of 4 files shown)


SantoDE:
Hey @faust64,

can we have some quick unit tests for that please? :)

traefik/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.15.2
+version: 9.15.5
```
SantoDE, line 5: 9.16.0 plx :)

SantoDE (changes requested):
One last step

traefik/Chart.yaml
faust64: :+1:

faust64:
See #390

Issue #335, opened 2021-01-14
Helm chart : replicas and DNS acme resolver


Leopol35:
Hello,

I am using Traefik 2.3.X in an "on premise" kubernetes cluster. I use GandiV5 DNS resolver to get certificates for traefik.

In the helm chart, you forbid to use an "acme" configuration when there is more than 1 pod replica.

```
{{- if and .Values.deployment.enabled (eq .Values.deployment.kind "Deployment") -}}
  {{- if gt (int .Values.deployment.replicas) 1 -}}
    {{- with .Values.additionalArguments -}}
      {{- range . -}}
        {{- if contains ".acme." . -}}
          {{- fail (printf "You can not enabled acme if you set more than one traefik replica") -}}
        {{- end -}}
      {{- end -}}
    {{- end -}}
  {{- end -}}
```

I totally understand this for HTTP-01, TLS-SNI and TLS-ALPN challenge but for DNS-01, it should be ok to have more than one replica. In case of DNS challenge, Traefik will check the DNS entry to the provider, it(s not an incoming traffic to traefik.

Thanks for your help,

Kind regards,

rdxmb:
According to [the closed pull request](https://github.com/traefik/traefik-helm-chart/pull/366) this issue could be closed, right?

(In my case, I use traefik 2.2 as a deployment with 2 replicas - event with HTTP-ACME. The acme.json is on NFS. This is working for about 130 days in 3 datacenters. Seems I've had just luck?)

Leopol35:
Hello,

Yes, I close, thanks a lot for your attention. I was not aware of this limitation about let's encrypt certificate.

Kind regards,

PR #366, opened 2021-02-20
Disallow only ACME with TLS or HTTP challenge

peat-psuwit:
We know exactly which challenges require special care. So why not
restrict only those, and let people that use the other (read: DNS)
challenge enjoy the Helm chart?

Fixes: https://github.com/traefik/traefik-helm-chart/issues/335

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.14.3
+version: 9.15.0
 appVersion: 2.4.5
 keywords:
   - traefik
--- a/traefik/templates/daemonset.yaml
+++ b/traefik/templates/daemonset.yaml
@@ -1,8 +1,8 @@
 {{- if and .Values.deployment.enabled (eq .Values.deployment.kind "DaemonSet") -}}
   {{- with .Values.additionalArguments -}}
     {{- range . -}}
-      {{- if contains ".acme." . -}}
-        {{- fail (printf "ACME functionality is not supported when running Traefik as a DaemonSet") -}}
+      {{- if (or (contains ".acme.tlschallenge" .) (contains ".acme.httpchallenge" .) ) -}}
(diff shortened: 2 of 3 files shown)


SantoDE:
Hey @peat-psuwit,

first of: sorry for the late answer. Last days were pretty busy.

While I would like to agree, I think, this challenge even has its problems right? How would Traefik know that a recreate is already in progress? All replicas would try to do the DNS challenge at the same time then. Wouldn`t that cause some issues at the LE side?

peat-psuwit:
I don't think LE has a problem dealing with it, because [LE's challenge types page](https://letsencrypt.org/docs/challenge-types/#dns-01-challenge) specifically says that:

> You can have multiple TXT records in place for the same name. For instance, this might happen if you are validating a challenge for a wildcard and a non-wildcard certificate at the same time.

It even says that:

> It works well even if you have multiple web servers.

So, I believe Traefik (more specifically Lego) doesn't need to know about any other instance creating TXT records at the same time. As long as Lego cleans up the record afterward (which it does), it should be fine.

SantoDE:
But wouldn`t it re-issue the certs multiple times (the replica count) then? Im wondering about potential LE ratelimits here.

peat-psuwit:
Ah, right, the rate limit. Yeah, the rate limit could be reached if replicas create the same certificate multiple times. [According to LE](https://letsencrypt.org/docs/rate-limits/), the limit for duplicate certificates is 5 per week, and if the replica size is bigger, this limit could be reached.

Hmm...

SantoDE:
Yeah, thats what I was afraid of :( I guess the PR won't work pretty nicely then, or better, its rather risky Im afraid...

SantoDE:
Unless, you can't think of something else, I guess I have to decline that PR

SantoDE:
No problem. Thats what we are here for :)

Leopol35:
Hello,

Yes, you can close, thanks a lot for your attention. I was not aware of this limitation about let's encrypt certificate.

Kind regards,

PR #372, opened 2021-02-25
Configurable PVC name

svengreb:
Currently the name of the [PVC for persistence][ref-pvc-pers] is “hardcoded“ to `data`. This could conflict when users add additional volumes and/or mounts that use the same name, resulting in “duplicate volume“ errors that are not as easy to debug and understand in the first place.
I ran into such a problem after trying to add an additional volume also named `data` and invested some hours trying to find the error within my own Kustomize + Flux setup while the actual root cause was not that obvious.

To prevent such situations, this PR adds a new `persistence.name` Helm values field, keeping `data` as default name, to allow users to change it when required.
I've added a new test and extended an existing one to also ensure that the default name still works fine. 

[ref-pvc-pers]: https://github.com/traefik/traefik-helm-chart/blob/6b4c3642/traefik/values.yaml#L328

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.14.4
+version: 9.15.1
 appVersion: 2.4.5
 keywords:
   - traefik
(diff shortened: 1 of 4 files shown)


SantoDE (approved):
LGTM. Thanks!

PR #373, opened 2021-02-27
Adds support for namespace overrides when Traefik is extended as a sub-chart

davidlundy:
### What problem does this solve?

This pull request improves the experience for cluster and namespace admins who may be extending this Helm chart via use as a sub-chart, which is an encouraged approach in the Traefik philosophy. 

An example use case is that as a cluster admin, I want to create an umbrella chart which bootstraps a bare cluster with all the infrastructure necessary to support application deployments; e.g. ingress (Traefik), service mesh, admission control, monitoring, configs for all of it, and some of our own tooling. 

The desired behavior for this use case is that all manifests provided directly by the umbrella chart (all my tools and configs) are installed to the Helm release namespace, and each dependency of that chart is installed to a separate dedicated namespace. 

The current situation is that the Traefik Helm chart assumes that the Helm release namespace is where it should deploy manifests and provides no way to opt-out of that behavior. This limits the ability to compose Traefik with other dependencies as in the use case above. 

### Solution Approach

This pull request provides an optional feature called `namespaceOverride`. As a non-critical optional feature, it has been added to `values.yaml` disabled by comment. 

If a value is not provided for `namespaceOverride`, the default behavior of using the Helm release namespace is preserved. 

Use cases as described above must be responsible for ensuring the target namespace exists and overriding the value. 

### Open Question

The experimental Gateway manifest ignores the existing behavior of the other namespaced resources in this chart and instead has a hard coded `default` as the namespace. Is this desired behavior?

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.14.3
+version: 9.14.4
 appVersion: 2.4.5
 keywords:
   - traefik
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -33,6 +33,19 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 
 {{/*
+Construct the namespace for all namespaced resources
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+Preserve the default behavior of the Release namespace if no override is provided
+*/}}
+{{- define "traefik.namespace" -}}
+{{- if .Values.namespaceOverride -}}
+{{- .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- .Release.Namespace -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 The name of the service account to use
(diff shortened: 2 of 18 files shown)


nerzhul:
hello @traefiker , can you look for this MR, i'm also interested in, as it permits to embed traefik as a subchart for special helm dependencies usecases.
Thanks

SayakMukhopadhyay:
Hey there, I was in the same boat and it would be great to see this PR merged. I don't think there is anything breaking here so should be a trivial merge.

thechristschn:
I'm also interested in this MR, but with another use-case.

We template the helm-chart, do some post-processing and then apply it with kubectl.
If the namespace is not defined in the helm chart, the template won't contain the namespace unless you define it with `namespace: {{ .Release.Namespace }}`.

Currently the chart is just applied to whatever namespace is currently defined in the kubectl context. This is error prone and can be avoided easily with this MR.

charlie-haley:
Hi @davidlundy, thanks for the PR! It looks like the behaviour of the `Gateway` resource namespace has changed since this PR was created, it now has its own override in `kubernetesGateway.namespace, so it seems like it's no longer an issue now

Could you look at rebasing the PR? Other than that, LGTM

charlie-haley:
Closing due to the original fork being deleted, superseded by https://github.com/traefik/traefik-helm-chart/pull/718

PR #377, opened 2021-03-02
Adding custom labels to deployment

### LINKED ISSUE #374: Cannot add Pod Labels in Daemonset deployment
There looks to be a small bug with the nindent for the pod labels in the daemonset.yaml template. When attempting to add additional pod labels to the daemonset e.g.:
```yaml
deployment:
  kind: DaemonSet
  podLabels:
    foo: bar
```
You get the following error when attempting to deploy the chart:
`error converting YAML to JSON: yaml: line 10: mapping values are not allowed in this context`

Adding annotation works just fine so I assume the issue with pod labels is due to the nindent of 8 where it should be 4
[daemonset.yaml line 21](https://github.com/traefik/traefik-helm-chart/blob/bbad36d8b871e35f1c2290b1a033d2f01c9a94ab/traefik/templates/daemonset.yaml#L21)

**Traefik Chart version:** 9.14.2
**Helm version:** 3.3.4

jakubhajek:
The change should fix #374 and allow adding custom labels for Deployment or DaemonSet.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.15.2
+version: 9.15.3
 appVersion: 2.4.6
 keywords:
   - traefik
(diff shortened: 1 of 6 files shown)


SantoDE (approved):
LGTM

PR #384, opened 2021-03-05
Throw error with explicit latest tag

### LINKED ISSUE #321: ingressclass template causes chart install to fail when using non-semver values in `image.tag`

faust64:
following-up on #360 

latest tag handling.

Fixes #321


traefik/templates/ingressclass.yaml
```diff
@@ -1,11 +1,13 @@
-{{- if and .Values.ingressClass.enabled (semverCompare ">=2.3.0" (default .Chart.AppVersion .Values.image.tag)) -}}
-  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/IngressClass" }}
+{{- if .Values.ingressClass.enabled -}}
+  {{- $hasigclass := "traefik.hasingressclass" }}
+  {{- if $hasigclass }}
+    {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/IngressClass" }}
 apiVersion: networking.k8s.io/v1
-  {{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1/IngressClass" }}
+    {{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1/IngressClass" }}
```
jakubhajek, line 6: I believe those extra whitespaces are not necessary.

jakubhajek (approved):
I tested that change locally, no issues occurred.

SantoDE:
@faust64 what do you think of the comment in #321? Do you think that would be sufficient? :)

faust64:
Sure, one should never use latest tag. At least when you're not pulling from a registry that expunges older images, or some rare exceptions when you need some missing feature that's not in an official release, ... doesn't rely apply here.
Then again, good practices are rarely observed.
Another take on it may be to throw an error instead.

traefik/templates/ingressclass.yaml
faust64: they're not indeed.
Looking at other parts in code, I can see there's some 2 spaces indentation each time we enter a if (daemonset, deployment, ...). Arguably, it may improve readability. Though I don't really mind, we can remove it.

SantoDE:
Like to throw an error if you are trying to run it with latest? I would like that :)

faust64:
allright, that should do it

jakubhajek (approved):
It works as expected preventing a user to deploy the latest tag.

jakubhajek (approved):

ldez (approved):
LGTM

Issue #266, opened 2020-09-22
My PV was deleted after chart upgrade and setting "existingClaim"


StarpTech:
This happened after I upgraded my release to the latest version and set `existingClaim` to my existing claim `"traefik"`. An existing PV should not be deleted when I want to upgrade. We should remove that option and only allow it to pass the name of an existing claim.

Using https://github.com/databus23/helm-diff reports the changes between both revisions:
```
default, traefik, PersistentVolumeClaim (v1) has been removed:
- # Source: traefik/templates/pvc.yaml
- apiVersion: v1
- kind: PersistentVolumeClaim
- metadata:
-   name: traefik
-   annotations:
-   labels:
-     app.kubernetes.io/name: traefik
-     helm.sh/chart: traefik-8.7.1
-     app.kubernetes.io/managed-by: Helm
-     app.kubernetes.io/instance: traefik
- spec:
-   accessModes:
-     - "ReadWriteOnce"
-   resources:
-     requests:
-       storage: "128Mi"
+
```

![image](https://user-images.githubusercontent.com/1764424/93924333-e3213900-fd14-11ea-9193-42e90f2240b8.png)

PR #385, opened 2021-03-08
Add helm resource-policy annotation on PVC

faust64:
As reported in https://github.com/traefik/traefik-helm-chart/issues/266

We could argue it is expected of Helm to purge objects based on a change of values, we could ensure that a PersistentVolumeClaim would never be removed automatically.


SantoDE:
Im super sorry, but our tooling won't merge PR's from your organization account (even when we accept the invite).

Can you redo them on your personal account?

faust64:
allright, see #389

PR #389, opened 2021-03-10
Add helm resource-policy annotation on PVC

faust64:
Following up on #385 

We could argue it is expected of Helm to purge objects based on a change of values, we could ensure that a PersistentVolumeClaim would never be removed automatically.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.17.3
+version: 9.17.4
 appVersion: 2.4.7
 keywords:
   - traefik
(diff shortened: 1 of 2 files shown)


jakubhajek (approved):
The version should be bumped.

SantoDE:
What I am wondering about, should that be configurable? Or can you think of no scenario where it should be deleted?

jakubhajek (approved):
The change looks good to me. It prevents removing the PVC created by Helm chart,  no matter what ReclaimPolicy has been set in default storage class. 
Regarding the reported #266 bug; it has no reference to `existingClaim` value that can be set in order to use the volume created prior to Helm install.

faust64:
Right, #266, not #355 ... Tells about:

> This happened after I upgraded my release to the latest version and set existingClaim to my existing claim "traefik"

AFAIU, there was an existing deployment managed by Helm, with persistence enabled. PV was lost by adding an existingClaim to the values file, updating that deployment.
The helm update command then compared existing objects, created by the previous helm install, to those that should be applied, resulting in the existing PVC destruction, since no longer managed by Helm.

I don't think any documentation would suggest to do this. It's most likely a misunderstanding.
Could have happened with pretty much any chart, definitely not a bug.
Annotating the PVC could serve as a safety net against pebkac.

ldez (approved):
LGTM

PR #390, opened 2021-03-10
feat(labelSelector): option matching Ingresses based on labelSelectors

faust64:
Following up on #361 

traefik can pick Ingress objects based on a labelSelector.
This can be useful in larger clusters, when ingress are dedicated to an environment (prod, stage, ...), when SDN is broken into security zones with dedicated ingresses, ...

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.17.4
+version: 9.17.5
 appVersion: 2.4.7
 keywords:
   - traefik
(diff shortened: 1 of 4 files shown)


faust64:
See #391

traefik/Chart.yaml
```diff
@@ -2,7 +2,9 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.16.2
+version: 9.16.0
+appVersion: 2.4.6
```
SantoDE, line 6: that needs cleanup :) Needs a new minor version + the correct appVersion

SantoDE (dismissed):
One comment :)

jakubhajek (approved):
The change looks good to me. We could consider adding a line of comment explaining that once the feature `labelSelector` is enabled the labels should be added to the Ingress object in order to be processed.

ldez (approved):
LGTM

PR #398, opened 2021-03-24
Add an option to toggle the pilot dashboard

SantoDE:
Adds an option to toggle the pilot dashboard through the helm chart, which is a new option since 2.4.8

PR #400, opened 2021-03-29
Do not disable Traefik Pilot in the dashboard by default

kevinpollet:
This PR fixes the behaviour of the option introduced in #398.
By default, Traefik Pilot must be enabled in the dashboard as it's implemented in Traefik.


SantoDE (approved):
LGTM

PR #408, opened 2021-04-16
Add shareProcessNamespace option to podtemplate

emourant:
This change relates to https://github.com/traefik/traefik/issues/3489. Currently if you're using Kubernetes with a sidecar to ship logs there is no way to rotate logs and signal Traefik to [close/re-open](https://doc.traefik.io/traefik/observability/logs/#log-rotation) its log files.

The addition of the [shareProcessNamespace](https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/) pod spec flag with a logrotate sidecar including the `SYS_PTRACE` securityContext.capabilities  allows one to send the `USR1` signal to the traefik process and trigger it to close/re-open its logs

There are a number of ways the same result could be achieved, so I would be open to alternatives if one is suggested.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.11.1
+version: 10.12.0
 appVersion: 2.6.0
 keywords:
   - traefik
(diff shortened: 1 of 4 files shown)


SantoDE:
Hey @emourant,

thanks for that contribution. First time I encountered that flag, but I have to agree I think it makes sense :)

Can we get a quick unit test for that change? Also, its a new minor version rather then a patch version.

kevinpollet (approved):
Thanks 👍

PR #411, opened 2021-04-22
allow termination grace period to be configurable

kevtainer:
noticed that this was hardcoded, and can be problematic if one configures traefik for `entrypoints.foo.transport.lifeCycle.graceTimeOut > 60`


SantoDE:
Hey @kevtainer,

needs a version bump to `9.19.0` :)

PR #419, opened 2021-05-15
Add Acme certificate resolver configuration

valerauko:
Being able to configure these options via values is much more friendly than having to write it through additionalArgs.

- [x] added tests for the new values
- [x] bumped chart version

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.20.0
+version: 10.20.1
 appVersion: 2.7.0
 keywords:
   - traefik
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -433,26 +433,26 @@ persistence:
   annotations: {}
   # subPath: "" # only mount a subpath of the Volume into the pod
 
-certResolvers:
-  letsencrypt:
-    # for challenge options cf. https://doc.traefik.io/traefik/https/acme/
-    email: email@example.com
-    dnsChallenge:
-      # also add the provider's required configuration under env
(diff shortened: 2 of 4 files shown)


traefik/values.yaml
```diff
@@ -349,6 +349,27 @@ persistence:
   annotations: {}
   # subPath: "" # only mount a subpath of the Volume into the pod
 
+certResolvers: {}
+#   letsencrypt:
```
jakubhajek, line 353:
```suggestion 
I think that indention is not correct, there should 2 white spaces instead of 3.

File: traefik/values.yaml [hunk 2]
```diff
@@ -349,6 +349,27 @@ persistence:
...
+#     # for challenge options cf. https://doc.traefik.io/traefik/https/acme/
+#     email: email@example.com
+#     dnsChallenge:
+#       # also add the provider's required configuration under env
```
jakubhajek, line 357:
```suggestion
We have to emphasize that Kubernetes Secret containing DNS provider data is mandatory.  Otherwise, a user might be confused about where the sensitive (APK_KEY, login) data should be added.

File: traefik/Chart.yaml [hunk 3]
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.19.1
+version: 9.19.2
```
jakubhajek, line 5:
```suggestion

Would you please update the chart and Traefik version?

jakubhajek (commented):
Thanks for submitting the PR and adding a user-friendly approach to implement the certResolver feature. 

Please review my comments and rebase your branch. Then, I think we are happy to merge it.

shadycuz:
@jakubhajek This is a pretty important feature, why the maintainers just finish the work that was started?

ddtmachado (approved):
LGTM

PR #311, opened 2020-11-19
Revert split LoadBalancers to single LoadBalancer with mixed protocols

0xThiebaut:
Commit 6d60914 acted as a work-around for LoadBalancers' single protocol requirement (see [bergmannf's comment](https://github.com/traefik/traefik-helm-chart/pull/145#issuecomment-610791540) on #145). Kubernetes now supports LoadBalancers with mixed protocols (kubernetes/kubernetes#94028).

This commit reverts the original work-around to reintroduce a single TCP/UDP LoadBalancer for Traefik, cutting LoadBalancer costs in half.

_Note: The mixed-protocol LoadBalancers are planned to be included in [Kubernetes milestone v1.20](https://github.com/kubernetes/kubernetes/milestone/48)._

PR #420, opened 2021-05-15
Only create one Service based on version check

valerauko:
This is the continuation of #311. I cherry-picked @0xThiebaut's commits and added the tests as requested in #311.

Since the rancher-maintained helm-unittest does not support mocking Capabilities, I switched to an apparently [better-maintained other fork](https://github.com/quintush/helm-unittest).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ endif
 # Execute Unit Testing
 unit-test: helm-unittest
 	@echo "== Unit Testing Chart..."
-	@helm unittest --color --update-snapshot ./traefik
+	@helm unittest --helm3 --color --update-snapshot ./traefik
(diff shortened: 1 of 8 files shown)


valerauko:
Ping @SantoDE since you asked for tests in the original PR

valerauko:
On the other hand, using this still requires a feature gate enabled in the apiserver...

https://kubernetes.io/docs/concepts/services-networking/service/#load-balancers-with-mixed-protocol-types

Elegant996:
Saw that this PR actually has the testing. Could we update the version from 1.20 to 1.24 for Kubernetes? Mixed protocol is now beta and is enabled by default going forward.

mloiseleur:
Superseded by #656

Issue #156, opened 2020-04-20
Add support for prometheus ServiceMonitor


arlyon:
It would be nice to be able to be able to include a CRD for a prometheus service monitor with the chart.

SantoDE:
Hello @arlyon,

as previously stated, up until know we declined feature requests like that. However, I'd like to acknowledge that we started some discussions around that topic, as we see that you, our community has a demand for it.

So for now, no real updates on it but please be aware, we're talking over it.

r3kzi:
Since then you can use the `additionalServiceMonitors` feature in the Prometheus-Operator. That's what we are using right now.

```
prometheus:
  additionalServiceMonitors:
    - name: traefik
      selector:
        matchLabels:
      ...
```

k3daevin:
It would be nice because the old stable/traefik Helm Chart had this option.
Might I create a PR?

CuBiC3D:
@SantoDE are there any news since April on the discussions? I would also like to have a possibility to setup a `ServiceMonitor` or `PodMonitor` and maybe a `PrometheusRule` from this chart, as many here do. This is a common pattern.

bclermont:
@k3daevin you probably don't need to ask to create a PR and just create a PR :stuck_out_tongue:

CuBiC3D:
@bclermont We already have two of them rejected ;)

azhang:
+1 for PodMonitor

cwrau:
Is there an update for this?

If you're not opposed to the idea anymore, maybe you can merge one of the existing PRs? Like #133?

StianOvrevage:
Moving from nginx-ingress to Traefik and I'm really surprised this is not available out of the box.

dmpe:
Any chance this will be included soon? I see there were some PR that have added service monitors, but on master AFAIK it is not available anymore

PR #425, opened 2021-05-28
✨ add optional ServiceMonitor & PrometheusRule CRDs

dioguerra:
* Add serviceMonitor to select ingress objects
* Add prometheusRules section

Closes: https://github.com/traefik/traefik-helm-chart/issues/156

Signed-off-by: Diogo Guerra <diogo.filipe.tomas.guerra@cern.ch>

### DIFF
--- a/traefik/templates/prometheusrules.yaml
+++ b/traefik/templates/prometheusrules.yaml
@@ -1,4 +1,8 @@
-{{- if .Values.metrics.prometheus.prometheusRule.enabled }}
+{{- if .Values.metrics.prometheus }}
+{{- if .Values.metrics.prometheus.prometheusRule }}
+{{- if (not (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1")) }}
+  {{- fail "ERROR: You have to deploy monitoring.coreos.com/v1 first" }}
+{{- end }}
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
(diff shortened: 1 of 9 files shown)


traefik/templates/prometheusrules.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if and ( .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" ) .Values.metrics.prometheusRule.enabled }}
```
TimDurward, line 1: I'm not too familiar with `.Capabilities`, but wouldn't this require a client connection to k8s to validate the API Group `monitoring.coreos.com/v1`? Wouldn't this break on things like `helm template` if it can't access a cluster?

dioguerra: That is correct.  But this way we guarantee that the chart does not fail if the user has no prometheus because the CRDs do not exist

dincho:
I'm running this in dev environment since few days and it seems to work as expected.
Helm:
```
version.BuildInfo{Version:"v3.4.2", GitCommit:"23dd3af5e19a02d4f4baa5b2f242645a1a3af629", GitTreeState:"dirty", GoVersion:"go1.15.5"}
```
k8s:
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.12-eks-7684af", GitCommit:"7684af4ac41370dd109ac13817023cb8063e3d45", GitTreeState:"clean", BuildDate:"2020-10-20T22:57:40Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

rayjanoka (approved):
LGTM

traefik/templates/prometheusrules.yaml
z0rc: In fact it's counterintuitive. This is an opt-in feature (value should be explicitly set to `true`), so user knows what they are doing. If CRDs aren't available in cluster helm's failure is desirable outcome. Otherwise helm will silently skip the resource, and user will end up with non-working solution.

lexfrei:
Any news?

askz:
Do you need any help to get this merged ? 

I really need this too.

dioguerra:
So, should i drop  .Capabilities.APIVersions.Has "monitoring.coreos.com/v1 then?

z0rc:
@dioguerra please do.

mloiseleur:
Hello @dioguerra,

We can consider merge this PR, but it will need a bit of work. 

Would you rebase it and takes into account that, there are now some options to enable prometheus metrics ?

dioguerra:
Did a quick fix. Still need to retest it. Is this something that looks acceptable?

mloiseleur (changes requested):
I left a few comments. It would be better with some tests, at least to check that by default you won't have ServiceMonitor and PrometheusRule
You may also fail gracefully when users tries to enable PrometheusRule or ServiceMonitor without the desired capabilites. See  #646 for an example if you need.

charlie-haley (approved):
LGTM

mloiseleur:
I updated the PR, rebased and added some tests. 
Before we merge it, anything we missed ?

mloiseleur (approved):

dioguerra:
@mloiseleur  thanks for making the final push, i could not sorry...

PR #429, opened 2021-06-08
Make it possible to have a udp service without a tcp service and vise…

KDSBest:
… versa


kevinpollet:
Hello @KDSBest,

Thanks for your contribution.

The UDP and TCP services are only created when ports with the corresponding protocols are configured, thus it is not needed to disable TCP/UDP services creation.

According to this, we will decline this pull request.

PR #439, opened 2021-07-03
Add support for apiextensions v1 CRDs

### LINKED ISSUE #438: apiextensions.k8s.io/v1beta1 is deleted in Kubernetes v1.22
I am preparing K3s for Kubernetes v1.22, and it appears that the Traefik Helm chart is still using `v1beta1` CRDs. This API version is going away in v1.22; the Helm charts do not work in their current state. 

As per https://github.com/kubernetes/sig-release/tree/master/releases/release-1.22 Kubernetes 1.22 is expected Wednesday, August 4th; it would be nice if the Traefik helm charts supported this release on day 1. beta.0 is out now, for anyone who wants to experiment with it.

brandond:
Autodetect v1 CRD type support

Resolves #438

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 9.19.2
+version: 9.20.0
 appVersion: 2.4.8
 keywords:
   - traefik
--- a/traefik/crds/ingressroute.yaml
+++ b/traefik/crds/ingressroute.yaml
@@ -1,10 +1,29 @@
+{{- if .Capabilities.APIVersions.Has "apiextensions.k8s.io/v1/CustomResourceDefinition" }}
+apiVersion: apiextensions.k8s.io/v1
+{{- else }}
 apiVersion: apiextensions.k8s.io/v1beta1
+{{- end }}
 kind: CustomResourceDefinition
 metadata:
   name: ingressroutes.traefik.containo.us
 spec:
   group: traefik.containo.us
+{{- if .Capabilities.APIVersions.Has "apiextensions.k8s.io/v1/CustomResourceDefinition" }}
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+    schema:
+      openAPIV3Schema:
+        type: object
+        properties:
+          spec:
+            x-kubernetes-preserve-unknown-fields: true
+          status:
+            x-kubernetes-preserve-unknown-fields: true
+{{- else }}
   version: v1alpha1
+{{- end }}
   names:
     kind: IngressRoute
     plural: ingressroutes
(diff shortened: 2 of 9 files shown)


faust64:
I'm not sure there's a need to include a schema, only to not define one and tell it to preserve unknown fields.Though it won't hurt.
I didn't react when 1.22 got released, my bad. I did prepare something simpler: https://github.com/traefik/traefik-helm-chart/pull/440/commits/b1d49e77770b76b7648c5b715e561f794dedc03a
But sure, let's keep yours, would be better to remain compatible with older versions.

brandond:
Anything else you need from me?

jakubhajek (approved):
Looks good to me, thank you for contributing to the project.

faust64:
Looks good to me, thanks!

ldez (approved):

onedr0p:
Unfortunately having the CRDs converted into a Helm template is breaking our use with Flux. We need to apply the CRDs before traefik is installed or we run into race conditions where the CRDs aren't present but we need to apply an ingress-route. 

v1beta1 was deprecated in 1.16 (almost 2 years ago) shouldn't this mean we don't need to condition for it?

jakubhajek:
hey @onedr0p, I also use Flux to maintain a cluster with Traefik and I haven't experienced any issue, yet. Would you please share what exact errors do you get? Thank you,

onedr0p:
Hi @jakubhajek as you can see in the following link I pull in the CRDs for traefik before the helm chart is applied:

https://github.com/onedr0p/home-cluster/blob/main/cluster/crds/traefik/crds.yaml

toboshii:
@jakubhajek The issue arises if you have ingress-routes or other Traefik CRDs in cluster that Flux attempts to reconcile BEFORE the Traefik helm deployment reconciles (which will always be the case on a fresh cluster since Flux runs kustomization first).

For this reason (among others) it's generally recommended that when using Flux CRDs be deployed by kustomization and not helmreleases (as @onedr0p linked in his example). This is kind of difficult to do if the only copy of the CRDs publicly available are now full of tempates...

toboshii:
Also this even just breaks manual helm deployments with helm 3 #441

anthr76:
It would be great if traefik can provide a kustomization for CRDs Something like 

https://github.com/fluxcd/flux2/blob/main/manifests/crds/kustomization.yaml

With flux, and I’m sure other gitops clusters. It’s critical for your cluster to have the custom resources in place *before* the custom resource gets applied otherwise you can land in a dead lock.

jakubhajek:
Thank you @anthr76 @onedr0p @toboshii for your valuable comments, we are on that and will keep you updated as soon as we will have the final solution.

brandond:
Sorry for the breakage y'all. In k3s we preprocess the helm chart and split the CRDs out into a separate chart instead of just using the built-in helm CRD support, so I didn't run into this problem when testing it locally.

Issue #446, opened 2021-07-09
Support configuring the GatewayAPI's Gateway namespace


ah-f3:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

The `Gateway` is currently installed into the `default` namespace: 

Can a config value be added here: https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml#L80-L86 to allow configuring this namespace?

https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/gateway.yaml#L7

Alternatively,  a flag to not install the Gateway and allow a user to install one outside the chart (possibly also not installing the GatewayClass?)

juranir:
Did you tried to use the parameter "-n NamespaceName"?
Ex. helm install MyPackage -n MyNamespace

ah-f3:
https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/gateway.yaml#L7 explicitly shows `default` so i'm not sure helm will override it.

Additionally, I don't want the Gateway in the namespace im installing traefik into.

faust64:
Indeed, this could be worth patching.
If you're up for it, feel free to submit a PR.
I'll look into it otherwise.

PR #447, opened 2021-07-10
feat(gateway): introduces param / pick Namespace installing Gateway

faust64:
### What does this PR do?

Allows to set a custom Namespace name creating Traefik initial Gateway

### Motivation

As reported by @ah-f3
Gateway is currently created into the default Namespace, regardless of where Traefik is being deployed to.
By default, we could want to install it alongside Traefik, into the same Namespace.
Though it could also make sense to put it somewhere else (default, kube-system, ...).
Could make sense to let end-user decide what's best for them..

Fixes https://github.com/traefik/traefik-helm-chart/issues/446

### More

### Additional Notes

Maybe keep the default Gateway namespace to "default", preserving the current behavior?
Gateway being experimental and relatively recent, we could probably still change our mind on what the default should be.
Whichever. It's not too late to un-comment the `experimental.kubernetesGateway.namespace: default` from values. Let me know.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.0.0
+version: 10.0.1
 appVersion: 2.4.9
 keywords:
   - traefik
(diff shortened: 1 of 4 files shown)


ah-f3 (commented):
:heart:

dtomcej (approved):
LGTM
:shipit:

PR #448, opened 2021-07-10
fix: missing service labels

faust64:
### What does this PR do?

Adds labels to the UDP service, according to values file.

### Motivation

Looking into something else, I realize that the labels we may configure in values file would only be added to the main TCP Service.
I would assume they should also be set on the UDP one.

### More

### Additional Notes

N/A


jakubhajek (approved):
Thanks for that fix, looks good to me.

jakubhajek (approved):
Looks good to me.

Issue #436, opened 2021-06-29
Add additional service for internal IP


muffl0n:
We're running a private GKE cluster that is reachable via internal IPs (via VPN) as well as via external IPs (via external Loadbalancer). We currently use ingress-nginx where you can configure an additional internal service. See:
* https://github.com/kubernetes/ingress-nginx/blob/master/charts/ingress-nginx/templates/controller-service.yaml
* https://github.com/kubernetes/ingress-nginx/blob/master/charts/ingress-nginx/templates/controller-service-internal.yaml

```
controller:
  service:
    internal:
      enabled: true
      annotations:
        networking.gke.io/internal-load-balancer-subnet: gke-services
        networking.gke.io/load-balancer-type: Internal
      loadBalancerIP: 10.10.10.10
```

The benefit is, that we can allow some Ingresses to only be accessible via private IPs.

I could not find a smiliar possibility to do this in Traefik. Am I missing something?

yogeek:
Hi ! Also interested by this feature.
Any update or roadmap please ?

PR #449, opened 2021-07-10
feat(internal-service): adds an optional "internal" service

faust64:
### What does this PR do?

Introduces an optional internal Service, that could be used distinguishing "internal" accesses (eg: company LAN) from public accesses.

### Motivation

As suggested by @muffl0n
See https://github.com/traefik/traefik-helm-chart/issues/436

### More

### Additional Notes

N/A

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.0.0
+version: 10.0.1
 appVersion: 2.4.9
 keywords:
   - traefik
--- a/traefik/templates/service-internal.yaml
+++ b/traefik/templates/service-internal.yaml
@@ -1,9 +1,9 @@
-{{- if .Values.service.enabled -}}
+{{- if .Values.service.internal.enabled -}}
 
 {{ $tcpPorts := dict }}
 {{ $udpPorts := dict }}
 {{- range $name, $config := .Values.ports }}
-  {{- if $config.expose }}
+  {{- if $config.exposeInternal }}
     {{- if eq (toString $config.protocol) "UDP" }}
       {{ $_ := set $udpPorts $name $config }}
     {{- else }}
@@ -16,29 +16,32 @@
 apiVersion: v1
 kind: List
 metadata:
-  name: {{ template "traefik.fullname" . }}
+  name: {{ template "traefik.fullname" . }}-internal
 items:
 {{- if $tcpPorts }}
   - apiVersion: v1
     kind: Service
     metadata:
-      name: {{ template "traefik.fullname" . }}
+      name: {{ template "traefik.fullname" . }}-internal
       labels:
         app.kubernetes.io/name: {{ template "traefik.name" . }}
         helm.sh/chart: {{ template "traefik.chart" . }}
         app.kubernetes.io/managed-by: {{ .Release.Service }}
         app.kubernetes.io/instance: {{ .Release.Name }}
-      {{- with .Values.service.labels }}
+      {{- $svclabels := default .Values.service.labels .Values.service.internal.labels }}
+      {{- with $svclabels }}
       {{- toYaml . | nindent 8 }}
       {{- end }}
       annotations:
-      {{- with .Values.service.annotations }}
+      {{- $svcannotations := default .Values.service.annotations .Values.service.internal.annotations }}
+      {{- with $svcannotations }}
       {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
       {{- $type := default "LoadBalancer" .Values.service.type }}
-      type: {{ $type }}
-      {{- with .Values.service.spec }}
+      {{- $typeint := default $type .Values.service.internal.type }}
+      type: {{ $typeint }}
+      {{- with .Values.service.internal.spec }}
       {{- toYaml . | nindent 6 }}
       {{- end }}
       selector:
@@ -54,36 +57,42 @@ items:
         nodePort: {{ $config.nodePort }}
         {{- end }}
       {{- end }}
-      {{- if eq $type "LoadBalancer" }}
-      {{- with .Values.service.loadBalancerSourceRanges }}
+      {{- if eq $typeint "LoadBalancer" }}
+      {{- with .Values.service.internal.loadBalancerSourceRanges }}
       loadBalancerSourceRanges:
       {{- toYaml . | nindent 6 }}
       {{- end -}}
       {{- end -}}
-      {{- with .Values.service.externalIPs }}
+      {{- with .Values.service.internal.externalIPs }}
       externalIPs:
       {{- toYaml . | nindent 6 }}
       {{- end -}}
 {{- end }}
 
-{{- if  $udpPorts }}
+{{- if $udpPorts }}
   - apiVersion: v1
     kind: Service
     metadata:
-      name: {{ template "traefik.fullname" . }}-udp
+      name: {{ template "traefik.fullname" . }}-internal-udp
       labels:
         app.kubernetes.io/name: {{ template "traefik.name" . }}
         helm.sh/chart: {{ template "traefik.chart" . }}
         app.kubernetes.io/managed-by: {{ .Release.Service }}
         app.kubernetes.io/instance: {{ .Release.Name }}
+      {{- $svclabels := default .Values.service.labels .Values.service.internal.labels }}
+      {{- with $svclabels }}
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
       annotations:
-      {{- with .Values.service.annotations }}
+      {{- $svcannotations := default .Values.service.annotations .Values.service.internal.annotations }}
+      {{- with $svcannotations }}
       {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
       {{- $type := default "LoadBalancer" .Values.service.type }}
-      type: {{ $type }}
-      {{- with .Values.service.spec }}
+      {{- $typeint := default $type .Values.service.internal.type }}
+      type: {{ $typeint }}
+      {{- with .Values.service.internal.spec }}
       {{- toYaml . | nindent 6 }}
       {{- end }}
       selector:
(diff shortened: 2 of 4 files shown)


muffl0n:
Awesome! Can't wait to give it a try! Thank you! 😘

dtomcej:
Although this PR creates a new service that fronts Traefik, it doesn't provide any differentiator to Traefik that the request came through the internal service instead. If this is just a duplicate service, I'm not convinced it should be bundled in the chart, as it doesn't really provide any further functionality.

For example, if you have:

```
Service A (external) --> Traefik web EP
and
Service B (internal) --> Traefik web EP
```
Traefik won't know (or care) whether the request came from an external service. This would require additional entrypoints, and wiring the entrypoints to the same service ports forwarded by the internal service.

To accomplish this, you would require:
```
Service A (external) --> Traefik web EP
and
Service B (internal) --> Traefik web-internal EP
```

Thoughts?

faust64:
if you have:

```
Service A (external) --> Traefik web EP
and
Service B (internal) --> Traefik web EP
```

This is one possibility. If I have two LB (metallb, ELB, ...) or one LB and a NodePort, or one NodePort and a ClusterIP, ... exposing the same endpoints doesn't mean you're exposing it to the same clients. By extension, exposing it with different services mean you may apply different firewalling/filtering.

Though for sure, trusting DNS resolution and selective firewalling would not prevent someone to query your public IPs, forging the host header of his request to access a service that would otherwise only be exposed using a private FQDN / private LB.

Now, that PR already implements the case for:

```
Service A (external) --> Traefik web EP
and
Service B (internal) --> Traefik web-internal EP
```

Which is precisely what was requested from https://github.com/traefik/traefik-helm-chart/issues/436

This would be done with the following values:

```
ports:
  traefik:
   [...]
  web:
    port: 8000
    expose: true
    exposeInternal: true
    redirectTo: websecure
  webinternal:
    port: 8001
    expose: false
    exposeInternal: true
  [...]
  websecure:
    port: 8443
    expose: true
    exposeInternal: true
  [...]
  websecureinternal:
    port: 8001
    expose: false
    exposeInternal: true
  [...]

service:
  enabled: true
  annotations:
    networking.gke.io/internal-load-balancer-subnet: gke-services
  type: LoadBalancer
  internal:
    enabled: true
    annotations:
      networking.gke.io/internal-load-balancer-subnet: gke-services
      networking.gke.io/load-balancer-type: Internal
    loadBalancerSourceRanges:
    - 10.42.0.0/16
    type: LoadBalancer
```

jakubhajek:
I like that feature and I see the need of adding Internal Service.
I run Traefik together with MetalLB and I have two IP pools: private and public. I also created additional entry points for Web internal, because some of the services need to reach Traefik internally. However, the Helm Chart provides only one service to expose ports.  If I would like to expose e.g. port metrics and currently, I can only manually create a service and expose it locally by binding it to the proper MetalLB pool. 

Here is the example service that I've created:

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    metallb.universe.tf/address-pool: private-ips
  labels:
    app: traefik
spec:
  loadBalancerIP: 172.16.16.1
  ports:
  - name: web-int
    port: 80
    protocol: TCP
    targetPort: web-int
  - name: web-int-tls
    port: 443
    protocol: TCP
    targetPort: web-int-tls
  selector:
    app.kubernetes.io/instance: traefik-traefik
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - ip: 172.16.16.1
 ```
It would ideally be if Helm chart could create that service for me and I could decide what service the specific ports should be exposed to, whether internal or external. 

Thanks for that contribution.

File: traefik/Chart.yaml [hunk 1]
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.1.6
+version: 10.1.7
```
jakubhajek, line 5: Please update the chart and Traefik version.

traefik/values.yaml
```diff
@@ -323,9 +355,18 @@ ports:
 #       - CurveP384
 tlsOptions: {}
 
-# Options for the main traefik service, where the entrypoints traffic comes
-# from.
+# Options for the traefik services, where the entrypoints traffic comes from.
```
jakubhajek, line 358: Options for the Traefik services, where the entrypoints traffic comes from.

traefik/values.yaml
```diff
@@ -323,9 +355,18 @@ ports:
...
+# Two services may be created: the first would target entries from your
+# `.Values.ports`, where `.expose` is `true`. The latter would include entries
+# whose `.exposeInternal` is `true`. Both of which are undefined by default,
+# assumed to be false.
+# In addition to having enabled those services, there need to be at least one
+# elligible entrypoint, for the corresponding Service object to be created. By
```
jakubhajek, line 364:
```suggestion
# eligible entrypoint, for the corresponding Service object to be created. By

File: traefik/values.yaml [hunk 4]
```diff
@@ -343,6 +384,25 @@ service:
...
+    enabled: false
+    # Service type for internal Service, defaults to `.Values.service.type`
+    # type: ClusterIP
+    # Additional labels for internal Service, defaults to `.Values.service.labels`
+    # labels: {}
+    # Additional annotations for internal Service, defaults to
+    # `.Values.service.annotations`
+    # annotations:
+    #   networking.gke.io/internal-load-balancer-subnet: gke-services
+    #   networking.gke.io/load-balancer-type: Internal
```
jakubhajek, line 400: May we please add MetalLB annotations as another example to use together with private IP pools? What do you think?
`metallb.universe.tf/address-pool: private-ips`

jakubhajek (changes requested):
There are minor changes that should be applied and we are ready to merge the PR. 
Thanks a lot for that feature.

traefik/values.yaml
faust64: Sure

traefik/values.yaml
faust64: ok

traefik/Chart.yaml
faust64: done

traefik/templates/service-internal.yaml
```diff
@@ -0,0 +1,123 @@
...
+      {{ $_ := set $udpPorts $name $config }}
+    {{- else }}
+      {{ $_ := set $tcpPorts $name $config }}
+    {{- end }}
+  {{- end }}
+{{- end }}
+
+{{- if or (gt (len $tcpPorts) 0) (gt (len $udpPorts) 0) }}
+apiVersion: v1
+kind: List
```
jakubhajek, line 17: What is this List Resource for?

traefik/templates/service.yaml
```diff
@@ -3,19 +3,22 @@
...
+      {{ $_ := set $udpPorts $name $config }}
+    {{- else }}
+      {{ $_ := set $tcpPorts $name $config }}
+    {{- end }}
   {{- end }}
 {{- end }}
 
+{{- if or (gt (len $tcpPorts) 0) (gt (len $udpPorts) 0) }}
 apiVersion: v1
 kind: List
```
jakubhajek, line 17: Same here, would you please explain what is this List for?

jakubhajek (changes requested):
Can you please rebase the PR and explain the List resource? I am not familiar with that resource and the documentation is poor around that feature. Thank you

traefik/templates/service.yaml
faust64: A List can include one or more objects, with no distinction on type -- we could concatenate most templates into a single file.

Another way to write it would have been to add separators (`---`) between objects, though I'm not sure unit tests would be able to check specific objects, as we currently do (`items[$index].(spec|metadata|...)`).

Arguably, we could have used separate templates for UDP and TCP services: there's not a lot of shared code, which could have been a reason for using a List.
Maybe there's another explanation I'm missing?
Otherwise I can split them and fix tests going with one file per Service.

traefik/templates/service-internal.yaml
faust64: Copy/pasted from the main Service template file: creating one or more objects from a single template file, while allowing unit tests to access those objects individually, querying a specific index from that List items array.

traefik/templates/service.yaml
jakubhajek: Thanks for explaining that. So, we need to just rebase the PR and then we can merge it.

WojoInc:
Would this work with the Kubernetes Ingress provider? Trying to mimic this manually (create separate ports, service, etc), but it seems like even if I use the appropriate annotations to set the entrypoint, any ingresses that are created are assigned the status (load balancer IP/hostname) from the "public" service.

Based on what I currently have set up with nginx, I imagine one would still need to deploy two copies of traefik with different IngressClasses to achieve the same public/private split when using the Ingress provider.

faust64:
As you point out, you can annotate your Ingress designating which entrypoints should be used exposing their service -- while the ingressroute allows you do to this setting an array in its spec.

Sadly, I don't have a cluster with a cloud provider to test this with, if you know anything please share.
Can anyone check?

>  if I use the appropriate annotations to set the entrypoint, any ingresses that are created are assigned the status (load balancer IP/hostname) from the "public" service.

In terms of traffic routing, as far as I've seen, entrypoints annotations are properly observed by Traefik. If my Ingress annotated restricting is service to entrypoint A, trying to query it through entrypoint B would not work, as expected.

Now, in terms of status, I am not sure.
If there actually is an issue here, then I guess you should report this to github.com/traefik/traefik .
There may be a bug or some edge case, within Traefik code itself. I don't think there is much to be done in Helm charts.

WojoInc:
Based on [this](https://doc.traefik.io/traefik/providers/kubernetes-ingress/#ingressendpoint), it seems like its a limitation of Traefik itself/Ingress in general. All ingresses for a given Ingress class get pointed a single endpoint. Deploying two instances of Traefik (each with their own IngressClass) seems to work like a charm however. Thanks for taking a look!

mloiseleur:
Superseded by #683

PR #450, opened 2021-07-12
feat: allow user to configure terminationGracePeriodSeconds

js-timbirkett:
### What does this PR do?

Hi 👋  - This PR enables users to set a custom `terminationGracePeriodSeconds` in their values. This will allow users who have Traefik sat behind load-balancers using keepAlive connections with idle timeouts higher than 60 seconds.

Through the use of `requestAcceptGraceTimeout`, `graceTimeOut` and `terminationGracePeriodSeconds`, users can avoid 502 and other timeout errors on re-deployment of Traefik.

### Motivation

We have ALBs that have an idle timeout of 5 minutes pointing to a Traefik `NodePort` service. In an effort to improve performance and reduce the number of connections to a service, the load balancers setup persistent TCP connections to the `NodePort`. On re-deployment of Traefik, the NAT'd connections from the ALBs to the old pod endpoints are still tracked in `conntrack` entries. This causes 502 and other timeout-related errors until the ALBs timeout the connections and they drop out of the `conntrack` tables.

In our case, this causes minutes of outages across all services and has caused hours of troubleshooting and investigation to understand the problem. 

- [x] Yes, I updated the chart version


js-timbirkett:
It turns out that reading this: https://learnk8s.io/kubernetes-long-lived-connections could have saved me hours of troubleshooting :grimacing: :lolsob:

js-timbirkett:
I hadn't seen this open as a PR in #411 🤦 - This PR probably means that one can be closed off.

js-timbirkett:
@SantoDE - any chance of casting an eye over this PR? It's pretty much the same as #411 with tests :)

js-timbirkett:
Trying to keep these long open PRs up to date with master is a fun game :)

js-timbirkett:
Closing due to inactivity of maintainers.

PR #452, opened 2021-07-19
Add metrics framework, and prom annotations

dtomcej:
### What does this PR do?

Adds a new entrypoint for metrics, and creates a base framework for metrics configuration

### Motivation

Simple prometheus implementation

### More

### Additional Notes

This PR implements a port lookup based on the name of the port. This could also be used to enable metrics on the Traefik EP if users wanted to, similar to the old chart behavior.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.0.2
+version: 10.1.0
 appVersion: 2.4.9
 keywords:
   - traefik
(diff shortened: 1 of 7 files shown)


SantoDE (approved):
LGTM

pysysops:
@SantoDE - Did this get tested with Prometheus? I'm pretty sure that the annotations need to be on the pods and not the deployment and daemonset objects.

pysysops:
The relevant Prometheus configuration that pretty much everyone uses:

https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml#L1608-L1650

Issue #309, opened 2020-11-16
IngressClass not working with helm template


heruscoder:
Hello, in chart version `9.10.1` was pushed a fix into IngressClass Resource #300 

The problem is, if i try to `helm template traefik/traefik --set ingressClass.enabled=true` i get this error: `<fail "\n\n ERROR: You must have atleast networking.k8s.io/v1beta1 to use ingressClass">` 

If i try to `helm template traefik/traefik --set ingressClass.enabled=true --api-versions networking.k8s.io/v1beta1/IngressClass`, it works.

The problem is, i cannot specify which `--api-versions` to use because we are using ArgoCD to manage our helm installs, and Argo by default only populates `--api-versions` with the Capabilities it can find.

I'm gonna use `9.10.0` for now, as it works.

`Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:32:58Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}`

Gallardo994:
Same issue with managed kubernetes in DigitalOcean (DOKS), also using ArgoCD to deploy Traefik 2.3.5. Tested with 1.18-2.do and 1.19-3do

mcavoyk:
Experiencing similar issue on ArgoCD 1.8.1 deploying Traefik chart 9.11.0. I can see in ArgoCD logs that `--api-versions networking.k8s.io/v1 --api-versions networking.k8s.io/v1beta1` is passed to the `helm template` command, but not the specific resources within the api-version.

Helm docs seem to indicate that both the version and resource should be available through [Capabilities.APIVersions](https://helm.sh/docs/chart_template_guide/builtin_objects/). This seems like possibly a bug in ArgoCD for not presenting each resource within the apiversion.

huang12zheng:
I also met. Is there any latest news?

huang12zheng:
[...]

PR #454, opened 2021-07-19
Add ipFamilyPolicy to service spec, proper support for dual-stack clusters

### LINKED ISSUE #453: Dual-Stack cluster, Helm chart results in SingleStack deployment
### Welcome!

10.0.2

2.4.9

### What did you do?

1. Install dual-stack Kubernetes cluster.
2. Install Calico
3. Install Traefik using Helm

### What did you see instead?

I saw that the deployment from `helm install traefik traefik/traefik` results in a SingleStack deployment, and it doesn't get a IPv6 address when it should.

Platform/architecture: Raspberry Pi 4, running Ubuntu 20.04 LTS (arm64)
Kubernetes version: 1.21.x

Default Helm traefik values.

### Additional Information
After deployment I executed `kubeadm get svc traefik -o yaml` and the resulting output is:

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    meta.helm.sh/release-name: traefik
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2021-07-19T20:33:08Z"
  labels:
    app.kubernetes.io/instance: traefik
    helm.sh/chart: traefik-10.0.2
  name: traefik
  namespace: default
  resourceVersion: "1335"
  uid: 831207c4-e46e-47d5-9c0a-6b85032cd36a
spec:
  clusterIP: 10.96.144.105
  clusterIPs:
  - 10.96.144.105
  externalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: web
    nodePort: 31611
    port: 80
    protocol: TCP
    targetPort: web
  - name: websecure
    nodePort: 30437
    port: 443
    protocol: TCP
    targetPort: websecure
  selector:
    app.kubernetes.io/instance: traefik
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer: {}
```
This shouldn't have ipFamilies as IPv4 when the cluster is dual-stack configured, nor have ipFamilyPolicy as singleStack.

samip5:
### What does this PR do?

This PR adds `ipFamilyPolicy` to service spec so that if cluster has IPv6 automatic address assignment in use, it will also get one according to kubernetes configuration.

### Motivation

The motivation for this change was that without it, kubernetes will default to IPv4-only deployment even on full networking stack cluster.

### More

### Additional Notes

Closes #453


jakubhajek:
Hello  @samip5 

Thanks a lot for contributing to this project. 

The `ipFamilyPolicy `feature you added is available in Kuberbetens v1.21, so users who will deploy the chart on older Kubernetes version will face the following issue:

```
Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(Service.spec): unknown field "ipFamilyPolicy" in io.k8s.api.core.v1.ServiceSpec
```

Would you please add the condition to check what Kubernetes version the Helm chart is going to be deployed? 

Thank you,

samip5:
> Hello @samip5
> 
> Thanks a lot for contributing to this project.
> 
> The `ipFamilyPolicy `feature you added is available in Kuberbetens v1.21, so users who will deploy the chart on older Kubernetes version will face the following issue:
> 
> ```
> Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(Service.spec): unknown field "ipFamilyPolicy" in io.k8s.api.core.v1.ServiceSpec
> ```
> 
> Would you please add the condition to check what Kubernetes version the Helm chart is going to be deployed?
> 
> Thank you,

I'm sorry, but I'm not sure how to do that?

faust64:
Dual stack support was introduced in 1.20 (https://kubernetes.io/blog/2020/12/08/kubernetes-1-20-release-announcement/#alpha-with-updates-ipv4-ipv6).
Reached beta in 1.21 (https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/dual-stack-support/).
Should be enabled by default as of 1.21 (https://kubernetes.io/docs/concepts/services-networking/dual-stack/#enable-ipv4ipv6-dual-stack).

It doesn't sound safe to just inject that param in the chart defaults right now. not every cluster would support this. Besides, I'm not sure upgrading an existing cluster to 1.21 would enable that feature.

samip5:
What I mean is that I don't know how to check the version in helm values/charts.

faust64:
See https://helm.sh/docs/chart_template_guide/builtin_objects/

faust64:
Although beware we've had reports regarding some tools (eg: ArgoCD) that somewhat would not have the same Helm internal variables than we usually do (such as #309).
The default really should be not to configure dualstack, at least right now.
If you want to set it, you can, in your own values.
Any reason you would want to expose your Ingress using IPv6?

Also, this brings to my mind, a post I saw on stack overflow (https://serverfault.com/a/1074822/293779)
You won't be able to create dual-stack LoadBalancer services. This is usually done with two individual services.
IMHO, we shouldn't set that option at all, or just force it to IPv4-only, and offer an option to add an IPv6 Service, if supported, when requested.

samip5:
> Any reason you would want to expose your Ingress using IPv6?

IPv4 is legacy, and IPv6-only is becoming the new normal, slowly but surely.
I require ingress to be dual-stacked as users connecting to it, will mostly prefer IPv6 as you should if you have IPv6 connectivity.

> Also, this brings to my mind, a post I saw on stack overflow (https://serverfault.com/a/1074822/293779)
> You won't be able to create dual-stack LoadBalancer services. This is usually done with two individual services.

Unfortunately, you're incorrect in that regard. I do have an LB with both on the single LB, even if that is though ExternalIPs.
The poster on stack overflow didn't specify how it "does not" support it.

```
NAME      TYPE           CLUSTER-IP     EXTERNAL-IP                              PORT(S)                                     AGE
traefik   LoadBalancer   10.43.15.210   2001:67c:1104:xxxx::443,185.xxx.xx.250   9000:31233/TCP,80:30075/TCP,443:31378/TCP   11d
```
> IMHO, we shouldn't set that option at all, or just force it to IPv4-only, and offer an option to add an IPv6 Service, if supported, when requested.

Why? Kubernetes already defaults to IPv4-only services and my change only effects the service if services have been configured to support dual-stack operation.

faust64:
> IPv4 is legacy, and IPv6-only is becoming the new normal, slowly but surely.

Yeah, that's what I've been told, as a student, like 20 years ago.
We even had the IPv6 day, t-shirts and all, 10 years ago.
For sure, it's been a success.

> The poster on stack overflow didn't specify how it "does not" support it.

Then again, you're not the first claiming "it works on my laptop".
I wouldn't know, I don't operate any cluster that would offer that feature today.

As a general rule, being a beta feature doesn't mean it would work with all cloud providers, all SDNs, or any kubernetes setup.
1.21 is the first version where that option is enabled by default. From now on, we should see Kubernetes administrator reporting outstanding issues, ... as of today: we do not have any return of experience about this, it's not a GA feature yet: you could tell me it works or it's broken, ... who knows, probably something in between.

> Why? Kubernetes already defaults to IPv4-only services

Well, precisely, Kubernetes already defaults to IPv4-only.
If adding an option on our own, we would better match kube default, instead of introducing something end-user may not have expected.

> my change only effects the service if services have been configured to support dual-stack operation

Well, you are incorrect in that regard: you are changing the values file.
As jakubhajek told us: it's broken, as of right now.

You can't guarantee a user applying those charts today would have k8s 1.21, nor that this feature gate would be enabled.

Maybe we don't need two services with all IAAS/SDN? Maybe some would.
Either way, we shouldn't set that option by default: Anyone applying that patched Chart, would automatically have this option set, which, again, is not supported by all versions of Kubernetes.

If adding that option to your own values file is not acceptable, you could make sure your patch won't break Traefik deployments on cluster that do not support dual-stack.

samip5:
Alright, I'm not entirely sure if that's the proper way to do the check, but comments?

faust64:
I'm having similar issues with another PR (https://github.com/traefik/traefik-helm-chart/pull/466#issuecomment-901784312).
You can test on your workstation with `make test` (assuming helm CLI and docker runtime are present).

The helm-unittest plugin is installed from a rancher repo, which hasn't been updated in a while.
The `.Capabilities.KubeVersion` is something that was introduced in a fork, AFAIU (quintush/helm-unittest).

Switching to that fork and fixing confs, ... I've seen regressions in the existing traefik-helm-chart tests, which definitely looks like bugs in that quintush fork. I'ld need to investigate further ...

PR #457, opened 2021-07-20
Version bumped 2.4.13

jakubhajek:
### What does this PR do?

Upgrading Traefik to 2.4.13

### Motivation

The new version of Traefik has been recently released. 

### More

### Additional Notes

N/A

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.1.1
-appVersion: 2.4.9
+version: 10.1.2
+appVersion: 2.4.11
 keywords:
   - traefik
   - ingress
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -3,7 +3,7 @@ name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
 version: 10.1.2
-appVersion: 2.4.11
+appVersion: 2.4.13
 keywords:
   - traefik
   - ingress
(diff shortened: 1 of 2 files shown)


Tchoupinax:
Hello,

What's up with this PR ? :) The bumped version may be increased to 2.4.12 (<https://github.com/traefik/traefik/releases/tag/v2.4.12>)

mnorrsken:
Afaik there is breaking changes between 2.4.9 and later:
v2.4.9 to v2.4.10
K8S CrossNamespace
In v2.4.10, the default value for allowCrossNamespace has been changed to false.

K8S ExternalName Service
In v2.4.10, by default, it is no longer authorized to reference Kubernetes ExternalName services. To allow it, the allowExternalNameServices option should be set to true.

It would be nice if the chart values had settings to enable these instead of having to use additonalArguments

It could even be a good idea to set these to "true" in the chart, or at least make a note about this when upgrading from a pre 2.4.11 version

docwhat:
This should probably be bumped in priority due to the security advisory [_Drop Headers via Malicious Connection Header_](https://github.com/traefik/traefik/security/advisories/GHSA-m697-4v8f-55qg)

SantoDE (approved):
LGTM :)

Tchoupinax:
According to the message of @mnorrsken, there is no change about the breaking change came with 2.4.10 ?

mnorrsken:
To keep the 2.4.9 defaults, add this to values.yaml:

```
additionalArguments:
  - "--providers.kubernetescrd.allowcrossnamespace=true"
  - "--providers.kubernetescrd.allowexternalnameservices=true"
  - "--providers.kubernetesingress.allowexternalnameservices=true"
```

Of course review if you really need these options first.

PR #475, opened 2021-08-13
Allow setting TCP and UDP service annotations separately

nicolaschan:
### What does this PR do?

Adds the following two options under `service`:
- `annotationsTCP: {}`
- `annotationsUDP: {}`

These annotations are applied to either the TCP or UDP service *only*. The option `service.annotations` remains and keeps the existing behavior of applying annotations to both services.

### Motivation

I am using [kubernetes-sigs/external-dns](https://github.com/kubernetes-sigs/external-dns) to set DNS records for my traefik services. DNS records are specified by annotating the service. Traefik creates TCP and UDP LoadBalancer services and each is assigned a different IP address by [metallb](https://metallb.universe.tf/). I would like to set DNS records independently for the TCP service and UDP service, but this is not currently possible because `service.annotations` applies to *both* services.

There are likely other cases where you would want different annotations between the TCP and UDP services.

### More

### Additional Notes

Open to design suggestions. I considered doing something like:
```yaml
service:
  annotations:
    tcp: {} # TCP-only annotations here
    udp: {} # UDP-only annotations here
```
This would break existing behavior. I think it is better to keep existing values working, so that is why I added `annotationsTCP` and `annotationsUDP` where I did.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.1.2
+version: 10.2.0
 appVersion: 2.4.13
 keywords:
   - traefik
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.1.7
+version: 10.2.0
 appVersion: 2.4.13
 keywords:
   - traefik
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.2.0
+version: 10.1.7
 appVersion: 2.4.13
 keywords:
   - traefik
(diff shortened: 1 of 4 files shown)


traefik/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.1.2
+version: 10.2.0
```
jakubhajek, line 5:
```suggestion
version:  10.1.7
```

jakubhajek (commented):
Thank you a lot for contributing to Traefik Helm Chart. I played a little bit with your PR and haven't experienced any issues, so we should be fine with merging that. Can you please just update the version of the chart?

jakubhajek (approved):
LGTM

traefik/Chart.yaml
nicolaschan: Chart version updated

nicolaschan:
@jakubhajek Sorry, looks like there was a merge conflict on the version. I've fixed it, so might need re-approval.

jakubhajek (approved):
Ready to be merged ;)

traefik/Chart.yaml
jakubhajek: My mistake. We are following semver, then: 

- enhancement: 1.X.0
- bug: 1.1.X
- breaking: x.0.0

The version you initially set was correct as it is an enhancement.

jakubhajek (approved):
The chart version is set to 10.2.0 as we are adding a new enhancement. Looks good to me.

Issue #479, opened 2021-08-17
Possibility to change readiness/liveness probes through values


andreasvikke:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

I've run into a problem where some pods restart due too high demand in node. The httpGet probes get timeout cause it's too slow to startup on node. Therefore i suggest the possibility to change readiness/liveness probes trough values.

This is the section of code that would be nice to change. I'd like to change the httpGet to a exec with curl like explained in this comment https://github.com/kubernetes/kubernetes/issues/89898#issuecomment-876539217
https://github.com/traefik/traefik-helm-chart/blob/bf315e74c8132ab74a5567b6cbf217e59a724c3c/traefik/templates/_podtemplate.tpl#L45-L62

jakubhajek:
We discussed that issue on PR #480 and replacing httpGet health probe to exec does not solve the problem described in that issue. No need to change the existing healthprobes.

PR #480, opened 2021-08-18
Allow changing the readiness/liveness probes

AndreasVikke:
### What does this PR do?
This will add the following two options under `ports>traefik`:
- `readinessProbe: {}`
- `livenessProbe: {}`

If any of these is set the default httpGet will be replace the corresponding pod probe.
The examples show how an exec with wget can be used instead.

### Motivation
Information about motivation can be found in the issue https://github.com/traefik/traefik-helm-chart/issues/479

### More

### Additional Notes
Please let me know if the solution is a bit too "open". I'd just like to switch the httpGet with exec, which can be done with less changes to values.

```yaml
readinessProbe: {}
  exec:
    command:
      - "wget"
      - "--no-verbose"
      - "--tries=1"
      - "--spider"
      - "http://localhost:9000/ping"
  failureThreshold: 1
  initialDelaySeconds: 10
  periodSeconds: 10
  successThreshold: 1
  timeoutSeconds: 2
```

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.1.6
+version: 10.1.7
 appVersion: 2.4.13
 keywords:
   - traefik
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -43,6 +43,11 @@
           {{- toYaml . | nindent 10 }}
           {{- end }}
         readinessProbe:
+          {{- if .Values.ports.traefik.readinessProbe }}
+          {{- with .Values.ports.traefik.readinessProbe }}
+          {{- toYaml . | nindent 10 }}
+          {{- end }}
+          {{- else }}
           httpGet:
             path: /ping
             port: {{ default .Values.ports.traefik.port .Values.ports.traefik.healthchecksPort }}
@@ -51,7 +56,13 @@
           periodSeconds: 10
           successThreshold: 1
           timeoutSeconds: 2
+          {{- end }}
         livenessProbe:
+          {{- if .Values.ports.traefik.livenessProbe }}
+          {{- with .Values.ports.traefik.livenessProbe }}
+          {{- toYaml . | nindent 10 }}
+          {{- end }}
+          {{- else }}
           httpGet:
             path: /ping
             port: {{ default .Values.ports.traefik.port .Values.ports.traefik.healthchecksPort }}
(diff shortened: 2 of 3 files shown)


jakubhajek:
Hello @AndreasVikke 

Thanks a lot for opening the PR. 

I've just read the related issue #479 and I wonder whether changing to exec from built-in httpGet solve your problem; that is related to the high resource consumption on a cluster where Traefik has been deployed. 

Personally, I was using `wget` in a Docker-based environment and had a real use case to replace Kubernetes healthprobes from HTTPGet to `exec` and use `wget` from the Traefik image.

Thanks for the additional information.

AndreasVikke:
@jakubhajek

I have run into a problem where 2-3 pods restarts cause the probes fail. I've noticed that the restarts happens around the same time the nodes are in demand. I run my traefik as a DaemonSet, and it only happens on those 2-3 nodes. 

I am not sure if exec will help with the problem but before i can test it in my setup i need to ability to change the probes in helm chart

faust64:
> but before i can test it in my setup i need to ability to change the probes in helm

Well. If this is really the case, you can publish your patched chart to any HTTP repository you have available, and use this as a source. Eg:

```
[... package from git root ...]
helm package ./traefik
cp ./traefik-x.y.z.tgz /path/to/local/helm-custom-charts/
helm repo index /path/to/local/helm-custom-charts --url https://repository.example.com/helm/charts/
rsync -avWxP /path/to/local/helm-custom-charts/* root@repository.example.com:/var/www/helm/charts/
[... then use your custom repo ...]
helm repo add utgb https://repository.example.com/helm/charts
helm repo update
helm repo list
```

Though in my experience, first you troubleshoot the issue, fix it, and only then you would submit a pull request.
If you're not sure this fixes anything: why should it be merged into upstream?

I kinda agree with jakubhajek here: I doubt running a wget through exec probe would be faster than the native httpGet.

As a starting point: when my ingress controllers are restarted by kubelet, the issue is always related to resource consumption on the corresponding node. specifically CPU. or arguably, liveness/readiness timeouts beeing to low, while overall CPU usage slow things down.
A way to work around this --- assuming this is indeed your problem -- would be to ensure that:
 - make sure your ingress have resource requests set (even better: set resource limits as well)
 - the difference between resource limits and requests allows for a reasonable overcommit (or no overcommit at all), for all workload running on your cluster, likely to be collocated with your ingress controllers
 - there are no pods with no resource limits/requests
 - maybe consider dedicating nodes to your infrastructure services (ingress, monitoring, logging, ...) as it used to be done in OpenShift3. That wasn't the worst idea: would prevent devs working on your cluster from tampering with infrastructure deployments

AndreasVikke:
I will look more into this. Thanks for the help, and sorry for the inconvenience.

PR #483, opened 2021-08-19
Fix Ingress RBAC for namespaced scoped deployment

velothump:
### What does this PR do?

Fixes the Ingress RBAC for namespaced scoped deployments.

### Motivation

Needed to allow Traefik to run successfully when deployed with only namespace scoped RBAC. 

### More


jakubhajek (approved):
Thanks for contributing to the project. The change looks good to me.

Issue #489, opened 2021-08-22
new crossnamespace behavior lead to downtime


HaveFun83:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

10.3.1

### What version of Traefik are you using?

traefik:2.5.0

### What did you do?

Upgrade traefik-helm-chart from 10.1.1 to 10.3.1 break crossnamespaced service in ingressroute crd as the default `providers.kubernetescrd.allowcrossnamespace` changed from true to false.

https://github.com/traefik/traefik/releases/tag/v2.4.11
https://github.com/traefik/traefik/pull/8260

### What did you see instead?

```
ingress/traefik-64d6f66b4b-gmr2m[traefik]: time="2021-08-22T22:09:00Z" level=error msg="service monitoring/infra-grafana not in the parent resource namespace ingress" namespace=ingress providerName=kubernetescrd ingress=grafana-ingress
```
**And all ingressroutes with crossnamspaced services went unreachable.**

### What is your environment & configuration?

k3s v1.21.4

### Additional Information

```shell
The following argument fixed this issue

    additionalArguments:
      - "--providers.kubernetescrd.allowcrossnamespace=true"
```
I used the following default provider config
```
    providers:
      kubernetesCRD:
        enabled: true
        namespaces: []

      kubernetesIngress:
        enabled: true
        namespaces: []
```
```

**Comment by faust64** [comment]
Hi,

That doesn't sound like something that would have changed recently. From 10.1.1 to 10.3.1, diff shows changes to annotations, some missing rbac apiGroups, and ingressClasses fallback API version.

Any chance you someone may have added that option to your deployment/daemonset without updating your values file?

**Comment by hannius** [comment]
HI,HaveFun83 ,Did you have faced the problem about crossnamespace of kubernetesGateway
```
# Enable experimental features
experimental:
  plugins:
    #enabled: false
    enabled: true
  kubernetesGateway:
    enabled: true
    #enabled: true
    appLabelSelector: "traefik"
    #certificates: []
    certificates:
     - group: "core"
       kind: "Secret"
       #name: "mysecret"
       name: "martinm-tls"
```

HaveFun83:
Hi i dont use the kubernetesGateway.
[...]

PR #491, opened 2021-08-29
added providers.kubernetesCRD.allowCrossNamespace

HaveFun83:
### What does this PR do?

This PR allow the configuration of  `providers.kubernetesCRD.allowCrossNamespace` via values.yaml.
Default is this value set to false so no container argument will be set.  As the default behaviour within the traefik container image changed from `true` to `false` since version v2.4.11 [providers/kubernetes-crd.md#allowcrossnamespace]( https://github.com/traefik/traefik/blob/master/docs/content/providers/kubernetes-crd.md#allowcrossnamespace
)
### Motivation

This PR fix the workaround in.
https://github.com/traefik/traefik-helm-chart/issues/489

### More

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -134,6 +134,9 @@
           {{- if .Values.providers.kubernetesCRD.enabled }}
           - "--providers.kubernetescrd"
           {{- end }}
+          {{- if .Values.providers.kubernetesCRD.allowCrossNamespace }}
+          - "--providers.kubernetescrd.allowCrossNamespace=true"
+          {{- end }}
           {{- if .Values.providers.kubernetesIngress.enabled }}
           - "--providers.kubernetesingress"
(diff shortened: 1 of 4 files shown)


traefik/templates/_podtemplate.tpl
```diff
@@ -134,6 +134,9 @@
           {{- if .Values.providers.kubernetesCRD.enabled }}
           - "--providers.kubernetescrd"
           {{- end }}
+          {{- if .Values.providers.kubernetesCRD.allowCrossNamespace }}
```
ddtmachado, line 137: because of the way Traefik flags works setting 
```yaml
providers:
  kubernetesCRD:
    enabled: false
    allowCrossNamespace: true
```
Would actually enable the CRD provider, which is weird

ddtmachado (changes requested):

ddtmachado:
Closing in favor of #496 which fixes the code review requested changes

PR #513, opened 2021-10-13
wip / shouldn't we set ingressclass for kubernetes ingresses as well?!

faust64:
### Motivation

Reading a question on SO, I  realize we may be missing something ... (question about nginx ... I've lost my focus ..)
An option that is documented, in Traefik docs, seems to be missing from our _podTemplate

### More

### Additional Notes

don't merge yet

would there be any rational for using separate IGC, dealing with Ingresses and traefik-specific objects?
maybe move providers.xxx.ingressClass to providers.ingressClass?

then: add tests

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -3,7 +3,7 @@ name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
 version: 10.9.1
-appVersion: 2.5.6
+appVersion: 2.5.7
 keywords:
   - traefik
   - ingress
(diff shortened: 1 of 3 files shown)


kevinpollet:
Hello @faust64,

This pull request seems to be a good candidate to fix https://github.com/traefik/traefik-helm-chart/issues/555.

Regarding your questions:

> would there be any rational for using separate IGC, dealing with Ingresses and traefik-specific objects?
> maybe move providers.xxx.ingressClass to providers.ingressClass?

As the `ingressClass` option is already available for the `kubernetesCRD` provider, adding it only to the `kubernetesIngress` provider sounds good. As you said, we could have only one config option and the same applies to all duplicated options between the providers, feel free to open an issue to discuss that.

Could you update this pull request (rebase + minor version bump)?

kevinpollet:
Superseded by #513

Issue #515, opened 2021-10-21
Support "behavior" in Horizontal Pod Autoscaler (autoscaling/v2beta2 | 1.18+)


Xyaren:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

I would like to request Support for HPA behavior in helm values.

API is in beta since K8s 1.18 and requires using the new autoscaling/v2beta2.
Documentation: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-configurable-scaling-behavior

This would allow much more control on scale in and out operations by the hpa.

PR #518, opened 2021-10-29
Update default HPA API Version to `v2` and add support for behavior

grieshaber:
### What does this PR do?

This MR updates the API-Version of the hpa to autoscaling/v2beta2. Since k8s 1.18, this api-version adds support for custom scaling behavior. Therefore i added some example values and the corresponding property in hpa.yaml.

### Motivation
Requested by @Xyaren in #515

### More

### Additional Notes

I wrapped the behavior property in the hpa in an additional if clause, because it is only supported in clusters >= 1.18, so it won't break things and needs to be enabled explicitly.

### DIFF
--- a/traefik/templates/hpa.yaml
+++ b/traefik/templates/hpa.yaml
@@ -1,5 +1,11 @@
 {{- if .Values.autoscaling.enabled }}
+{{- if .Capabilities.APIVersions.Has "autoscaling/v2beta2/HorizontalPodAutoscaler" }}
 apiVersion: autoscaling/v2beta2
+{{- else if .Capabilities.APIVersions.Has "autoscaling/v2beta1/HorizontalPodAutoscaler" }}
+apiVersion: autoscaling/v2beta1
+{{- else }}
+{{- fail "\n\n ERROR: You must have at least autoscaling/v2beta1 to use HorizontalPodAutoscaler" }}
+{{- end }}
 kind: HorizontalPodAutoscaler
 metadata:
   name: {{ template "traefik.fullname" . }}
--- a/traefik/templates/hpa.yaml
+++ b/traefik/templates/hpa.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
 metadata:
   name: {{ template "traefik.fullname" . }}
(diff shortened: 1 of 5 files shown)


junaid-ali:
@kevinpollet any update on this one, could you please have a look at this again?

cilindrox:
this is also relevant since v2beta2 allows specifying multiple metrics for the HPA to scale on. Right now we're constrained to a single metric.

grieshaber:
I rebased my branch to the latest changes and would appreciate someone having a look at this PR.

traefik/templates/hpa.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2beta2
```
jburke-godaddy, line 2:
```suggestion
{{- if .Capabilities.APIVersions.Has "autoscaling/v2beta2/HorizontalPodAutoscaler" }}
apiVersion: autoscaling/v2beta2
{{- else if .Capabilities.APIVersions.Has "autoscaling/v2beta1/HorizontalPodAutoscaler" }}
apiVersion: autoscaling/v2beta1
{{- else }}
{{- fail "\n\n ERROR: You must have at least autoscaling/v2beta1 to use HorizontalPodAutoscaler" }}
{{- end }}
```
Similar adaption is in: https://github.com/traefik/traefik-helm-chart/blob/859adf66447d8242ae6c697fc901473d4684c282/traefik/templates/ingressclass.yaml#L2-L10

This code should allow the template to adapt to older versions.

jburke-godaddy (changes requested):

mrx88:
FYI - maybe due to helm version problems (did not verify),  but this part is not rendering correctly when using the example configuration with Helm v3.10.1:
```
{{- if .Values.autoscaling.behavior }}
  behaviour: 
{{ toYaml .Values.autoscaling.behavior | indent 4 }}
{{- end }}
```
Output is:
```
 behavior:
    scaleDown:
      policies:
      - periodSeconds: 60
        type: Pods
        value: 1
      stabilizationWindowSeconds: 300
```

but should be:
```
  behavior:
    scaleDown:
      stabilizationWindowSeconds: 300
      policies:
      - type: Pods
        value: 1
        periodSeconds: 60
```
(Note the -type is missing).

mloiseleur:
Hello @mrx88,

There is `type: Pods` in the output. It's expected (and alphabetically ordered) output. 
In this example, `periodSeconds`, `type` and `value` are part of the same policy. It doesn't matter if `periodSeconds` is before or after `type` as long as they are in the same policy.

mrx88:
> Hello @mrx88,
> 
> There is `type: Pods` in the output. It's expected (and alphabetically ordered) output. In this example, `periodSeconds`, `type` and `value` are part of the same policy. It doesn't matter if `periodSeconds` is before or after `type` as long as they are in the same policy.

I was pointing out to "-" char missing for Type, but if it works that way too then no worries ;)

PR #522, opened 2021-11-10
Add lifecycle values property

danielwegener:
### What does this PR do?

This PR adds the capability to add a `deployment.lifecycle` value property to the traefik podTemplate. This allows setting Kubernetes Lifecycle Hooks for traefik containers.

### Motivation

Running traefik as ingress controller behind an AWS pod-ip-based Network Load Balancer (NLB), the update of a TargetGroupBinding is propagated by the AWS Load Balancer Controller relatively quickly to the NLB target-group. 
Terminating a traefik Pod (due to rebalancing or HPA scale-down) however still creates a time frame in which the traefik container in the pod has received SIGTERM and stops serving requests while the NLB Target Group drains connections from pod that is to be terminated.

This causes single requests running into timeouts - especially on longer living connections that are not always terminated instantly (http with keepalive or ssh).

A common solution to get this deployment interruption free is to delay the SIGTERM to traefik via preStop Container Lifecycle Hook (simple sleep command) for a time frame that allows the TargetGroup to gracefully drain and deregister the Pod as target.

### More

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.6.1
+version: 10.7.0
 appVersion: 2.5.3
 keywords:
   - traefik
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -37,6 +37,9 @@
       containers:
       - image: "{{ .Values.image.name }}:{{ default .Chart.AppVersion .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
+        {{- with .Values.deployment.lifecycle }}
+        lifecycle: {{ . | toYaml | nindent 10 }}
+        {{- end }}
         name: {{ template "traefik.fullname" . }}
         resources:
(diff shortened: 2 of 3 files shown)


kevinpollet:
Hello @danielwegener and thanks for your interest in Traefik,

If we understand well, adding the lifecycle `preStop` hook allows making sure that all connections are drained by the NLB before Traefik stops, right?

We are wondering if using the [`entrypoints.lifeCycle`](https://doc.traefik.io/traefik/routing/entrypoints/#lifecycle) config options with a corresponding `terminationGracePeriodSeconds` should be enough?

danielwegener:
@kevinpollet thanks for bringing that up. I have not seen the entrypoints.lifeCycle config option and I agree that your solution is definitely clearer.
For reference, adding:
```
traefik:
  additionalArguments:
     - '--entryPoints.websecure.transport.lifeCycle.graceTimeOut=60'
```
should do it.

kevinpollet:
Good to know!

We will decline this pull request accordingly.

Thanks again for your contribution!

FredrikAugust:
Hi! Is there any reason the `requestAcceptGraceTimeout` isn't included in the Helm chart under `ports.websecure.lifeCycle`? Nor any of the other lifecycle parameters for that matter.

PR #525, opened 2021-11-15
Add support to specify minReadySeconds on Deployment/DaemonSet

mikesir87:
### What does this PR do?

This PR adds the ability to specify the `.spec.minReadySeconds` for both Deployments and DaemonSets.

### Motivation

When deploying to EKS clusters with Service LoadBalancers, we often see a little blip of downtime when rolling out the updates due how fast the update occurs. Since the load balancers have to run their own healthchecks and registered listeners, if pods are scheduled onto other machine instances, it might take a few seconds to update. By having this ability, we can control the rollout a little bit better.

### More

- [ ] Yes, I updated the [chart version](https://github.com/traefik/traefik-helm-chart/blob/9b32ed1b414cc0be1ad46bcb335fcfc93ded06f3/traefik/Chart.yaml#L5)

Not yet, but happy to do so.


kevinpollet (approved):
Thanks 👍

Issue #258, opened 2020-09-08
Pod disruption budget should allow percentages


hintofbasil:
A pod disruption budget should allow percentages for the `.spec.minAvailable` and `.spec.maxUnavailable` fields.
https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget

The chart casts them to integers which disallows this.
https://github.com/containous/traefik-helm-chart/blob/289d2b2/traefik/templates/poddisruptionbudget.yaml#L17

LudovicTOURMAN:
We used to face a downtime of our production because of this conversion which lead to change `50%` to `0`.
Thanks for opening this issue guys, I created a corresponding PR to fix it.

mloiseleur:
Fixed with #528

PR #528, opened 2021-11-25
Fix pod disruption when using percentages

LudovicTOURMAN:
### What does this PR do?

It fixes the issue where setting `podDisruptionBudget.minAvailable` or `podDisruptionBudget.maxUnavailable` as percentage will be convert to `0` value

### Motivation

Having an issue in production while setting a `podDisruptionBudget.minAvailable` to `50%` shows us that it was converted to value `0`.
Which leads to no disruption budget and a downtime on traefik service.

### More

- [X] Yes, I updated the [chart version](https://github.com/LudovicTOURMAN/traefik-helm-chart/blob/8297d33e0564ebff58f10953ebaa3efa4db97a7f/traefik/Chart.yaml#L5)

### Additional Notes

This fix relates to the current opened issue: https://github.com/traefik/traefik-helm-chart/issues/258
It follows Kubernetes specification: https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget

> .spec.minAvailable which is a description of the number of pods from that set that must still be available after the eviction,
> even in the absence of the evicted pod. minAvailable can be either an absolute number or a percentage.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.7.0
+version: 10.7.1
 appVersion: 2.5.4
 keywords:
   - traefik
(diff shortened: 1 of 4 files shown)


LudovicTOURMAN:
Hi @kevinpollet,
Could you tell me what should I do in order this bug fix to be reviewed and released?

kevinpollet (approved):
Thanks 👍

PR #337, opened 2021-01-14
fix(misc): hostNetwork deployment when hostPort mismatches containerPort 

faust64:
- fixes hostNetwork configuration, when containerPort != hostPort
- fixes PodSecurityPolicy, when running privileged
- fixes rbac, when rbac.namespaced=true, access to ingressclass is defined
- adds `--providers.kubernetescrd.allowcrossnamespace` option to podTemplate, defaults to false, can be changed setting `providers.kubernetesCRD.allowCrossNamespace=true`. A default to false might be preferable, as Kubernetes Ingresses doesn't usually allow referring to resources in other namespaces. When you implement RBAC, maybe even NetworkPolicies: that does not sound like an option you would enable. Even though for sure, it could be useful in some chosen scenario.

Fixes #336

PR #530, opened 2021-12-05
fix(ingressclass): somewhat duplicates #377 / fixes #499

### LINKED ISSUE #499: Ingress not working when rbac.namespaced=true
### Welcome!

10.3.4

2.5.1

### What did you do?

We deployed latest helm chart and traefik version on kubernetes 1.22 with `rbac.namespaced: true` 

```
helm install traefik traefik/traefik --set rbac.namespaced=true
```

### What did you see instead?

We got an error in traefik log:

```
E0921 14:07:54.739657       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.21.0/tools/cache/reflector.go:167: Failed to watch *v1.IngressClass: failed to list *v1.IngressClass: ingressclasses.networking.k8s │
│ .io is forbidden: User "system:serviceaccount:default:traefik" cannot list resource "ingressclasses" in API group "networking.k8s.io" at the cluster scope  
```

Routing does not work.

We tested on KinD with kubernetes 1.22.1

### Additional Information

```shell
We also tested with traefik 2.5.3 with same result.
```

faust64:
### What does this PR do?

Fixes #499 

### Motivation

https://github.com/traefik/traefik-helm-chart/pull/337#issuecomment-942111555

### More

### Additional Notes

should test

### DIFF
--- a/traefik/templates/rbac/podsecuritypolicy.yaml
+++ b/traefik/templates/rbac/podsecuritypolicy.yaml
@@ -13,13 +13,17 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/instance: {{ .Release.Name }}
 spec:
-  privileged: false
-  allowPrivilegeEscalation: false
   requiredDropCapabilities:
     - ALL
 {{- if not .Values.securityContext.runAsNonRoot }}
   allowedCapabilities:
     - NET_BIND_SERVICE
+  allowPrivilegeEscalation: true
+  privileged: true
+{{- else }}
+  allowedCapabilities: []
+  allowPrivilegeEscalation: false
+  privileged: false
 {{- end }}
   hostNetwork: {{ .Values.hostNetwork }}
   hostIPC: false
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.6.2
+version: 10.6.3
 appVersion: 2.5.4
 keywords:
   - traefik
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.9.1
+version: 10.9.2
 appVersion: 2.5.6
 keywords:
   - traefik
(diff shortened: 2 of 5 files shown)


traefik/templates/rbac/podsecuritypolicy.yaml
```diff
@@ -13,13 +13,17 @@ metadata:
...
 spec:
-  privileged: false
-  allowPrivilegeEscalation: false
   requiredDropCapabilities:
     - ALL
 {{- if not .Values.securityContext.runAsNonRoot }}
   allowedCapabilities:
     - NET_BIND_SERVICE
+  allowPrivilegeEscalation: true
+  privileged: true
```
ddtmachado, line 22: why would you run as non-root and yet set it as privileged and allow privilege escalation?

ddtmachado (commented):
Thanks @faust64 , just one question on the proposed changes

traefik/templates/rbac/podsecuritypolicy.yaml
faust64: As a general rule: I would do this for my unprivileged process being able to bind on privileged ports.
Though I'm a little rusty on the specifics of this PR. I recall the traefik container image not setting the EP setcap on traefik binary.... Does this trick still applies, then?
I know I ended up using 8080/8443 as container ports, while exposing 80/443 as hostPorts ... my allowPrivilegeEscalation is false ...

faust64: ... And now I get it. "if not .Values.xxx.runAsNonRoot".
runAsNonRoot not set, so we run as root. So we allow for this in PSP.

Though I wouldn't recommend it, personally.
The traefik container image runs as root, by default

```
$ docker run --entrypoint /bin/sh -it docker.io/traefik
/ # id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
```

Which isn't really a good practice .. but "fixing" the image would most definitely break lots of existing installs ... I would avoid changing that bit now, to avoid the "I can't read/renew/... my certificates" cases

Thus, by default, with PSP enabled: traefik pods won't start. Either set runAsNonRoot in values, or allow for privileged containers in PSP.

faust64: now .... in retrospective ... that should have stayed in #337 ...

mloiseleur:
This PR is not applicable in this current state.
@faust64 If you have the time, you can refresh it, take @ddtmachado  comments into account and re-open it.

faust64:
Really? I've been refreshing that PR, and the parent one, for over a year.
If you want something from it, go ahead. I'm done here.

mloiseleur:
@faust64 Sorry, I didn't mean to offend you. 
I see that you have opened multiple PRs in the past and I understand.
=> I'll prioritize your PRs and try to merge them.

PR #531, opened 2021-12-07
Create CRDs only if kubernetesCRD provider is enabled

olljanat:
### What does this PR do?
Afaiu CRDs can be only used when `kubernetesCRD` provider is enabled so this PR will add check which skips creating those in case it is not which is nice for those of use who prefer to use `kubernetesIngress` instead.

### Motivation
I'm running multi-tenant Kubernetes setup on way that there is multiple [vcluster](https://www.vcluster.com/)s running on top of prefer to keep those without CRDs so they are easier to understand by users.

### More

### Additional Notes
Currently I have done only very light testing for this one but I assume that there is nice CI here which should tell about if this change breaks something or not.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.7.1
+version: 10.7.2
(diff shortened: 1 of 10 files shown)


olljanat:
Closing as `helm install --skip-crds` can be used instead of this.

PR #534, opened 2021-12-22
Support liveness and readyness probes customization

osipovdaniil:
### What does this PR do?

`Liveness` and `readiness` probes cannot be changes. So this PR will add ability customize params of probes and we are able to  disable probes.

### Motivation

I have specific case, i need to disable `liveness` probe and do more aggressive readiness probe. Now, i can't customize these probes.

### More

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.17.0
+version: 10.18.0
 appVersion: 2.6.2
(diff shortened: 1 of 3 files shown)


kevinpollet:
Hello @osipovdaniil and thanks for your interest in Traefik,

Allowing the configuration of the probes seems a legit use case, but why should we allow disabling the probes? Could you give more details about your use case?

osipovdaniil:
@kevinpollet Hi! I think we can do without the ability to disable probes.
I updated my PR. Please check again

kevinpollet (approved):
Thanks 👍

Issue #537, opened 2022-01-04
Allow to specify http3 entrypoints


tete17:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

With the addition of http3 in traefik since 2.5.0 it would be nice to be able to specify this feature as a first class option, instead of relying on hacky solutions such as:

>=2.5.0 <2.6.0
```yaml
additionalArguments:
 - "--experimental.http3=true"
 - "--entrypoints.websecure.enabledHTTP3=true"
```

>=2.6.0
```yaml
additionalArguments:
 - "--experimental.http3=true"
 - "--entrypoints.websecure.http3"
 - "--entrypoints.websecure.http3.advertisedPort=443"
```

PR #488, opened 2021-08-22
Add experimental HTTP/3 support

timoschwarzer:
### What does this PR do?

This PR adds support for experimental HTTP/3 features in Traefik.

### Motivation

HTTP/3 :tada:

Fixes #537

### More

- [X] Yes, I updated the [chart version](https://github.com/traefik/traefik-helm-chart/blob/9b32ed1b414cc0be1ad46bcb335fcfc93ded06f3/traefik/Chart.yaml#L5)

### Additional Notes

I also updated `appVersion` to 2.5.1 because 2.5.0 contained a bug where all `POST` requests failed with HTTP/3.

Co-authored-by: tete17 <miguel_tete17@hotmail.com>

PR #538, opened 2022-01-04
Add http3 support

tete17:
### What does this PR do?
https://github.com/traefik/traefik-helm-chart/issues/537

Add http3 support for the helm chart

### Motivation

I want to enable http3 on my small k3s cluster because it sounds cool and I am a bit frustrated this can't be done in a "preferential kind of way". In the linked issue I discuss how to do it before this patch but it is very hacky in my opinion and ver fragile since it depends on you typing the entrypoint names correctly. This should improve things

### More

### Additional Notes

I took care of the config changes between 2.5 and 2.6 version so everything should be working when we decide to upgrade the app version. It is also tested but I am open to add more if is is necessary.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.9.1
+version: 10.9.2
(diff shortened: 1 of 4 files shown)


tete17:
Closed in favour of https://github.com/traefik/traefik-helm-chart/pull/488

PR #540, opened 2022-01-05
tweak default values to avoid downtime when updating

ddtmachado:
### What does this PR do?

Tweak the default values to avoid downtime while updating or restarting the deployment.

### Motivation

Currently if an user just installs using the default chart values they'll get `maxUnavailable: 1` and `initialDelaySeconds: 10` while running just one single replica, which means any restart or update will cause a big downtime since Kube will destroy the running pod before starting a new one that then will only be considered ready after at least 10 seconds.

### More


traefik/values.yaml
```diff
@@ -11,7 +11,7 @@ image:
 deployment:
   enabled: true
   # Can be either Deployment or DaemonSet
-  kind: Deployment
+  kind: DaemonSet
```
mmatur, line 14:
```suggestion
  kind: Deployment
```

traefik/values.yaml
```diff
@@ -104,10 +104,10 @@ ingressRoute:
     labels: {}
 
 rollingUpdate:
-  maxUnavailable: 1
+  maxUnavailable: 0
```
mmatur, line 107:
```suggestion
  maxUnavailable: 1
```

mmatur (changes requested):

mloiseleur:
@ddtmachado and @mmatur  : I updated the PR. Wdyt ?

mmatur (approved):

charlie-haley (approved):
LGTM

PR #545, opened 2022-01-15
Add support for ipFamilies

peterfromthehill:
Signed-off-by: peterfromthehill <peter@1qay.net>

### What does this PR do?

Allows setting the service's .spec.ipFamilies as documented in https://kubernetes.io/docs/concepts/services-networking/dual-stack/

### Motivation

#453 is not enough to provide in an dualstack cluster a IPv6 only LoadBalancer object.

### More


kevinpollet (approved):
Thanks 👍

PR #548, opened 2022-01-25
Add setting to enable Instana tracing

cedricziel:
### What does this PR do?

This PR introduces a helm chart setting to enable Instana tracing without the need of manually supplying arguments.

### Motivation

We want to make the Instana integration easier to use :)

### More

### Additional Notes

Enabling the Instana Tracing and Sensor integration is very straight forward now. Just use `--set tracing.instana=true` on the helm deployment and be done. The tracer will now seamlessly report to the Instana Agent.


kevinpollet (approved):
Thanks 👍

PR #551, opened 2022-01-28
Allow setting revisionHistoryLimit for Deployment and DaemonSet

Benoit382:
### What does this PR do?

This PR introduces a helm chart setting to enable the revision history limit.

### Motivation

I want limit the number of replica sets.

### More

### Additional Notes

Co-authored-by: Julien Aubert <julien.aubert@malt.com>


kevinpollet (approved):
Thanks 👍

PR #553, opened 2022-01-31
Prefix cluster resources to allow multiple cluster level installs

mueschm:
### What does this PR do?
Adds a new value under rbac called prefix. This allows multiple versions of traefik to be installed on each cluster without conflicts.

### Motivation
As part of our deployment, we tend to need multiple instances of traefik running in our clusters. We could set ```rbac.namespaced=true``` but then we are limited to having 1 instance of traefik per namespace. What we are actually looking for is to allow N number of traefik instances to work with M namespaces. 
Essentially this adds a third option to the configuration.
1) 1 traefik instance per cluster
2) 1 traefik instance per namespace
3) (New) N traefik instances per cluster

### More

- [x ] Yes, I updated the [chart version](https://github.com/traefik/traefik-helm-chart/blob/9b32ed1b414cc0be1ad46bcb335fcfc93ded06f3/traefik/Chart.yaml#L5)

### Additional Notes

Right now the main issue with deploying multiple instances of traefik with this configuration is because of the cluster role and cluster role binding templates. So we added ```rbac.prefix``` as a value. Though would like some feedback on if this should actually be a root value since the goal is to allow any cluster level resources to be prefixed.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.12.0
+version: 10.12.1
 appVersion: 2.6.0
 keywords:
   - traefik
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -37,7 +37,7 @@ Create a default fully qualified cluster name. This allows multiple traefik inst
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "traefik.rbac.clusterFullname" -}}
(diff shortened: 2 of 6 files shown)


mueschm:
Hello, was wondering what a would a timeframe look like for getting this PR approved? Happy to make any additional changes and upgrade the Chart.yml file version again once I know the changes are good enough.

mloiseleur:
Hello @mueschm ,

I was able to install two different Traefik on the same namespace with current version (v12) of this chart. 
With this values files:
```yaml
rbac:
  enabled: true
  namespaced: true
```

One can install the chart twice, with different name, on the same namespace: 
```bash
kubectl create namespace traefik
helm install -f rbac.yaml --namespace=traefik traefik1 traefik/traefik 
helm install -f rbac.yaml --namespace=traefik traefik2 traefik/traefik 
```

@mueschm Anything I missed ? Does this works for you ?

PR #558, opened 2022-02-09
Add missing inFlightConn TCP middleware CRD

tomMoulard:
### What does this PR do?

This PR updates the `middlewaretcp` CRD to have access to missing `inFlightConn` middleware.

### Motivation

Have the possibility to use more TCP middleware

### More

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.14.0
+version: 10.14.1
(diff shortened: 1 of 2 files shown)


kevinpollet (approved):
Thanks 👍

z0rc:
Can changes to CRDs be properly announced? Majority of users aren't aware of this change, because it's patch release. And helm doesn't upgrade CDRs by design, so upgrade should be applied manually via kubectl. See how kube-prometheus-stack handles CRDs upgrades https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#upgrading-chart

PR #563, opened 2022-02-21
Add ingressclasses to traefik role

nicolaelapteanu:
### What does this PR do?
Add ingressclasses to traefik role

### Motivation
This PR solves the issue when Traefik is being deployed with `.Values.rbac.namespaced`
```
User system:serviceaccount:sc-services:traefik cannot list resource ingressclasses in API group networking.k8s.io at the cluster scope
```

### More

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.25.0
+version: 10.25.1
 appVersion: 2.8.7
 keywords:
   - traefik
--- a/traefik/templates/rbac/role.yaml
+++ b/traefik/templates/rbac/role.yaml
@@ -24,6 +24,7 @@ rules:
       - networking.k8s.io
     resources:
       - ingresses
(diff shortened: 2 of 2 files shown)


nicolaelapteanu:
this is a fix for https://github.com/traefik/traefik/issues/7729#issuecomment-790019616

raresociopath (approved):

bmv126:
Does this really fix the issue ? 
As per discussion https://github.com/traefik/traefik/issues/7729#issuecomment-790717134 the solution does not work.
Also as per https://github.com/kubernetes/kubernetes/issues/99824 it is not supported

SPodjasek:
As @bmv126 described this won't work. 
I've run into similar issue and for me solution was to set `rbac.namespaced` to `false`, so that created `ClusterRole` will allow listing of `ingressclass` objects. And to limit `kubernetes*` providers to specified namespace afterwards. As for now, the second step has to be done by hand, as I've described in this issue traefik/traefik-helm-chart#615.

mloiseleur (approved):
LGTM. Current rbac on traefik recommends this resources.
I'll update the version and merge this PR. Thanks !

PR #567, opened 2022-02-24
Add readinessGates and lifecycle properties

romainDavaze:
### What does this PR do?

Add the `readinessGates` and `lifecycle` properties to the pod template

### Motivation

In order to use the AWS LB Controller in conjunction with Traefik with minimal downtime, we need to setup a readinessGate as well as a preStop command (cf. [this article](https://easoncao.com/zero-downtime-deployment-when-using-alb-ingress-controller-on-amazon-eks-and-prevent-502-error/))

```
        lifecycle:
          preStop:
            exec:
              command:
              - /bin/sh
              - -c
              - sleep 40
      readinessGates:
      - conditionType: target-health.alb.ingress.k8s.aws/<ingress_name>_<service_name>_<service_port>
```

### More

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -57,10 +57,21 @@ deployment:
   # Additional imagePullSecrets
   imagePullSecrets: []
     # - name: myRegistryKeySecretName
-  # Use readiness gates
+  # Use readiness gates (e.g to specify a list of additional conditions for pod readiness)
   readinessGates: []
+    # Example with AWS LB Controller to set pod healthiness based on its registration state on a given Application Load Balancer
+    # - conditionType: target-health.alb.ingress.k8s.aws/traefik-ingress_traefik-service_9000
   # Pod lifecycle actions
   lifecycle: {}
+    # preStop:
+    #   exec:
+    #     command: ["/bin/sh", "-c", "sleep 40"]
+    # postStart:
+    #   httpGet:
+    #     path: /ping
+    #     port: 9000
+    #     host: localhost
+    #     scheme: HTTP
 
 # Pod disruption budget
 podDisruptionBudget:
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -57,6 +57,10 @@ deployment:
   # Additional imagePullSecrets
   imagePullSecrets: []
     # - name: myRegistryKeySecretName
+  # Use readiness gates
+  readinessGates: []
+  # Pod lifecycle actions
+  lifecycle: {}
 
 # Pod disruption budget
 podDisruptionBudget:
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.14.1
-appVersion: 2.6.0
+version: 10.14.2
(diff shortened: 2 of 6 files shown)


jakubhajek (changes requested):
Could you please also add tests for those new enhancements?

rdavaze-talend:
Hello @jakubhajek ,

I have added some tests as well as some examples for both `readinessGates` and `lifecycle` fields.
Is it ok for you ?

I'm available if you have any question regarding this.

romainDavaze:
Hello @jakubhajek ,

I have added some tests as well as some examples for both `readinessGates` and `lifecycle` fields.
Is it ok for you ?

I'm available if you have any question regarding this.

jakubhajek (commented):
Thank you for adding tests and more informative examples to the `values.yaml`.  The examples are always welcome since they are helping users with Traefik deployment.

jakubhajek (approved):
LGTM. Thank you a lot for your contribution.

z0rc:
@romainDavaze FYI you're acting based on incorrect recommendations. Linked article uses legacy style of readiness gates, that were used by AWS LB Controller v1. For up-to-date documentation see https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/deploy/pod_readiness_gate/. There is no need to configure readiness gates in deployment, AWS LB controller injects them on the fly.

In general, there is no need to specify `readinessGates` and `lifecycle` properties for traefik, when using AWS LB Controller v2.

traefik/values.yaml
```diff
@@ -57,6 +57,21 @@ deployment:
   # Additional imagePullSecrets
   imagePullSecrets: []
     # - name: myRegistryKeySecretName
+  # Use readiness gates (e.g to specify a list of additional conditions for pod readiness)
+  readinessGates: []
+    # Example with AWS LB Controller to set pod healthiness based on its registration state on a given Application Load Balancer
+    # - conditionType: target-health.alb.ingress.k8s.aws/traefik-ingress_traefik-service_9000
```
z0rc, line 63: This is legacy configuration and shouldn't be recommended as example

jakubhajek: Thank you @z0rc for keeping an eye on that. Do you know what is the correct example for the readinessgate? If so, can you please share it with us? Thank you,

z0rc: I don't have a good example. AWS LB Controller injects readinessgate on the flight, gate names are also generated and don't make much sense to user. I'm not aware of other examples where and how user-defined readinessgates can/should be used.

miadabrin: You are right about the readiness gate (as it can be done by adding a label to the namespace) but we still need the lifecycle properties to let the pod be in terminated state long enough so that the ALB can drain it

miadabrin: basically the idea is that the pod can still serve traffic even if it is technically terminated. this helps to remove it from target group without causing outage

miadabrin:
```yml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    elbv2.k8s.aws/pod-readiness-gate-inject: enabled
  name: traefik
```

miadabrin: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1064 reference for the need of preStop hook

PR #569, opened 2022-03-07
Add metrics.prometheus.addRoutersLabels option

nmengin:
### What does this PR do?

This PR adds the option `metrics.prometheus.addrouterslabels`to the values file.
Then the routers metrics can be set using the values.

### Motivation

Set easily the routers metrics when we install Traefik using the Helm Chart.

### More

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -128,6 +128,9 @@
           {{- if .Values.metrics.prometheus }}
           - "--metrics.prometheus=true"
           - "--metrics.prometheus.entrypoint={{ .Values.metrics.prometheus.entryPoint }}"
+          {{- if .Values.metrics.prometheus.addrouterslabels }}
+          - "--metrics.prometheus.addrouterslabels={{ .Values.metrics.prometheus.addrouterslabels }}"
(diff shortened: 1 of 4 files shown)


traefik/templates/_podtemplate.tpl
```diff
@@ -128,6 +128,9 @@
           {{- if .Values.metrics.prometheus }}
           - "--metrics.prometheus=true"
           - "--metrics.prometheus.entrypoint={{ .Values.metrics.prometheus.entryPoint }}"
+          {{- if .Values.metrics.prometheus.addrouterslabels }}
+          - "--metrics.prometheus.addrouterslabels={{ .Values.metrics.prometheus.addrouterslabels }}"
+          {{- end }}
```
tomMoulard, line 133:
```suggestion
          {{- if semverCompare ">=2.5.0" (default $.Chart.AppVersion $.Values.image.tag)}}
          {{- if .Values.metrics.prometheus.addrouterslabels }}
          - "--metrics.prometheus.addrouterslabels=true"
          {{- end }}
          {{- end }}
```

kevinpollet (approved):
Thanks 👍

PR #570, opened 2022-03-09
CRD: remove 'status:' key from generated configs

jeremyvisser:
The CRDs in this Helm chart all follow the form:

    kind: CustomResourceDefinition
    metadata:
      # ...
    spec:
      # ...
    status:
      acceptedNames:
        kind: ""
        plural: ""
      conditions: []
      storedVersions: []

The problem with the `status:` key is that it is usually
runtime-populated, for example by a controller.

In the case of a cluster managed by Anthos Config Management, it fails
to apply the Helm-generated Traefik configs with the error:

    KNV1045: Configs with "status" specified are not allowed.
    To fix, either remove the config or remove the "status" field in the config.
    For more information, see https://g.co/cloud/acm-errors#knv1045

If you remove the `status:` key from the CRDs, this now applies cleanly to an
Anthos-managed cluster.

### More

### Additional Notes

This does not really affect using `helm install`, but it does affect a situation where you `helm template` into a config-managed directory which is then pushed to the cluster using a separate mechanism (e.g. Anthos).

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.15.0
+version: 10.15.1
 appVersion: 2.6.1
 keywords:
   - traefik
--- a/traefik/crds/ingressroute.yaml
+++ b/traefik/crds/ingressroute.yaml
@@ -188,9 +188,3 @@ spec:
         type: object
     served: true
     storage: true
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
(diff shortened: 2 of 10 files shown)


kevinpollet:
Hello @jeremyvisser,

Thanks for your contribution,

The CRDs are generated by Traefik, which means that if something is wrong with this generation an [issue](https://github.com/traefik/traefik/issues) should be opened in the Traefik project, to avoid diverging CRD definitions.

After digging a bit more on the subject, it seems that the issue is a bug of the controller-gen project (used to generate the CRDs), see https://github.com/kubernetes-sigs/controller-tools/issues/456

According to this, we will decline this pull request.

PR #573, opened 2022-03-15
Support allowEmptyServices for KubernetesIngress provider

fty4:
### What does this PR do?

Add option [`allowEmptyServices`](https://doc.traefik.io/traefik/providers/kubernetes-ingress/#allowemptyservices) to allow creating empty services within Kubernetes + Traefik.

The configuration was comment out in `values.yaml` as non-critical feature.

### Motivation

I have a usecase in my cluster to create a service which is mostly empty.
To enable traefik to create a router I need to enable this feature.

### More


kevinpollet (approved):
Thanks 👍

PR #584, opened 2022-04-05
Allow updateStrategy to be configurable

Kyos:
### What does this PR do?

This PR allow the customisation of the updateStrategy in the value file

### Motivation

I use traefik as an Load balancer  configured with hostPort behind multiple ALB, when I update traefik I would like to avoid pod to be directly updated (because this would create a downtime, if I don't drain the nodes on the target group)

### More

### Additional Notes

I used a global updateStrategy variable and removed the existing rollingUpdate key. This creates a breaking change (that's why I updated the chart version to a new major), I can easily avoid this by playing with a "default" or a "with" (but this would add some dirt in the value file), let me know what you prefer :)

Thanks by advance


chendo:
We're also looking at using Traefik for our primary load balancer and we would also require the option to manually recreate the pods so we can take nodes out of active rotation to perform these changes

charlie-haley:
Hi @Kyos, thanks for the PR. The changes look good to me! Could you also add some tests to the PR?

Kyos:
@charlie-haley Just did (I had to adapt a few other test to match the new updateStrategy flag). Let me know if everything looks good to you :)

charlie-haley (approved):

PR #586, opened 2022-04-14
feat: add create gateway option

kahirokunn:
### Motivation

There are situations where the Gateway that is created is not always necessary.
If it is managed by a CD tool, etc., it cannot be deleted, so it is optioned.

### More

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -89,7 +89,8 @@ experimental:
     enabled: false
   kubernetesGateway:
     enabled: false
-    create: true
+    gateway:
+      enabled: true
     # certificate:
     #   group: "core"
     #   kind: "Secret"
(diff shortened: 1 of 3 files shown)


kahirokunn:
Hello👀

traefik/values.yaml
```diff
@@ -89,6 +89,7 @@ experimental:
     enabled: false
   kubernetesGateway:
     enabled: false
+    create: true
```
charlie-haley, line 92:
```suggestion
    gateway:
      enabled: true
```

charlie-haley:
Hi @kahirokunn, thanks for the PR, I've left a couple suggestions on the property in the `values.yaml`, if you could take a look, thanks!

kahirokunn:
@charlie-haley Thx! LGTM!

kahirokunn:
Done!

charlie-haley (approved):

PR #593, opened 2022-05-07
Support environment variable usage for Datadog

lukerogers:
### What does this PR do?
Updates the chart to only specify the Datadog metrics address when it's explicitly supplied. As part of https://github.com/traefik/traefik/pull/7721 it should allow using ENV vars to hit the host directly. The ENV vars are required for  accessing the host, other wise the metrics requests fail

### Motivation

I get around this now by using command line parameters directly. It would be nice to use the YAML definition.

### More


kevinpollet (approved):
Thanks 👍

PR #619, opened 2022-07-29
Suggest an alternative port for metrics

yann-soubeyrand:
When using hostNetwork and metrics default port (9100), there can be a conflict with node exporter which is also listening on this port. Therefore, an alternative port has been reserved (https://github.com/prometheus/prometheus/wiki/Default-port-allocations). This PR adds a comment suggesting to use this port in the values file.


mloiseleur (approved):
Thanks. LGTM.
I updated the comment to put it over the port.

PR #627, opened 2022-09-02
feat: add lifecycle for prestop and poststart

vmarlier:
### What does this PR do?

Add the lifecycle property to the pod template.

### Motivation

In order to allow minimal downtime, we need to setup a preStop command on our Traefik deployments.

```yaml
        lifecycle:
          preStop:
            exec:
              command:
              - /bin/sh
              - -c
              - sleep 40
```

### More

### Additional Notes

This PR is based on this [one](https://github.com/traefik/traefik-helm-chart/pull/567), but we only need the lifecycle configuration.

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 10.27.0
+version: 10.28.0
 appVersion: 2.8.7
 keywords:
   - traefik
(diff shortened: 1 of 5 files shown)


traefik/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
 version: 10.24.2
```
charlie-haley, line 5: The `version` will need incrementing here, `appVersion` shouldn't be touched as it's related to the Traefik version and not the Helm chart version
```suggestion
version: 10.25.0
```

charlie-haley (changes requested):

charlie-haley (approved):
LGTM

PR #662, opened 2022-10-12
:bug: Fix `ClusterRole`, `ClusterRoleBinding` names and `app.kubernetes.io/instance` label

LorianColtof:
### What does this PR do?

1. It adds the name of the namespace to `ClusterRole` and `ClusterRoleBinding` names.
2. It appends the namespace to `app.kubernetes.io/instance` label.

### Motivation

This is to prevent `ClusterRole` and `ClusterRoleBinding` resources with the same names when there are multiple releases with the same name but in different namespaces.

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -32,31 +32,12 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 {{- end -}}
 
-{{/* Generate common labels */}}
-{{- define "traefik.commonLabels" -}}
+{{/* Generate basic labels */}}
+{{- define "traefik.labels" -}}
 app.kubernetes.io/name: {{ template "traefik.name" . }}
 helm.sh/chart: {{ template "traefik.chart" . }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
-{{- end }}
-
-{{/* Generate basic labels */}}
-{{- define "traefik.labels" -}}
-{{ template "traefik.commonLabels" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
-{{- end }}
-
-{{/*
-Generate basic labels for ClusterRole and ClusterRoleBinding.
-Adds the namespace to app.kubernetes.io/instance when it is
-a namespaced release.
-*/}}
-{{- define "traefik.clusterRoleLabels" -}}
-{{ template "traefik.commonLabels" . }}
-{{ if .Values.rbac.namespaced -}}
 app.kubernetes.io/instance: {{ .Release.Name }}-{{ .Release.Namespace }}
-{{- else -}}
-app.kubernetes.io/instance: {{ .Release.Name }}
-{{- end -}}
 {{- end }}
 
 {{/*
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -53,11 +53,7 @@ Adds the namespace to name to prevent duplicate resource names when there
 are multiple namespaced releases with the same release name.
 */}}
 {{- define "traefik.clusterRoleName" -}}
-{{- if .Values.rbac.namespaced -}}
 {{- template "traefik.fullname" . -}}-{{ .Release.Namespace }}
-{{- else -}}
-{{- template "traefik.fullname" . -}}
-{{- end -}}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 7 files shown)


mloiseleur:
Hello @LorianColtof 

Thanks for your PR.  It's interesting and it's also adding complexity for an edge case. 

What do you think about 
1. Adding `app.kubernetes.io/instance: {{ .Release.Name }}-{{ .Release.Namespace }}` for all objects in the existing `labels` helper ? 
2. Set name of ClusterRole and ClusterRoleBinding to `{{- template "traefik.name" . -}}-{{ .Release.Namespace }}` instead of creating a specific helper ?

LorianColtof:
Hi @mloiseleur,

I agree with 1, wasn't entirely sure what to do with the `app.kubernetes.io/instance` label. For 2, this theoretically gives a problem when there are multiple chart installations in the same namespace. What do you think about `{{- template "traefik.fullname" . -}}-{{ .Release.Namespace }}`?

mloiseleur:
@LorianColtof 
> What do you think about {{- template "traefik.fullname" . -}}-{{ .Release.Namespace }}?

Yes, it makes sense, since it's current name is based on fullname.

traefik/templates/_helpers.tpl
```diff
@@ -47,6 +47,19 @@ The name of the service account to use
...
+Adds the namespace to name to prevent duplicate resource names when there
+are multiple namespaced releases with the same release name.
+*/}}
+{{- define "traefik.clusterRoleName" -}}
+{{- if .Values.rbac.namespaced -}}
+{{- template "traefik.fullname" . -}}-{{ .Release.Namespace }}
+{{- else -}}
+{{- template "traefik.fullname" . -}}
+{{- end -}}
+{{- end -}}
```
mloiseleur, line 61:
```suggestion
{{- define "traefik.clusterRoleName" -}}
{{- template "traefik.fullname" . -}}-{{ .Release.Namespace }}
{{- end -}}
```

It's quite unexpected to have a different name when you enable a feature. 
And this issue can also happens without `rbac.namespaced`.

mloiseleur (changes requested):

mloiseleur:
Thanks @LorianColtof 
I left a last suggestion. After that, we should probably be able to merge it.

LorianColtof:
I applied your suggestion @mloiseleur.

mloiseleur (approved):

charlie-haley (approved):

darkweaver87 (approved):
LGTM

traefik/templates/_helpers.tpl
```diff
@@ -37,7 +37,7 @@ If release name contains chart name it will be used as a full name.
 app.kubernetes.io/name: {{ template "traefik.name" . }}
 helm.sh/chart: {{ template "traefik.chart" . }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
-app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/instance: {{ .Release.Name }}-{{ .Release.Namespace }}
```
Project0, line 40: Could this potentially break something? I also do not see why it needs a suffix as the (helm) release name should be already unique per namespace.

LorianColtof: According to the Kubernetes documentation this label should contain "A unique name identifying the instance of an application". So for the global objects `ClusterRole` and `ClusterRoleBinding` this would not be unique without the suffix if there were multiple releases with the same name but in different namespaces. I do agree though that the suffix is not necessary for the namespaced objects, which is why I initially used a separate label helper for `ClusterRole` and `ClusterRoleBinding`, but maybe it is nice to make this label consistent throughout all objects like @mloiseleur suggested.

mloiseleur: @LorianColtof Discussing this change on `app.kubernetes.io/instance` label, we see potentially a breaking change here and we are not sure of the use case. Can you detail us more why you also need this change ? or do you think we are good to merge this PR without this label change ?

LorianColtof: I don't need this label change in particular. I added it just to keep this label unique for the `ClusterRole`/`ClusterRoleBinding` cases.

mloiseleur: @LorianColtof Thanks. It's more clear. We bumped to a Major version, since it can breakj something for people using this label.

PR #464, opened 2021-08-01
rbac: aggregate to admin

faust64:
### What does this PR do?

Aggregates Traefik ClusterRole to Kubernetes admin role.
Users with admin privileges should be able to operate Traefik resources within their Namespaces.

Arguably, we could even aggregate to Kubernetes edit role.

### Motivation

Suggested by @AndreKR - https://github.com/traefik/traefik-helm-chart/issues/461

### More

- [X] Yes, I updated the [chart version](https://github.com/traefik/traefik-helm-chart/blob/9b32ed1b414cc0be1ad46bcb335fcfc93ded06f3/traefik/Chart.yaml#L5)

PR #664, opened 2022-10-13
feat: Add `aggregate-to-admin` label on ClusterRole (#464)

### LINKED ISSUE #461: Aggregate CRD access to default user-facing roles
### Welcome!

I'm *relatively* new to Kubernetes, so there might be a good reason not to do this, but let me describe my experience.

In a deploy script I'm creating an Ingress with a path and I'd like to strip that path. In Traefik 1.7 I used `PathPrefixStrip`, but I'm migrating to Traefik 2.4. So in my deploy script I tried to create a StripPrefix Middleware for the respective path currently being deployed.

I got an error:

```
User "system:serviceaccount:deploy-413:deploy-413-service-account" cannot get resource "middlewares" in API group "traefik.containo.us" in the namespace "deploy-413"
```

My deploy job is running with Kubernetes' [default `admin` role](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles).

I was curious why I can for example create [cert-manager](https://cert-manager.io/) CRDs but not Traefik CRD. It turns out that the cert-manager Helm chart creates [a ClusterRole](https://github.com/jetstack/cert-manager/blob/1b7e706c72655b9244fa547c3029d89ee9cb01e2/deploy/charts/cert-manager/templates/rbac.yaml#L434-L452) that has `rbac.authorization.k8s.io/aggregate-to-admin: "true"` set.

I think the Traefik Helm chart should set `rbac.authorization.k8s.io/aggregate-to-admin: "true"` on the ClusterRoles to make sure that [predefined admins](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles) can automatically create resources like Middleware.

mloiseleur:
### What does this PR do?

It's an updated version of #464, since I cannot rebase PR of @faust64. 
> Users with admin privileges should be able to operate Traefik resources within their Namespaces.

### Motivation

Fixes #461 

### More


charlie-haley (approved):
LGTM

PR #666, opened 2022-10-14
:bug: Allow to watch on specific namespaces without using rbac.namespaced

### LINKED ISSUE #395: Watch specific namespaces without using rbac.namespaced
Allow using 

```yaml
  providers:
    kubernetesCRD:
      enabled: true
      namespaces:
        - default
    kubernetesIngress:
      enabled: true
      namespaces:
        - default
```

without making the traefik deployment namespaced with `rbac.namespaced`

Use case:
Upgrading a cluster incrementally and migrating traefik configuration one namespace at a time

I believe this line should be changed 
https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/_podtemplate.tpl#L124

to check if `len(namespaces) >0`

mloiseleur:
### What does this PR do?

It allows to configure namespaces on provider as expected, even without using `rbac.namespaced`

### Motivation

Fixes #395 

### More

### DIFF
--- a/traefik/Changelog.md
+++ b/traefik/Changelog.md
@@ -1,5 +1,33 @@
 # Change Log
 
+## 15.2.0 
+
+**Release date:** 2022-10-17
+
+![AppVersion: 2.9.1](https://img.shields.io/static/v1?label=AppVersion&message=2.9.1&color=success&logo=)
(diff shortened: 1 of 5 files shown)


charlie-haley (approved):
LGTM

PR #668, opened 2022-10-17
:bug: Set dashboard ingressroute as an Helm resource

### LINKED ISSUE #115: Disabling dashboard ingressroute does not delete it
If during initial setup you go with default settings then dashboard ingressroute is created. If you disable it afterwards and apply, ingressroute is still present and accessible.
I assume, if you disable dashboard, this ingress route should be deleted, even `diff` is showing that it will be deleted.
I suspect that actual problem is that the this route is created as hook.

mloiseleur:
### What does this PR do?

Set dashboard `IngressRoute` as a regular helm resource instead of a hook.

### Motivation

It's not needed anymore to install it with a hook. 
Fixes #115

### More

- [ ] Is there a way to avoid break during upgrade between hook version and regular version ? 
```bash
Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. 
Unable to continue with update: IngressRoute "traefik-dashboard" in namespace "default" exists and cannot be imported into the current release: invalid ownership metadata;
annotation validation error: missing key "meta.helm.sh/release-name": must be set to "traefik"; 
annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "default"
```

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 15.2.0
+version: 15.2.1
 appVersion: 2.9.1
 keywords:
   - traefik
--- a/traefik/templates/dashboard-ingressroute.yaml
+++ b/traefik/templates/dashboard-ingressroute.yaml
@@ -4,7 +4,6 @@ kind: IngressRoute
 metadata:
   name: {{ template "traefik.fullname" . }}-dashboard
   annotations:
-    helm.sh/hook: "post-install,post-upgrade"
     {{- with .Values.ingressRoute.dashboard.annotations }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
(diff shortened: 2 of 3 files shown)


jnoordsij:
https://github.com/traefik/traefik-helm-chart/issues/763#issuecomment-1343036762
> Thanks for this report, @jnoordsij. For #668, that's right, but it may break upgrade path for users who haven't yet upgraded their Helm Chart to a recent versions. Maybe #668 can be merged when this Helm Chart will switch to Traefik Proxy v3.0.

Merging this with a Traefik Proxy v3.0 major chart upgrade sounds like a good idea. Maybe add an upgrade note in that case? I think manually deleting the old ingressroute, using something like `kubectl delete ingressroute traefik-ingress-controller-dashboard`, then upgrading the chart should work. This would cause minimal downtime on the dashboard, but I think that's not a huge issue for most people.

mloiseleur:
@jnoordsij This command will be needed only when upgrading from v15.3.0 or below. Since v15.3.1 (and PR #674), there is no command needed since all needed annotations are provided for the switch.

maxnitze:
Really looking forward to this change! It currently kind-of breaks my setup with ArgoCD (see https://community.traefik.io/t/why-dashboard-ingress-in-helm-chart-with-post-install-and-post-upgrade-hooks/17235)

Is there any chance this will be backported to Traefik v2?

PR #673, opened 2022-10-18
Add forwardedHeaders and proxyProtocol config

hesstobi:
### What does this PR do?

It extend the port configuration for the forwardedHeaders and proxyProtocol configuration per entrypoint.

### Motivation

With the current version of the helm chart, the forwardedHeaders and proxyProtocol configuration must be configured with additionalArguments.  This is somehow error-prone because it will be easily forgotten, when adding a new entrypoint.

### More

### DIFF
--- a/traefik/Changelog.md
+++ b/traefik/Changelog.md
@@ -1,5 +1,58 @@
 # Change Log
 
+## 16.2.0 
+
+**Release date:** 2022-10-20
+
+![AppVersion: 2.9.1](https://img.shields.io/static/v1?label=AppVersion&message=2.9.1&color=success&logo=)
+![Helm: v3](https://img.shields.io/static/v1?label=Helm&message=v3&color=informational&logo=helm)
+
+
+* Add forwardedHeaders and proxyProtocol config 
+
+### Default value changes
+
+```diff
+diff --git a/traefik/values.yaml b/traefik/values.yaml
+index 9b5afc4..6a90bc6 100644
+--- a/traefik/values.yaml
++++ b/traefik/values.yaml
+@@ -403,6 +403,16 @@ ports:
+     # Added in 2.2, you can make permanent redirects via entrypoints.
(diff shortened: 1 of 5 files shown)


hesstobi:
@mloiseleur

ok but i don't know how to test this.

When i have this test case

``` yaml
  - it: should not set forwardedHeaders and proxyProtocol configuration by default
    asserts:
      - notContains:
          content: ""--entrypoints.websecure.forwardedHeaders.trustedIPs"
      - notContains:
          content: ""--entrypoints.websecure.proxyProtocol.trustedIPs"
````

it will not fail if the value of trustedIPs are set to a value.

Of course i could test complete content of `spec.template.spec.containers[0].args` with

``` yaml
  - it: schould set the the container args by default
    asserts:
      - equal:
          content:
            - --global.checknewversion
            - --global.sendanonymoususage
            - --entrypoints.metrics.address=:9100/tcp
            - --entrypoints.traefik.address=:9000/tcp
            - --entrypoints.web.address=:8000/tcp
            - --entrypoints.websecure.address=:8443/tcp
            - --api.dashboard=true
            - --ping=true
            - --metrics.prometheus=true
            - --metrics.prometheus.entrypoint=metrics
            - --providers.kubernetescrd
            - --providers.kubernetesingress
            - --entrypoints.websecure.http.tls=true
```

In this case it would fail, but this would require to change the test every time when there is a change in the default settings.

Do you have any hints or succession for me.

By the way, the tests:
-  `it: should not set namespace on providers by default`
- `it: should not have the watched namespace when provider is disabled`
will also not fail if the default config is changing.

mloiseleur (approved):
You're right.
There is a MatchRegex, but it cannot be used on Arrays. 
So I guess that, apart from a better use of _snapshot_, we cannot. 
Thanks for this PR. We will merge it soon.

charlie-haley (approved):

PR #700, opened 2022-11-03
🎨 Don't require exposed Ports when enabling Hub

mloiseleur:
### What does this PR do?

Allows to install Traefik with Traefik Hub, enabled Service and without exposed Ports.

### Motivation

When Traefik Hub is enabled, it makes sense to support no Service by default. It allows 
```bash
helm upgrade --install traefik traefik/traefik \
  --namespace hub-agent --create-namespace \
  --set hub.enabled=true \
  --set ports.web=null --set ports.websecure=null 
```
instead of 
```bash
helm upgrade --install traefik traefik/traefik \
  --namespace hub-agent --create-namespace \
  --set hub.enabled=true \
  --set ports.web=null --set ports.websecure=null --set service.enabled=false
```

### More

### DIFF
--- a/traefik/Changelog.md
+++ b/traefik/Changelog.md
@@ -1,14 +1,30 @@
 # Change Log
 
-## 19.0.2 
+## 19.0.3 
 
-**Release date:** 2022-10-31
+**Release date:** 2022-11-03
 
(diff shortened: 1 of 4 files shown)


darkweaver87 (approved):
LGTM

charlie-haley (approved):

PR #707, opened 2022-11-08
🐛 fix chart releases

charlie-haley:
### What does this PR do?
Helm [chart-releaser-action](https://github.com/helm/chart-releaser-action) doesn't support remote repositories for pushing artifacts. This PR moves to [helm-gh-pages](https://github.com/stefanprodan/helm-gh-pages) which supports remote commits.


mloiseleur (approved):
LGTM

PR #710, opened 2022-11-08
🔧 Adds more settings for dashboard ingressRoute

ksubileau:
### What does this PR do?

Add `matchRule` and `middlewares` settings on dashboard IngressRoute.
This allows adding custom middlewares and overriding the default match rule with a custom one.

### Motivation

To expose the dashboard in a secure way, it can be useful to be able to add middlewares (defined either outside the chart or using the `extraObjects` parameter) on the IngressRoute. It may also be interesting to add other middlewares, such as a *RedirectRegex* to redirect requests from `http://traefik.example.com/` to `http://traefik.example.com/dashboard/`.

Also, as [recommended in the documentation](https://doc.traefik.io/traefik/operations/dashboard/#dashboard-router-rule), it may be useful to override the router rule to specify a domain to match, or accept requests on the root path (`/`) in order to redirect them to `/dashboard/`.

As an example use case, with this pull request, we can configure the chart like this:
```yaml
# Create an IngressRoute for the dashboard
ingressRoute:
  dashboard:
    enabled: true
    # Custom match rule with host domain
    matchRule: Host(`traefik.example.com`)
    entryPoints: ["websecure"]
    # Add custom middlewares : authentication and redirection
    middlewares:
      -name: traefik-dashboard-redirect
      -name: traefik-dashboard-auth

# Create the custom middlewares used by the IngressRoute dashboard (can also be created in another way).
extraObjects:
  - apiVersion: traefik.containo.us/v1alpha1
    kind: Middleware
    metadata:
      name: traefik-dashboard-auth
    spec:
      basicAuth:
        secret: traefik-dashboard-auth-secret

  - apiVersion: traefik.containo.us/v1alpha1
    kind: Middleware
    metadata:
      name: traefik-dashboard-redirect
    spec:
      redirectRegex:
        permanent: true
        regex: ^(https?:\/\/(\[[\w:.]+\]|[\w\._-]+)(:\d+)?)\/$
        replacement: ${1}/dashboard/
```

### More


mloiseleur:
Thanks for this PR @ksubileau !
It may fix #186, also.

mloiseleur (approved):
LGTM

project0 (approved):

ksubileau:
Thank you ❤️

jbergler:
Awesome, thanks for doing this!

PR #725, opened 2022-11-16
Add overwrite option for instance label value

### LINKED ISSUE #724: Add template option for instance label
### Welcome!

As described in the [Readme](https://github.com/traefik/traefik-helm-chart/blob/master/README.md#upgrading-from-16x-to-17x), we need to remove certain resources in order to upgrade the chart from 16.x to 17.x.
Alternativly, you propose to setup an additional installation next to the existing one and switch over.

In our case, the LB-Service expliclty binds certain AWS Elastic IPs to the LoadBalancer - this means, we cannot easily setup another instance and move over without causing a downtime. Besides that, the needed changes to the IaC just to make this upgrade happen are not to be neglected.

Therefore, we'd propose to add a templating option to this line https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/_helpers.tpl#L46, so the user is able to configure the value of the label. This way, he can set the value to the previous values and without avoid the "breaking" change causing downtime.

I'll prepare a corresponding MR.

grieshaber:
Signed-off-by: Frederik Grieshaber <freddy.grieshaber+github@gmail.com>

Resolve #724 

### What does this PR do?

This PR adds the option to customize the value of the `app.kubernetes.io/instance:` label via a new value `instanceNameOverride`.
As both nameOverride and fullnameOverride is not provided in the default values.yaml (not even as a comment), i decided to omit the new override value aswell.

### Motivation

As described in #724, we want to avoid a (longer lasting) downtime at one of our installations. Therefore, we need the ability to overwrite the breaking change to fit the previous settings.

### More

- [-] Yes, I updated the tests accordingly (there were no tests for the selector behaviour at all)

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ order to be able to upgrade.
 You may also upgrade by deploying another Traefik to a different namespace and
 removing after your first Traefik.
 
-Alternatively, since version 20.3.0 of this chart, you may set `instanceNameOverride` to the previous value of that label.
+Alternatively, since version 20.3.0 of this chart, you may set `instanceLabelOverride` to the previous value of that label.
 This will override the new `Release.Name-Release.Namespace` pattern to avoid any (longer) downtime.
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -146,6 +146,9 @@ order to be able to upgrade.
 You may also upgrade by deploying another Traefik to a different namespace and
 removing after your first Traefik.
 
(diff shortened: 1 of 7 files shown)


charlie-haley:
Thanks for the PR @grieshaber! Could you add some tests for the selector behavior?

grieshaber:
@charlie-haley i did and rebased to the latest changes from master. :)

charlie-haley (approved):

mloiseleur:
Thanks @grieshaber !

PR #727, opened 2022-11-18
Add (optional) dedicated metrics service

### LINKED ISSUE #726: Decouple service serving metrics from hub-service
### Welcome!

## Issue
With our latest traefik chart upgrade, we now make use of the built-in ServiceMonitor object.
Unfortunately, is does not work otb because there is no suitable service exposing the metrics port to that ServiceMonitor.
The only service, that references the metrics port is the `traefik hub service` - which is (stated from the documentation) explicitly set up for the usage traefik-hub. 

## Proposal
Template and allow the creation of an additional service, which will serve the metrics endpoint.
So far, one could do it via the `extraObjects`-template, but that is not very handy.

grieshaber:
Signed-off-by: Freddy Grieshaber <freddy.grieshaber+github@gmail.com>

Resolves #726 

### What does this PR do?

This PR adds a dedicated service object to serve the metrics port when Traefik Hub is not enabled.

### Motivation

Until now, it was coupled with the "hub" service. As one might want to use the metrics & servicemonitor without enabling the hub-integration, this PR decouples the hub and the metrics feature.

### More

### DIFF
--- a/traefik/templates/service-metrics.yaml
+++ b/traefik/templates/service-metrics.yaml
@@ -1,0 +1,24 @@
+{{- if (and .Values.metrics.prometheus.service.enabled (not .Values.hub.enabled)) -}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: traefik-metrics
+  namespace: {{ template "traefik.namespace" . }}
+  {{- template "traefik.metrics-service-metadata" . }}
+  annotations:
+  {{- with .Values.metrics.prometheus.service.annotations }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  type: ClusterIP
+  selector:
+    {{- include "traefik.metricslabelselector" . | nindent 4 }}
+  ports:
+  - port: {{ .Values.ports.metrics.port }}
+    name: "metrics"
+    targetPort: metrics
+    protocol: TCP
+    {{- if .Values.ports.metrics.nodePort }}
+    nodePort: {{ .Values.ports.metrics.nodePort }}
+    {{- end }}
+{{- end -}}
--- a/traefik/templates/service-metrics.yaml
+++ b/traefik/templates/service-metrics.yaml
@@ -1,8 +1,13 @@
-{{- if (and .Values.metrics.prometheus.service.enabled (not .Values.hub.enabled)) -}}
+{{- $fullname := include "traefik.fullname" . }}
+{{- if ge (len $fullname) 50 }}
+  {{- fail "ERROR: Cannot create a metrics service when name contains more than 50 characters" }}
+{{- end }}
+
+{{- if (and (.Values.metrics.prometheus.service).enabled (not .Values.hub.enabled)) -}}
 apiVersion: v1
 kind: Service
 metadata:
-  name: traefik-metrics
+  name: {{ $fullname }}-metrics
   namespace: {{ template "traefik.namespace" . }}
   {{- template "traefik.metrics-service-metadata" . }}
   annotations:
@@ -12,7 +17,7 @@ metadata:
 spec:
   type: ClusterIP
   selector:
-    {{- include "traefik.metricslabelselector" . | nindent 4 }}
+    {{- include "traefik.labelselector" . | nindent 4 }}
   ports:
   - port: {{ .Values.ports.metrics.port }}
     name: "metrics"
--- a/traefik/templates/service-metrics.yaml
+++ b/traefik/templates/service-metrics.yaml
@@ -27,3 +27,4 @@ spec:
     nodePort: {{ .Values.ports.metrics.nodePort }}
     {{- end }}
 {{- end -}}
+
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -266,9 +266,9 @@ metrics:
   #    address: localhost:8125
 ##
 ##  enable optional CRDs for Prometheus Operator
-## 
-    # create a dedicated metrics service for use with servicemonitor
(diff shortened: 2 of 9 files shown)


mloiseleur:
Hello @grieshaber, 

Thanks for your interest and this second PR in two days !
It's not only changing, this PR will **break** current usage (and documentation) around Traefik Hub with Traefik Proxy.

I have two different approach to suggest:
a. Since this Chart provides an internal service, how about putting the metrics on this internal `Service` ?
b. Since this `Service` is only needed with `ServiceMonitor`, adds an option in ServiceMonitor to create this new `Service` ?

grieshaber:
Hey @mloiseleur,

thanks for the fast feedback!

I thought about these options as well. I decided for that approach in order to keep the flexibility as high as possible.
But, as your argument is absolutely valid, i'm happy to switch to any other approach.

Regarding option a)
That was the alternative i had in mind as well. I would then need to change the selector of the ServiceMonitor to point to only the new service (currently it uses the templated selectorLabel value). 
Otherwise, it would potentially discover two services with metrics enabled (when Hub is used). If this specific service is not set to true, we could use the default selectors which will then discover the hub.
I'll tend to make use of this option and prepare the changes.

Regarding option b)
I understood, that via this service object one could also create an additional (internal) LoadBalancer. In that case, i would like to keep the metrics away from the "exposed" service. As the type is flexibel for that Service, i think it would add some more complexity and potential for misunderstanding

traefik/values.yaml
```diff
@@ -267,6 +267,7 @@ metrics:
 ##
 ##  enable optional CRDs for Prometheus Operator
 ##
+  #  createMetricsService: true
```
charlie-haley, line 270: We should rename the value to something like this, it will give us more flexibility for configuration in the future
```suggestion
  #  service:
  #      enabled: true
```

grieshaber: Exactly what i thought about while implementing the other approach. I also added labels and annoations for that service, so that the least minimum is configurable

grieshaber:
@mloiseleur @charlie-haley i implemented more or less option b) of the suggested alternatives.
Although it feels a little over-engineered now, i' pretty confident that with that solution everything is as fleixble as needed. 

I'll just add another test-case, then i from my pov it should be good to go :)

traefik/values.yaml
```diff
@@ -266,7 +266,13 @@ metrics:
   #    address: localhost:8125
 ##
 ##  enable optional CRDs for Prometheus Operator
-##
+## 
+    # create a dedicated metrics service for use with servicemonitor
+    # will only be created, when enabled set to true and hub.enabled is false
+    service:
+      enabled: false
```
mloiseleur, line 273: This should be commented and indented following [GuideLines](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/Guidelines.md#optional-non-critical-feature-example)

mloiseleur (changes requested):

traefik/values.yaml
grieshaber: Ok - i did. I used a different solution for the nested nil issue
```
{{- if (and (.Values.metrics.prometheus.service).enabled (not .Values.hub.enabled)) -}}
```
instead of 
```
{{- if (and .Values.metrics.prometheus.service -}}
{{- if (and .Values.metrics.prometheus.service.enabled (not .Values.hub.enabled)) -}}
```

I did so, because otherwise the if-clause in the servicemonitor labels would become overly complex. Hope that's fine for you.

traefik/templates/service-metrics.yaml
```diff
@@ -0,0 +1,24 @@
...
+  namespace: {{ template "traefik.namespace" . }}
+  {{- template "traefik.metrics-service-metadata" . }}
+  annotations:
+  {{- with .Values.metrics.prometheus.service.annotations }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  type: ClusterIP
+  selector:
+    {{- include "traefik.metricslabelselector" . | nindent 4 }}
```
mloiseleur, line 15: With a selector on `app.kubernetes.io/component: metrics`, how this service will select a Traefik Pod ?

traefik/templates/service-metrics.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if (and (.Values.metrics.prometheus.service).enabled (not .Values.hub.enabled)) -}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: traefik-metrics
```
mloiseleur, line 5: You should use `traefik.fullname`. See service-internal.yaml for an example

mloiseleur (changes requested):
It looks good. 
There are two small issues to fix, one test to add (ensure service metrics selector is using traefik pod label) and we're good to go.

traefik/templates/service-metrics.yaml
grieshaber: Oh of course - i fixed it and use the default selector labels.

grieshaber: The previous test failed, so i removed the expected value (comp: metrics) and it went green. Imho that's exactly the test-case to requested, isn't it?

traefik/templates/service-metrics.yaml
grieshaber: I did not see that one - c&p the logic from internal service into the metrics service.

mloiseleur (approved):
LGTM

grieshaber:
Thanks @mloiseleur 👍🏼

charlie-haley (approved):
LGTM

Issue #730, opened 2022-11-18
I believe the Cluster Role Binding should check for the value namespaceOverride


savager:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

20.2.0

### What version of Traefik are you using?

2.9.4

### What did you do?

Install as a dependency in our Helm chart.

Use these as the value overrides:
```
traefik:
  namespaceOverride: traefik
  namespace: traefik
  serviceAccountName: dxc-traefik
  serviceAccount:
    name: dxc-traefik
```

### What did you see instead?

Verified that the cluster role binding does not use the service account in the namespace value given "traefik". It uses the release.namespace https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/rbac/clusterrolebinding.yaml

Manually editing the CRB service account namespace to traefik did fix the problem.

### What is your environment & configuration?

Basic helm chart no other dependencies. Yaml

I saw a similar question here with no response.

https://community.traefik.io/t/why-is-the-rbac-of-traefik-ingress-controller-restricted-to-namespace-default/3853

mloiseleur:
Thanks for your report @savager. 
You're right.

savager:
Thanks so much for the quick fix and helpful install!

PR #733, opened 2022-11-21
🐛 fix namespace references to support namespaceOverride

charlie-haley:
### What does this PR do?
This PR updates the remaining namespace references to support the override. Fixes https://github.com/traefik/traefik-helm-chart/issues/730


mloiseleur (approved):
LVGTM

Issue #187, opened 2020-05-27
Provide a "default" Certificate for TLS


SantoDE:
If a user provides us his certificate data, we could create a `TLSStore` and attach the cert as a default certificate.

Partially fixes #185

flexchar:
I was searching a lot how I could swap Traefik auto generated certificates with mine provided as default ones. I think this is big issue so I'll make an informational post here.

Specifically I use Kubernetes and Ingresses with Cloudflare as DNS and Firewall provider. Cloudflare provides their own Edge certificates (for connection between CF and host server) that I aimed to upload and serve for all the hosts.

I created secret of type tls. And tried using TLSStore (found suggestion somewhere, sorry no link)
```yaml
apiVersion: traefik.containo.us/v1alpha1
kind: TLSStore
metadata:
  name: cloudflare-tls
  namespace: traefik
spec:
  defaultCertificate:
    secretName: tls-secret
```
but seems to be ignored by Kubernetes Ingress resources. This may work if you are using Kubernetes CRD [IngressRoute](https://doc.traefik.io/traefik/routing/providers/kubernetes-crd/)

Then I went the dynamic config way:
Created ConfigMap:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: traefik-config
  labels:
    name: traefik-config
  namespace: traefik
data:
  dyn.yaml: |
    # https://doc.traefik.io/traefik/https/tls/
    tls:
      stores:
        default:
          defaultCertificate:
            certFile: '/certs/tls.crt'
            keyFile: '/certs/tls.key'
```
and used these values when deploying Traefik chart
```yaml
additionalArguments:
  - '--providers.file.filename=/config/dyn.yaml'
volumes:
  - name: tls-secret
    mountPath: '/certs'
    type: secret
  - name: traefik-config
    mountPath: '/config'
    type: configMap
```
and it works!

It can be challenging to understand Traefik, especially when Kubernetes are involved, hope this will shed some light.

tillepille:
I would like to see the ConfigMap and therefore my custom configuration to be managed within this HelmChart as well.
So in `values.yaml`

```yaml
additionalConfig: |
   # https://doc.traefik.io/traefik/https/tls/
    tls:
      stores:
        default:
          defaultCertificate:
            certFile: '/certs/tls.crt'
            keyFile: '/certs/tls.key'
```

and a ConfigMap like 

```yaml
{{ if .Values.additionalConfig }}
[...]

PR #601, opened 2022-06-01
Add TLSStore resource to chart

DrFaust92:
Supersedes #428 

### What does this PR do?

Adds support for creating TLSStore CR as part of chart

### More

- [x] Yes, I updated the [chart version](https://github.com/traefik/traefik-helm-chart/blob/9b32ed1b414cc0be1ad46bcb335fcfc93ded06f3/traefik/Chart.yaml#L5)

PR #754, opened 2022-12-06
:memo: Add a complete example on TLS Store with Cert-Manager

mloiseleur:
### What does this PR do?

 Add a complete example on TLS Store with Cert-Manager

### Motivation

Users still have hard time to configure it. Last seen [here](https://github.com/traefik/traefik-helm-chart/issues/187#issuecomment-1338147700).

### More

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -206,3 +206,97 @@ env:
         name: yyy
         key: zzz
 ```
+
+# Provide default certificate with cert-manager and CloudFlare DNS
+
+Setup:
+
+* cert-manager installed in `cert-manager` namespace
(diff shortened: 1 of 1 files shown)


charlie-haley (approved):

samos667:
I have lose some hours with this issue: https://github.com/traefik/traefik-helm-chart/pull/601#issuecomment-1379608172

Maybe u can add a note about this for save others ^^

I have fix this with: `kubectl label -n <traefik_namespace> tlsstores.traefik.containo.us default environment=production,method=traefik`

environment=production,method=traefik is to be replaced by what u set in kubernetesCRD labelSelector in values.yaml

Issue #759, opened 2022-12-08
Helm rendering fails on metrics service even if i don't need it


alecerf:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

20.7.0

### What version of Traefik are you using?

v2.9.6

### What did you do?

I don't have `.Values.metrics.prometheus.service` set in my values file (so it should be disabled by default)

I execute `helm template --name-template=azertyuiopqsdfghjklmwxcvbnazertyuiopqsdfghjklmwxcvbn .`

### What did you see instead?

`Error: execution error at (ingress-traefik/charts/traefik/templates/service-metrics.yaml:7:6): ERROR: Cannot create a metrics service when name contains more than 50 characters`

### What is your environment & configuration?

I use the default Traefik values file + some override but not on for the metrics part.

### Additional Information

```shell
If i move the


{{- if .Values.metrics.prometheus }}
{{- if .Values.metrics.prometheus.service }}
{{- if (and (.Values.metrics.prometheus.service).enabled (not .Values.hub.enabled)) -}}
```

before

```
{{- $fullname := include "traefik.fullname" . }}
{{- if ge (len $fullname) 50 }}
  {{- fail "ERROR: Cannot create a metrics service when name contains more than 50 characters" }}
{{- end }}
```

on https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/service-metrics.yaml , then it works again.
```

mloiseleur:
Thanks for your report, @alecerf.

PR #765, opened 2022-12-09
fix(service-metrics): invert prometheus svc & fullname length checking

alecerf:
### What does this PR do?

Invert Prometheus metrics service creation check and the Traefik fullname length check.
Fix: https://github.com/traefik/traefik-helm-chart/issues/759

### Motivation

If we have a release name with more than 50 chars but we don't want to create a metrics service, then the Helm rendering will fail.

With this PR, it will not fail anymore.

### More


mloiseleur:
Thanks @alecerf . 
I added a test, just to be sure it won't happen again.

charlie-haley (approved):
Thanks @alecerf for the PR!

PR #766, opened 2022-12-12
Update CRDs to match the new api version traefik.io/v1

sdelicata:
### What does this PR do?

This PR updates CRDs to match the new API version.
`traefik.containo.us/v1alpha1` -> `traefik.io/v1`

Related to: https://github.com/traefik/traefik/pull/9600

### More

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -78,7 +78,7 @@ extraObjects:
       username: admin
       password: changeme
 
-  - apiVersion: traefik.containo.us/v1alpha1
+  - apiVersion: traefik.io/v1
     kind: Middleware
     metadata:
       name: traefik-dashboard-auth
@@ -130,7 +130,7 @@ extraObjects:
       username: admin
       password: changeme
 
-  - apiVersion: traefik.containo.us/v1alpha1
+  - apiVersion: traefik.io/v1
     kind: Middleware
     metadata:
       name: traefik-dashboard-auth
@@ -357,7 +357,7 @@ tlsStore:
 
 They should use websecure entrypoint like this:
 ```yaml
-apiVersion: traefik.containo.us/v1alpha1
+apiVersion: traefik.io/v1
 kind: IngressRoute
 metadata:
   name: example-com-tls
--- a/traefik/crds/ingressroute.yaml
+++ b/traefik/crds/ingressroute.yaml
@@ -1,3 +1,4 @@
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -5,9 +6,9 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.6.2
   creationTimestamp: null
-  name: ingressroutes.traefik.containo.us
+  name: ingressroutes.traefik.io
 spec:
-  group: traefik.containo.us
+  group: traefik.io
   names:
     kind: IngressRoute
     listKind: IngressRouteList
@@ -15,7 +16,7 @@ spec:
     singular: ingressroute
   scope: Namespaced
   versions:
-  - name: v1alpha1
+  - name: v1
     schema:
       openAPIV3Schema:
         description: IngressRoute is the CRD implementation of a Traefik HTTP Router.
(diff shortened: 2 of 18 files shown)


z0rc:
Oh boy, this is going to be so much PITA for end users.

There should be a transition period for at least couple of traefik's minor releases (ideally for a whole major release) where trafik accepts BOTH old and new API versions. If you decide to just make a hard cut from old api version to new one, this will:
- inability for new traefik to read existing resources
- if old crds are removed, existing custom resources will be removed too

This, even draft, change is a set up to outages once users start to upgrade to traefik v3.

I, as a cluster operator, need a transition period, so I can instruct users of clusters with traefik ingress controller installed, to work on updating their custom resources to use new API version. This process will take time, and we need to keep support of of both api versions.

Have you seen how Kubernetes itself handles api deprecation? It takes 3 minor versions (or about a year) from deprecation to api removal.

ldez:
The v2 will still be maintained after the release of the v3.

This PR is just a draft.

z0rc:
@ldez you ignored all my arguments about crd management. My question is not about traefik v2, but how traefik handles migration of crd api versions.

DrFaust92:
consider 2 things

1 .https://artifacthub.io/packages/helm/linkerd2-edge/linkerd-crds doing like linkerd and split charts.
2. having upgrade notes like https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-42x-to-43x

oscrx:
I assume some is this is already included in version 22.3.0?
https://github.com/traefik/traefik-helm-chart/compare/v22.2.0...v22.3.0

sjmiller609:
It might be nice to start a v3 branch or similar on this repository to prepare changes for Traefik version 3. Since the plan is to [support only traefik.io/v1 in version 3](https://github.com/traefik/traefik/pull/9600#issuecomment-1462254130), this would allow for installing Traefik version 3 with the correct CRDs.

My interest here is just to start off on version 3 for a feature my team needs that is only available in Traefik v3 (Postgres StartTLS support). I am working on getting [Postgres StartTLS](https://github.com/traefik/traefik/pull/9377) + SNI routing working with [ipAllowList](https://doc.traefik.io/traefik/v3.0/migration/v2-to-v3/#ipwhitelist), which is how I ended up looking into API versions of Traefik.

I'm not yet sure if the pre-release version 3 of Traefik is already supposed to work with the v1 CRDs, or if that is forthcoming. Currently, I am using the old API versions and it is working for StartTLS and ignoring MiddlewareTCP (or I didn't figure out how to set it properly yet).

sjmiller609:
I explained more about the above reasoning here: https://github.com/traefik/traefik/pull/9600#issuecomment-1753313008

gcaracuel:
Any progress on this? 
This is blocking usage of ipAllowList when running v3 and now v2 go a step further (https://github.com/traefik/traefik/releases/tag/v2.11.0-rc2 ) while Helm Chart keeps discussing the solution...

rtribotte:
Hello,

Because we took into consideration the feedback we had regarding the ease of migration to v3.0, the Traefik v3.0 release strategy has been revisited (https://github.com/traefik/traefik/issues/10260).
The apiVersion `traefik.io/v1` is no longer introduced in v3.0, and one, when migrating from v2 to v3, should only update the apiVersion `traefik.io/v1alpha1` CRDs.

gcaracuel:
One more time... https://github.com/traefik/traefik-helm-chart/blob/master/traefik/crds/traefik.io_middlewares.yaml#L605 is still wrong and a breaking change if apiVersion is reused to the one introduced by https://github.com/traefik/traefik/pull/9765 

Part of this PR still makes sense but changing apiVersion back to v1alpha1. Yes this cause problems, but all this mess with migration strategies changes still not considering this breaking changes.

mloiseleur:
@gcaracuel This Chart support _stable_ version of Traefik Proxy. The CRDs of this charts follows Traefik Proxy stable version and so, they are _correct_. If one wants to use both this Chart and Traefik Proxy v3 _beta_, it's at its own risk. For doing this, one needs to deploy v3 beta CRDs without the Chart and after to deploy the Chart without CRDs.

BTW, there is work in progress to make the transition as smooth as possible between v2 and v3, see for instance [this PR](https://github.com/traefik/traefik/pull/10355). 

When the Traefik Proxy v3 is ready and becomes _stable_, a new major version of this Chart will include valid CRDs for Traefik Proxy v3.

Is this more clear for you ?

PR #777, opened 2023-01-09
Fix existing tag detection in workflow

jnoordsij:
### What does this PR do?

Fix failing workflows for commits on `master` with existing tags.

### Motivation

Currently the step `Check if tag exists` on the `release` job fails to detect existing tags, because it does not account for the prefix `v` added (by default) by the `Tag release` action. This PR explicitly adds the prefix to both stages.

See also [`succesful example run on my fork`](https://github.com/jnoordsij/traefik-helm-chart/actions/runs/3874275347/jobs/6605281682) and [`currently failing run on master`](https://github.com/traefik/traefik-helm-chart/actions/runs/3828467971/jobs/6514093053).

### More

- [ ] ~~Yes, I updated the tests accordingly~~
- [ ] ~~Yes, I ran `make test` and all the tests passed~~

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - master
 
+env:
+  tag_prefix: v
+
 jobs:
   test:
     uses: "traefik/traefik-helm-chart/.github/workflows/test.yml@master"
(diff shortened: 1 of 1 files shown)


charlie-haley (approved):
Thanks for the PR! LGTM

mloiseleur (approved):
Interesting. Thanks @jnoordsij , it looks good !

Issue #625, opened 2022-08-25
Update this Chart to track each Traefik release?


rgl:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Please make this Chart (e.g. appVersion) track the https://github.com/traefik/traefik releases.

Even thou we can set the image.tag chart value, that makes it harder to the user to known which chart versions are compatible with which traefik versions. It also means, we'll end up updating two versions, instead of one.

ldez:
Hello,

The `appVersion` is mainly updated on a minor version (v1.**X**.y) of Traefik, the bug fix releases (v1.x.**Y**) of Traefik are always compatible with the minor version.

z0rc:
@ldez why? It much more practical to have a chart release for each Traefik release. This is common practice across helm community. Majority of chart users don't track traefik releases, they do track helm repo. Your approach breaks end user release cycle, where have to follow patch releases, but skip the minor and major.

z0rc:
> everybody can create a PR

Problem with this, that community contributions remain stale for months without review. While PRs from traefik team members merged in seconds, like this https://github.com/traefik/traefik-helm-chart/pull/614.

ldez:
we are working on a new maintenance mode.

shadycuz:
I could help with the CICD of this... perhaps have something like dependapot update the docker tag for traefik and raise a PR that does some basic tests to ensure things are working? Even just getting the PR up might be nice.

mloiseleur:
Yes, sure. It would be awesome, @shadycuz !

shadycuz:
@mloiseleur 

Here are the rough steps that would need to happen.

1. The maintainers of this repo would need to enable dependabot or renovate to handle dependency updates. I personally prefer renovate and it would probably do a better job but I think dependabot might also work for this.
2. We would need to create a dummy `Dockerfile` with the line ` from traefik:v2.9.6` inside the repo. This is the file that dependabot/renovate will watch and if a new version is released than a new PR will be opened. 
[...]

PR #784, opened 2023-02-10
:boom: Rename image.name => image.repository

mloiseleur:
### What does this PR do?

Rename helm value from `image.name` to `image.repository`
It's just a renaming, no change on the behaviour.

### Motivation

It's in fact a repository path and it's needed for Renovate compatibility.
See #625 & https://github.com/renovatebot/renovate/issues/7010

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -19,7 +19,7 @@ Create the chart image name.
 */}}
 
 {{- define "traefik.image-name" -}}
-{{- printf "%s:%s" .Values.image.name (.Values.image.tag | default .Chart.AppVersion) }}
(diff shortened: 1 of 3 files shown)


charlie-haley (approved):
LGTM

juanitoddd:
Friends, Its not just a rename, This broke our production system.
It needs backwards compatibility !

charlie-haley:
@juanitoddd when releasing we follow semver. We released this by incrementing a major version (X.0.0), this means it will include breaking changes.
As mentioned before, we try not make these changes unless we have to. I'd recommend checking the `traefik/Changelog.md` and GitHub Release before updating a major version in the future to ensure you're not affected.

PR #794, opened 2023-02-15
fix: traefik image name for renovate

mmatur:
### What does this PR do?

This PR fix the image name for renovate. 

Traefik is built as an official image.


ldez (approved):
LGTM

PR #799, opened 2023-02-17
Separate UDP hostPort for HTTP/3

jilleJr:
### What does this PR do?

Changed the podSpec to have 2 separate ports for HTTP/3 when using hostPort

### Motivation

The Kubernetes Service provides multiple port specs for HTTP/3, with one for TCP and one for UDP. But when using hostPort, the host bypasses the Service and goes straight for the container. Because of this, the duplicate ports spec needs to be specified both in the Service (for inter-pod usage) and inside the podSpec (for hostPort usage).

My setup needs this, because I don't want to pay 10$ more per month to my cloud provider just because I need to use a LoadBalancer Service, especially since I only have a small single-node cluster.

### More

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -76,6 +76,15 @@
           hostIP: {{ $config.hostIP }}
           {{- end }}
           protocol: {{ default "TCP" $config.protocol | quote }}
+        {{- if $config.http3 }}
+        {{- if and $config.http3.enabled $config.hostPort }}
+        {{- $http3Port := default $config.exposedPort $config.http3.advertisedPort }}
+        - name: "{{ $name }}-http3"
+          containerPort: {{ $config.port }}
+          hostPort: {{ $config.hostPort }}
+          protocol: UDP
+        {{- end }}
+        {{- end }}
         {{- end }}
         {{- end }}
         {{- if .Values.hub.enabled }}
--- a/traefik/tests/pod-config_test.yaml
+++ b/traefik/tests/pod-config_test.yaml
@@ -464,3 +464,42 @@ tests:
           of: Deployment
       - isAPIVersion:
           of: apps/v1
+  - it: should have only one http3 port when not using hostport
+    set:
+      ports:
+        websecure:
+          http3:
+            enabled: true
+    asserts:
+      - contains:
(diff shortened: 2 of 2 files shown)


traefik/templates/_podtemplate.tpl
```diff
@@ -76,6 +76,15 @@
           hostIP: {{ $config.hostIP }}
           {{- end }}
           protocol: {{ default "TCP" $config.protocol | quote }}
+        {{- if $config.http3 }}
+        {{- if and $config.http3.enabled $config.hostPort }}
+        {{- $http3Port := default $config.exposedPort $config.http3.advertisedPort }}
```
project0, line 81: `$http3Port` does not seem to be used here?

traefik/templates/_podtemplate.tpl
```diff
@@ -76,6 +76,15 @@
           hostIP: {{ $config.hostIP }}
           {{- end }}
           protocol: {{ default "TCP" $config.protocol | quote }}
+        {{- if $config.http3 }}
```
project0, line 79: what happens if `http3.enabled` is set to false?

project0: never mind, its catched below :+1:

traefik/templates/_podtemplate.tpl
jilleJr: Oops, you're right. Too much copy paste from me.

```suggestion
```

project0:
I think you need to respect advertisedPort, it influences on what port the udp entrypoint is configured when set.

see also:
https://github.com/traefik/traefik-helm-chart/blob/6c6de90e7fa31875fc317f5bfc58f65aaf490f31/traefik/templates/_podtemplate.tpl#L537

jilleJr:
> I think you need to respect advertisedPort, it influences on what port the udp entrypoint is configured when set.
> 
> see also:
> 
> https://github.com/traefik/traefik-helm-chart/blob/6c6de90e7fa31875fc317f5bfc58f65aaf490f31/traefik/templates/_podtemplate.tpl#L537

Good note! I've added that in e59e883

project0 (approved):
LGTM

mloiseleur (approved):
LGTM

darkweaver87 (approved):
LGTM

Issue #800, opened 2023-03-01
Chart installation failed on latest k8s version


furynick:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

21.1.0

### What version of Traefik are you using?

v2.9.7

### What did you do?

```
helm install \
  --namespace traefik \
  --create-namespace \
  --set service.externalIPs=x.x.x.x \
  --set ingressRoute.dashboard.enabled=false \
  --set autoscaling.enabled=true \
  --set autoscaling.maxReplicas=4 \
  --set autoscaling.minReplicas=2 \
  --set hostNetwork=true \
  --set podSecurityPolicy.enabled=true \
  --set securityContext.runAsNonRoot=false \
  --set logs.general.format=json \
  --set logs.access.enabled=true \
  --set ports.traefik.exposedPort=8080 \
  --set ports.web.exposedPort=80 \
  --set ports.websecure.exposedPort=443 \
  --set ports.amqp-pf.port=56720 \
  --set ports.amqp-pf.exposedPort=5672 \
  --set ports.amqp-pf.protocol=TCP \
  --set ports.amqp-fmk.port=56721 \
  --set ports.amqp-fmk.exposedPort=56721 \
  --set ports.amqp-fmk.protocol=TCP \
  --set ports.amqp-app1.port=56722 \
  --set ports.amqp-app1.exposedPort=56722 \
  --set ports.amqp-app1.protocol=TCP \
  --set ports.amqp-app2.port=56723 \
  --set ports.amqp-app2.exposedPort=56723 \
  --set ports.amqp-app2.protocol=TCP \
  --set ports.memcache-pf.port=11210 \
  --set ports.memcache-pf.exposedPort=11210 \
  --set ports.memcache-pf.protocol=TCP \
  --set ports.memcache-fmk.port=11211 \
  --set ports.memcache-fmk.exposedPort=11211 \
  --set ports.memcache-fmk.protocol=TCP \
  --set ports.memcache-app1.port=11212 \
  --set ports.memcache-app1.exposedPort=11212 \
  --set ports.memcache-app1.protocol=TCP \
  --set ports.memcache-app2.port=11213 \
  --set ports.memcache-app2.exposedPort=11213 \
[...]

PR #801, opened 2023-03-01
🚨 Fail when enabling PSP on Kubernetes v1.25+

mloiseleur:
### What does this PR do?

It produces an error when enabling PodSecurityPolicy on Kubernetes 1.25+.

### Motivation

It has been [removed](https://kubernetes.io/docs/concepts/security/pod-security-policy/) by Kubernetes.
Some users missed it and does not understand why it fails to install, see https://github.com/traefik/traefik-helm-chart/issues/800

### More

### DIFF
--- a/traefik/templates/rbac/podsecuritypolicy.yaml
+++ b/traefik/templates/rbac/podsecuritypolicy.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.podSecurityPolicy.enabled }}
 {{- if semverCompare ">=1.25.0-0" .Capabilities.KubeVersion.Version }}
-  {{- fail "ERROR: PodSecurityPolicy has been deprecated on Kubernetes v1.25+" }}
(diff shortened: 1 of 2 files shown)


traefik/templates/rbac/podsecuritypolicy.yaml
```diff
@@ -1,4 +1,7 @@
 {{- if .Values.podSecurityPolicy.enabled }}
+{{- if semverCompare ">=1.25.0-0" .Capabilities.KubeVersion.Version }}
+  {{- fail "ERROR: PodSecurityPolicy has been deprecated on Kubernetes v1.25+" }}
```
charlie-haley, line 3:
```suggestion
  {{- fail "ERROR: PodSecurityPolicy has been removed in Kubernetes v1.25+" }}
```

charlie-haley (approved):
LGTM

PR #802, opened 2023-03-06
Allow overriding of image registry and add explicit default

jnoordsij:
### What does this PR do?

Allow setting `image.registry` value to specify the registy the image should be pulled from.

### Motivation

Checking the full list of images present in my cluster, I noticed a few of them not having a fully quantified image name, including this image. While technically `docker.io` does not require explicit specification, I think it's slightly better to make it explicit, and from a quick glance it might also be a tiny bit more secure and stable.

### More


project0:
I am not sure i understand the change, what makes you stop setting the image to `docker.io/traefik` (In fact i believe some people just do it, so it will result into a breaking change).

It feels unnecessarily to me compared to just setting one simple value containing the full repo name.

How about just setting docker.io/traefik as default. @mloiseleur wdyt?

mloiseleur:
> How about just setting docker.io/traefik as default. @mloiseleur wdyt?

Yes I agree. One value seems enough and is [documented as a single field](https://docs.docker.com/engine/reference/run/#imagetag) in docker run documentation.

jnoordsij:
I think the main pros of using two separate values are:
1. easier to use and maintain: if you use e.g. a local (offline) mirror to pull images from, you can set this specific variable, without having to look up the exact image name and without the need to change anything in your own configuration if the image name changes upstream
2. semantically I think it's a little more clear

I've also seen this pattern used commonly in quite a few other Helm charts; though there are plenty that just use the image name to set this as well.

On the other side, I don't have a strong preference either, so just updating the default value and not adding a new one sounds fine as well. That also is a bit more backwards compatible.

mloiseleur:
Since we have many users on this chart, would you please rework this PR in order to be (more) backwards compatible ?

jnoordsij:
What I mean with "backwards compatible" in this case, is that this proposed change is incompatible for those who are already overriding the `image.name` value. I think this group will be very limited, as I think the main usecase where someone would do that is if they would actually want to override the registry, which this change allows in an easier way. 

So it basically boils down to: do you want to have an easier way to do this in the future, but requires manual intervention now once (though *only* for those already overriding the image name), or do you want to keep the old behavior and just update the default value to be `docker.io/traefik` instead of `traefik`.

mloiseleur:
> _Edit:_ additional note: those who are currently using this chart and upgrade without value overrides (or with the `--reuse-values` flag), which I think will be the largest group of users, will not receive an updated image name value if the second option is picked, since any updated upstream chart value is ignored when upgrading through Helm. So if you want more users to benefit from the fully qualified image name, going with the proposed change as is might be preferable.

@jnoordsij I was not expecting that new default values are not taken into account when upgrading with `reuse-values`. I will double check and think about it. I have planned to work on this on next Friday.

jnoordsij:
Thanks for checking this @mloiseleur! I've been confused by this behavior of Helm as well; https://medium.com/@kcatstack/understand-helm-upgrade-flags-reset-values-reuse-values-6e58ac8f127e explains it quite nicely IMO.

mloiseleur (approved):

project0 (approved):

viceice (changes requested):
this is a breaking change as patch release 😭

mloiseleur:
@viceice ah damn it ! You're right, we didn't precise it in the title and this patch release was for 2.9.9 version of Traefik. I guess we'll need to make another release.

viceice:
@mloiseleur you should publish a fix release for v21 to revert the breaking changes

PR #815, opened 2023-03-31
add additional labels to tlsoption

eddykaya:
### What does this PR do?
When using custom TLSOptions together with labelSelectors, the helm chart was unable to generate a matching label when using tlsOption key.
This PR enables users to add custom labels to any custom TLSOption defined so that you have fine grained control over which TLSOption a traefik intance should pick up.

### Motivation
A possible workaround was to use extraObject, however it seems a little bit unintuitive for me as a user tries to use built in TLSOption first.

### More

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -654,12 +654,15 @@ ports:
 
 # TLS Options are created as TLSOption CRDs
 # https://doc.traefik.io/traefik/https/tls/#tls-options
+# You can add additional labels to a tlsOption
 # Example:
 # tlsOptions:
 #   default:
+#     labels: {}
 #     sniStrict: true
 #     preferServerCipherSuites: true
-#   foobar:
+#   customOptions:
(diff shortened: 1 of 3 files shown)


traefik/values.yaml
```diff
@@ -654,12 +654,15 @@ ports:
 
 # TLS Options are created as TLSOption CRDs
 # https://doc.traefik.io/traefik/https/tls/#tls-options
+# You can add additional labels to a tlsOption
```
mloiseleur, line 657:
```suggestion
# When using `labelSelector`, you'll need to set labels on tlsOption accordingly.
```

mloiseleur (approved):
Thanks for this PR ! I only have a small suggestion on documentation.

darkweaver87 (approved):
LGTM

mloiseleur (approved):

project0 (approved):

Issue #165, opened 2020-04-27
Traefik can't create acme.json (read-only filesystem, permission denied)


Jawastew:
With the new securityContext from #155 I can no longer mount a pv with the default settings.
```
kubectl logs -n kube-system service/traefik
...
time="2020-04-27T18:43:23Z" level=error msg="The ACME resolver \"letsencrypt-test\" is skipped from the resolvers list because: unable to get ACME account: open acme.json: read-only file system"
time="2020-04-27T18:43:23Z" level=error msg="The ACME resolver \"le\" is skipped from the resolvers list because: unable to get ACME account: open /data/acme.json: permission denied"
```

When I try to access my mounted /data volume, it is owned by root by default and I can't read or create files. Even after I chown the directory on the host machine where the volume is mounted I still can't write to the directory because of the ```securityContext.readOnlyRootFilesystem = true``` default.

my helm values:
```
...
persistence:
  enabled: true
  storageClass: export-nfs
  accessMode: ReadWriteMany
  subPath: "traefik"
  size: 1Gi
...
```

When I add:
```
securityContext:
  readOnlyRootFilesystem: False
```
I am able to get rid of the read-only filesystem warning, but I am still getting 'permission denied'.

To get everything running again under the default root user you need the following chart values:
```
securityContext:
  readOnlyRootFilesystem: False
  runAsGroup: 0
  runAsUser: 0
  runAsNonRoot: false
```

I would propose to change the security context back to a root user as default, since this will not introduce any breaking changes when updating from an earlier version of the helmchart where the acme.json would be owned by root by default.

SimonTheLeg:
Can you share your whole values.yml?

With these settings I am still stuck one step earlier (same as #164) at:

```
time="2020-04-29T09:56:43Z" level=error msg="The ACME resolver \"le\" is skipped from the resolvers list because: unable to get ACME account: permissions 660 for /data/acme.json are too open, please use 600"
```

Jawastew:
values.yaml
```yaml
image.tag: "2.2.0"
service:
persistence:
  enabled: true
  storageClass: export-nfs
  accessMode: ReadWriteMany
  subPath: "traefik"
  size: 1Gi
securityContext:
  readOnlyRootFilesystem: false
  runAsGroup: 0
  runAsUser: 0
  runAsNonRoot: false
additionalArguments:
  - "--certificatesresolvers.le.acme.storage=/data/acme.json"
[...]

PR #827, opened 2023-04-15
Allow readOnly to be specified on volume mounts

Justice4Joffrey:
### What does this PR do?

Passes through the `readOnly` flag from the `volumes[]` section. The current value is hardcoded to `true`.

### Motivation

TLDR: I want to provide my own volume and have letsencrypt manage an `acme.json` in it.

I've spent a fair chunk of time weeding through file-system permissions issues. Seems like others have been [struggling too](https://github.com/traefik/traefik-helm-chart/issues/165). The `persistence` configuration option didn't work for me and I've forgotten why (sorry!).

It seems reasonable that volume mounts shouldn't only be `readOnly`, but perhaps there are other ideas how one should approach this. If that's the case, maybe a troubleshooting readme section would suffice.

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -123,7 +123,7 @@
           {{- range .Values.volumes }}
           - name: {{ tpl (.name) $root | replace "." "-" }}
             mountPath: {{ .mountPath }}
-            readOnly: true
+            readOnly: {{ .readOnly }}
           {{- end }}
           {{- if .Values.experimental.plugins.enabled }}
           - name: plugins


mloiseleur:
Hello @Justice4Joffrey,

Thanks for your interest in Traefik. 

This PR needs a default value and some tests before being reviewed.

Btw, did you tried to mount acme.json following [this example](https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#use-traefik-native-lets-encrypt-integration-without-cert-manager) ?

mloiseleur:
I'm closing this PR since there is a working example and no details on why/how it can fail.
Feel free to re-open it if you have more info / a more complete PR.

PR #837, opened 2023-04-25
feat(helm): added values README via helm-docs cli

justmike1:
### What does this PR do?

Added README file to possess all values that traefik's helm chart reads

### Motivation

I checked the full value path I needed to write and reading from the yaml is difficult, for example `ports` is a long tree.

### More

### DIFF
--- a/traefik/README.md
+++ b/traefik/README.md
@@ -1,0 +1,169 @@
+# traefik
+
+![Version: 22.3.0](https://img.shields.io/badge/Version-22.3.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v2.10.0](https://img.shields.io/badge/AppVersion-v2.10.0-informational?style=flat-square)
+
+A Traefik based Kubernetes ingress controller
+
+**Homepage:** <https://traefik.io/>
+
+## Maintainers
+
+| Name | Email | Url |
+| ---- | ------ | --- |
+| emilevauge | <emile@vauge.com> |  |
+| dtomcej | <daniel.tomcej@gmail.com> |  |
+| ldez | <ldez@traefik.io> |  |
+| mloiseleur | <michel.loiseleur@traefik.io> |  |
+| charlie-haley | <charlie.haley@traefik.io> |  |
+
+## Source Code
+
+* <https://github.com/traefik/traefik>
+* <https://github.com/traefik/traefik-helm-chart>
+
+## Requirements
+
+Kubernetes: `>=1.16.0-0`
+
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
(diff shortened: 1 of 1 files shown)


mloiseleur:
Hello @justmike1, 

Thanks for your interests in Traefik and for this PR !
This can definitely help when using this chart.

I have some comments:
1. Wdyt about adding a Makefile target to re-generate this file?
2. There are a lot of parameters, so maybe we should keep this doc inside a specific file (like `VALUES.md`), in order to keep readability
3. README.md is copied to `traefik/` in order to be displayed in [artifacthub page](https://artifacthub.io/packages/helm/traefik/traefik/). See [here](https://github.com/traefik/traefik-helm-chart/blob/master/.github/workflows/release.yml#L30).

justmike1:
@mloiseleur 
1. Added the Makefile target
2. You can't change the name of the file the `helm-docs` output. so I added renaming via Makefile
3. Changing to `VALUES.md` mitigates the issue, though in the future, when we will add also proper descriptions and default values via `helm-docs` supported comments and have a descriptive file, we will need to think a way to merge them sensibly and not just one on top of the other.

I use `traefik` for a year already and I don't see it change any time soon, so I will improve it with time.

justmike1:
I also added a test step in order for the file to stay updated upon PRs

mloiseleur:
That's better. I tried it by adding the needed `--` into values.yaml for two first section and it looks good. 
I guess we'll need to find how to properly display examples provided directly into values.yaml.

justmike1:
@mloiseleur anything more required to push this PR? I don't like big PRs, so I would suggest steps:
1. merging this PR 
2. PR for refactoring values.yaml comments to `helm-docs`
3. PR for main README to have those values as well with keeping it's structure, like injecting them before `contributions` header, in order to be seen also in artifacthub

mloiseleur:
I don't have a strong opinion of PR size.
I do have a strong opinion that we should not merge an empty VALUES.md.
I agree with you that it could be merged within the README.

=> For this PR, I think it's missing two things:
1. Add the needed `--` in values.yaml, to have a base on VALUES.md
2. In [Guidelines](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/Guidelines.md#non-critical-feature-example), non critical features are commented. How could those features be documented with this system ?

justmike1:
@mloiseleur Sorry for repeating commits, `VALUES.md` is full now, pending your next review.
Currently, I don't see a way in `helm-docs` to document commented out parameters, maybe in future versions, so I believe keeping it as is, and maybe we can layer the documentation, and then make a full one that will be structured from the files in CD flow.

mloiseleur:
I should be able to review it on next friday. I'll keep you informed.

mloiseleur (approved):

justmike1:
Thank you guys! @darkweaver87 @mloiseleur

Issue #826, opened 2023-04-15
Support setting `loadBalancerClass` in Service spec


z0rc:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class

I'd like to set this field in Traefik's Service of type LoadBalancer, when using with NLB managed by AWS LBC, see https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/guide/service/nlb/#configuration

mloiseleur:
Hello @z0rc,

Thanks for your suggestion, it makes a lot of sense.  Unfortunately, this would not make it to our roadmap for a while as we are focused elsewhere.  If you or another community member would like to build it, let us know and we will work with you to make sure you have all the information needed so that it can be merged.  

Don’t forget to check out the [contributor docs] (https://github.com/traefik/traefik-helm-chart/blob/master/CONTRIBUTING.md) and to link the PR to this issue.

project0:
Will be available in the next helm release ;-)

PR #843, opened 2023-04-28
feat: allow specifying service loadBalancerClass

project0:
### What does this PR do?

implements #826

### More


mloiseleur (approved):
LGTM

z0rc:
This change is confusing.

It appears it wasn't needed in the first place, as Service's spec.loadBalancerClass was possible to set using `service.spec.loadBalancerClass` value. Added value `service.loadBalancerClass` (as implied by commented out entry in values.yaml) doesn't have effect when set. I haven't debug this further, but seems it's conflicting with `service.spec` map somehow.

PR #846, opened 2023-05-04
Fix Let's Encrypt example

mhmxs:
Without the `volumeMount` the init fails, because `/data` doesn't exists.

### More


mloiseleur (approved):

mloiseleur:
You're right @mhmxs 
Thanks for this PR !

PR #858, opened 2023-06-02
Add appProtocol field

baarde:
### What does this PR do?

This PR enables the user to specify an optional value for the [`.spec.ports[*].appProtocol` field](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol).

### Motivation

The `appProtocol` field may be set on a Service as a hint for implementations to offer richer behavior for protocols that they understand. For example, this field is used by the GKE Gateway controller to [enable TLS between the load balancer and the backend](https://cloud.google.com/kubernetes-engine/docs/how-to/secure-gateway#load-balancer-tls).

**Use case:** I want to deploy a Google External HTTP(S) load balancer (so I can use Cloud Armor) and then forward the traffic over HTTPS (so it remains encrypted) to Traefik (so I can use Traefik).

### Compatibility

`appProtocol` is available as an optional feature in Kubernetes 1.18, enabled by default in Kubernetes 1.19, and general available since Kubernetes 1.20.

Setting a value for `ports.*.appProtocol` breaks the compatibility with older version of Kubernetes. Therefore, no default value is set.

### More


mloiseleur (commented):
Thanks for this PR, @baarde !
LGTM

mloiseleur (approved):

mloiseleur:
Current [example on GCP](https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#install-on-gcp) is not covering this use case.
Do you think you can add a working example with this setup ?

darkweaver87 (approved):
LGTM

mloiseleur (approved):

baarde:
Thank you @mloiseleur.

> Current [example on GCP](https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#install-on-gcp) is not covering this use case.
> Do you think you can add a working example with this setup ?

Here is an example of my setup:

```yaml
ports:
  websecure:
    appProtocol: HTTPS # Hint for Google L7 load balancer
service:
  type: NodePort
extraObjects:
- apiVersion: gateway.networking.k8s.io/v1beta1
  kind: Gateway
  metadata:
    name: traefik
    annotations:
      networking.gke.io/certmap: "myCertificateMap"
  spec:
    gatewayClassName: gke-l7-global-external-managed
    addresses:
    - type: NamedAddress
      name: "myGlobalIPName"
    listeners:
    - name: https
      protocol: HTTPS
      port: 443
- apiVersion: gateway.networking.k8s.io/v1beta1
  kind: HTTPRoute
  metadata:
    name: traefik
  spec:
    parentRefs:
    - kind: Gateway
      name: traefik
    rules:
    - backendRefs:
      - name: traefik
        port: 443
- apiVersion: networking.gke.io/v1
  kind: HealthCheckPolicy
  metadata:
    name: traefik
  spec:
    default:
      config:
        type: HTTP
        httpHealthCheck:
          port: 9000
          requestPath: /ping
  targetRef:
    group: ""
    kind: Service
    name: traefik
```

It's very verbose (and I've kept out HTTP-to-HTTPS redirection). Unfortunately, I don't think I can make it more compact: Google External HTTP(S) load balancer insists on performing their own health check instead of relying on the Pod's readiness.

If you think that would be helpful, I'll create a new PR to add it.

**Note:** To anyone using the GKE Gateway controller, make sure you haven't enabled IPv6 on the Service. It's not supported yet. I lost several hours failing to understand why the NEG controller wasn't attaching the Pod's IPs to my network endpoint group.

mloiseleur:
\o Thanks for this.
Yes, I think that would be helpful !

PR #863, opened 2023-06-13
feat: add option to opt-out API credential automounting.

pgebert:
### What does this PR do?

This PR adds an option to opt-out API credential automounting. The default behaviour stays the same.

See [kubernetes - opt-out-of-api-credential-automounting](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#opt-out-of-api-credential-automounting)

### Motivation

This opt-out option would help us to meet our kubernetes security policies.

### More

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -149,7 +149,7 @@ Kubernetes: `>=1.16.0-0`
 | service.single | bool | `true` |  |
 | service.spec | object | `{}` | Cannot contain type, selector or ports entries. |
 | service.type | string | `"LoadBalancer"` |  |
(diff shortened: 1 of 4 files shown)


mloiseleur (approved):

mloiseleur:
The PR is valid but I'm wondering :thinking:, when setting it to false, it will be entirely disabled, on both `ServiceAccount` and `Deployment`.
How Traefik Proxy will be able to work with Kube API Server to connect `Ingress` or `IngressRoute` with `Service` ?

pgebert:
That's a valid point. The scenarios where this makes sense are probably quite rare. 🙄 And after thinking about it again, I'm not so sure anymore whether we really need this feature. It would be fine for me to close this ticket.

Issue #855, opened 2023-05-26
Not customisable IngressClass name


kochetov-dmitrij:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

23.0.1

### What version of Traefik are you using?

v2.10.1

### What did you do?

I expected that `providers.kubernetesIngress.ingressClass` is used to set the IngressClass name.

### What did you see instead?

But apparently the release name is used instead:
https://github.com/traefik/traefik-helm-chart/blob/1f29975682db2da0737e85182cda9592e3dd707c/traefik/templates/ingressclass.yaml#L12-L18

The value is referenced only here:
https://github.com/traefik/traefik-helm-chart/blob/1f29975682db2da0737e85182cda9592e3dd707c/traefik/templates/_podtemplate.tpl#L505-L506

Is this a bug?

### What is your environment & configuration?

YAML

mloiseleur:
Hello @kochetov-dmitrij,

Thanks for this report ! It's weird it has not been detected before, but it seems you are right.
Do you think you can submit a PR to fix it ?

mloiseleur:
It's now possible in v23.2.0

PR #864, opened 2023-06-14
fix: ingressclass name should be customizable

kupnu4x:
### What does this PR do?

Fixes issue https://github.com/traefik/traefik-helm-chart/issues/855

### More

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -64,7 +64,7 @@ Kubernetes: `>=1.16.0-0`
 | image.registry | string | `"docker.io"` | Traefik image host registry |
 | image.repository | string | `"traefik"` | Traefik image repository |
 | image.tag | string | `""` | defaults to appVersion |
-| ingressClass | object | `{"enabled":true,"isDefaultClass":true}` | Create a default IngressClass for Traefik |
+| ingressClass | object | `{"enabled":true,"isDefaultClass":true,"name":"traefik"}` | Create a default IngressClass for Traefik |
 | ingressRoute.dashboard.annotations | object | `{}` | Additional ingressRoute annotations (e.g. for kubernetes.io/ingress.class) |
 | ingressRoute.dashboard.enabled | bool | `true` | Create an IngressRoute for the dashboard |
 | ingressRoute.dashboard.entryPoints | list | `["traefik"]` | Specify the allowed entrypoints to use for the dashboard ingress route, (e.g. traefik, web, websecure). By default, it's using traefik entrypoint, which is not exposed. /!\ Do not expose your dashboard without any protection over the internet /!\ |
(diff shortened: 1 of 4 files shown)


mloiseleur:
@kupnu4x: Thanks for this PR 

Why did you add a new value instead of using `.Values.providers.kubernetesIngress.ingressClass` ?

traefik/templates/ingressclass.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
     ingressclass.kubernetes.io/is-default-class: {{ .Values.ingressClass.isDefaultClass | quote }}
   labels:
     {{- include "traefik.labels" . | nindent 4 }}
-  name: {{ template "traefik.fullname" . }}
+  name: {{ .Values.ingressClass.name }}
```
z0rc, line 18: This will break backward compatibility for users who deploy chart under non-default release name, which is common use case and should be taken into account here.

z0rc (changes requested):

traefik/templates/ingressclass.yaml
darkweaver87, line 18:
```suggestion
  name: {{ .Values.ingressClass.name | default (include "traefik.fullname" .) }}
```

traefik/values.yaml
```diff
@@ -105,6 +105,7 @@ podDisruptionBudget:
 ingressClass:
   enabled: true
   isDefaultClass: true
+  name: traefik
```
darkweaver87, line 108:
```suggestion
  # name: my-custom-class
```

traefik/VALUES.md
```diff
@@ -64,7 +64,7 @@ Kubernetes: `>=1.16.0-0`
 | image.registry | string | `"docker.io"` | Traefik image host registry |
 | image.repository | string | `"traefik"` | Traefik image repository |
 | image.tag | string | `""` | defaults to appVersion |
-| ingressClass | object | `{"enabled":true,"isDefaultClass":true}` | Create a default IngressClass for Traefik |
+| ingressClass | object | `{"enabled":true,"isDefaultClass":true,"name":"traefik"}` | Create a default IngressClass for Traefik |
```
darkweaver87, line 67:
```suggestion
| ingressClass | object | `{"enabled":true,"isDefaultClass":true}` | Create a default IngressClass for Traefik |
```

darkweaver87 (commented):
WDYT about handling it that way to avoid breaking changes ?

mloiseleur (approved):

traefik/templates/ingressclass.yaml
mloiseleur: now, with commit from @darkweaver87 , it won't break.

darkweaver87 (approved):
LGTM

PR #879, opened 2023-06-28
Add support for running the dashboard as ingress

davhdavh:
### What does this PR do?

Add support for running the dashboard as ingress
Add documentation for why this may be desired
Add documentation for how to get external ip and loadbalancer ip working in own vm/hardware

### Motivation

Add some extra documentation for how to get started with traefik on a bare k8s cluster.
Easier support for ingress access to the dashboard
Implicit documentation of how to use Ingress with attributes (something the traefik docs are quite bad at showing)

### More

### DIFF
--- a/traefik/Changelog.md
+++ b/traefik/Changelog.md
@@ -1,8 +1,77 @@
 # Change Log
 
+## 24.0.0  ![AppVersion: v2.10.4](https://img.shields.io/static/v1?label=AppVersion&message=v2.10.4&color=success&logo=) ![Kubernetes: >=1.16.0-0](https://img.shields.io/static/v1?label=Kubernetes&message=%3E%3D1.16.0-0&color=informational&logo=kubernetes) ![Helm: v3](https://img.shields.io/static/v1?label=Helm&message=v3&color=informational&logo=helm)
+
+**Release date:** 2023-08-10
+
+* chore(release): 🚀 publish v24.0.0
+* fix: http3 support broken when advertisedPort set
+* fix: tracing.opentelemetry.tls is optional for all values
+* chore(deps): update docker.io/helmunittest/helm-unittest docker tag to v3.12.2
+* chore(tests): 🔧 fix typo on tracing test
+* fix: 💥 BREAKING CHANGE on healthchecks and traefik port
+* feat: multi namespace RBAC manifests
+
+### Default value changes
+
+```diff
+diff --git a/traefik/values.yaml b/traefik/values.yaml
+index 947ba56..aeec85c 100644
+--- a/traefik/values.yaml
++++ b/traefik/values.yaml
+@@ -28,6 +28,13 @@ deployment:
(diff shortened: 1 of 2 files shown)


mloiseleur (changes requested):
Thanks for this other PR @davhdavh 
Interesting. I have two main comments:
1. We have moved all example to EXAMPLES.md, wdyt about add your example in this file ?
2. `api.insecure` is not linked to port. So this should be put elsewhere.

davhdavh:
> api.insecure is not linked to port. So this should be put elsewhere.

https://doc.traefik.io/traefik/v3.0/operations/api/#insecure
Well, port and entryPoint is used intermixed in the helm config, but this binds specifically to the `traefik` entryPoint, and that is configured in ports.traefik. So yes I think it very much belongs here

davhdavh:
> We have moved all example to EXAMPLES.md, wdyt about add your example in this file ?

Primarily because the main place to read the documentation does not have the examples accessible:
https://artifacthub.io/packages/helm/traefik/traefik

traefik/values.yaml
```diff
@@ -541,6 +541,71 @@ envFrom: []
 ports:
   traefik:
     port: 9000
+    # By default do not host the dashboard on this port
+    # instead, use ingressRoute.dashboard.enabled (this uses the dashboard@internal service, instead of a seperate port)
+    # but you might need it as an ingress rather than ingressRoute, and this would allow you to do so
```
darkweaver87, line 546: Well the ingress method works but is a bit hacky.

traefik/values.yaml
```diff
@@ -541,6 +541,71 @@ envFrom: []
...
+    # ---
+    # apiVersion: traefik.io/v1alpha1
+    # kind: Middleware
+    # metadata:
+    #   name: traefik-dashboard-auth
+    #   namespace: sameAsTraefik
+    # spec:
+    #   basicAuth:
+    #     secret: dashboard-auth          
+    # The creation of the secret resource "dashboard-auth" is left to the reader (it needs username and password fields)
```
darkweaver87, line 606:
```suggestion
    # The creation of the secret resource "dashboard-auth" is left to the reader (it needs username and password fields). You may prefer better authentication method like OIDC.
```

darkweaver87:
> > api.insecure is not linked to port. So this should be put elsewhere.
> 
> https://doc.traefik.io/traefik/v3.0/operations/api/#insecure Well, port and entryPoint is used intermixed in the helm config, but this binds specifically to the `traefik` entryPoint, and that is configured in ports.traefik. So yes I think it very much belongs here

Well yes and no ;-) 

I would tend to agree with @mloiseleur. `api.insecure` is not in `entrypoints` section of traefik configuration and `ports` tries to stick as much as possible at it  whereas `api.insecure` deals with `routers`.

mloiseleur:
@davhdavh Will you update this PR or should we close it ?

PR #889, opened 2023-07-13
release: rocket publish v23.1.1

darkweaver87:
### Motivation

Master branch includes already 11 bug/fixes.


kapilag123:
Can we get this moving??

davhdavh:
should probably be a v24 instead of v23.1.1 with the beta3 support

mloiseleur (approved):

mloiseleur:
> should probably be a v24 instead of v23.1.1 with the beta3 support

There is no breaking change, those with `experimental.v3.enabled` in values won't have any side effect. It was already needed to set the `image.tag`.

Issue #890, opened 2023-07-14
It's impossible to disable a port defined in the default config by setting it to null


marekvospel:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

23.1.0

### What version of Traefik are you using?

2.10.1

### What did you do?

Traefik helm chart creates 4 entrypoints by default, web, websecure, traefik and prometheus. I only need web and websecure, so I've been trying to disable traefik and prometheus ports.

I've found https://github.com/traefik/traefik-helm-chart/issues/220, which said it is possible to use null, which had correct behavior, but fails during installation with an error in trying to expose the ports through kubernetes

### What did you see instead?

```
Error: INSTALLATION FAILED: template: traefik/templates/_podtemplate.tpl:63:36: executing "traefik.podTemplate" at <.Values.ports.traefik.port>: nil pointer evaluating interface {}.port
```

### What is your environment & configuration?

```
ports:
  web:
    port: 80
  websecure:
    port: 443
    http3:
      enabled: true
      advertisedPort: 443
  traefik: null
  prometheus: null
```

mloiseleur:
It's now possible in v23.2.0.

mantoine96:
Hey @mloiseleur 

It seems this is still an issue, at least for disabling `web`:

```
template: traefik/templates/_podtemplate.tpl:60:40: executing "traefik.podTemplate" at <.Values.ports.web.port>: nil pointer evaluating interface {}.port
```

This is because now `ports.web.port` is taken as the default healthcheck port before we evaluate whether the Traefik endpoint is available: https://github.com/traefik/traefik-helm-chart/pull/891/files#diff-c64349ddaeb8064687b0839b8b93e77fb6733c2dae67b8c0fface1cac0652fccR60-R65

I'm not sure what's the expected way forward here. Up until now we had always had `web` and `websecure` disabled as per https://github.com/traefik/traefik-helm-chart/issues/220 (using our own endpoints).

This is what we had and was working until v23.2.0:
```
                ports:
                  traefik:
                    port: 8082
                  admin:
                    port: 8444
                    tls:
                      enabled: true
                  http-internal:
                    port: 8443
                    tls:
                      enabled: true
                    expose: true
[...]

PR #891, opened 2023-07-21
fix: 🐛 traefik or metrics port can be disabled

mloiseleur:
### What does this PR do?

Solve error encountered when trying to disable `traefik` or `metrics` port.
/!\ Note that when you disable `traefik` ports, liveness probe will be directly on web port.

### Motivation

See #890.

### More


darkweaver87 (commented):
You may wanna test if ports are really disabled when it's disabled through values

darkweaver87 (approved):
LGTM

PR #897, opened 2023-07-27
Multi namespace RBAC manifests

jnoordsij:
### What does this PR do?

When adding multiple namespaces to watch by the Traefik and using namespaced RBAC, this PR ensures Roles and RoleBindings are added for all relevant namespaces.

### Motivation

Currently, only a Role + RoleBinding is added for the chart namespaces. When watching more namespaces, manual adding (and maintaining) of additional resources is required, as by default the ServiceAccount has no access to the resources.

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -108,10 +108,10 @@ Users can provide an override for an explicit service they want bound via `.Valu
 Construct a comma-separated list of whitelisted namespaces
 */}}
 {{- define "providers.kubernetesIngress.namespaces" -}}
-{{- default .Release.Namespace (join "," .Values.providers.kubernetesIngress.namespaces) }}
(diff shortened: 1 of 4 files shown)


mloiseleur (approved):
Interesting PR
Many thanks @jnoordsij !

mloiseleur (changes requested):
Would you please add two tests, one with namespaces only on `kubernetesIngress` and an other one only on `kubernetesCRD` ?

jnoordsij:
> Would you please add two tests, one with namespaces only on `kubernetesIngress` and an other one with namespaces only on `kubernetesCRD` ?

Sure! I added two additional tests in the newest commit.

mloiseleur (approved):

darkweaver87 (approved):
LGTM

PR #903, opened 2023-08-04
fix: http3 support broken when advertisedPort set

davhdavh:
### What does this PR do?

when http3 enabled and http3.advertisedPort set, it would not actually set the argument to enable http3

### Motivation

I broke it


davhdavh:
temporary workaround:

```yaml  
  additionalArguments:
    - --entrypoints.websecure.http3
```

mloiseleur:
@davhdavh Would you please add a test ? So this won't break again ?

mloiseleur (approved):
Thanks, it looks good.

PR #907, opened 2023-08-11
fix: allow updateStrategy.rollingUpdate.maxUnavailable to be passed in as an int or string

### LINKED ISSUE #906: Helm parameter updateStrategy.rollingUpdate.maxUnavailable type checking is too strict
### Welcome!

v24.0.0

v2.10.4

### What did you do?

I Installed the traefik helm chart via ArgoCD's Application CRD and attempted to override `updateStrategy.rollingUpdate.maxUnavailable` with the value of `1` through the `source.helm.parameters` portion of the application spec.

### What did you see instead?

Due to the way ArgoCD's app of apps functions, it forces the value of any helm parameter overrides to come in as a string (at least in my case). 

The traefik helm chart performs the following  check, assuming that the value is an integer and trips up with a failure.

https://github.com/traefik/traefik-helm-chart/blob/2dd3b1848ff88f053a7cb01fdc997ab47bc3afd3/traefik/templates/daemonset.yaml#L13-L15

Is there any particular reason the chart is enforcing int over string in this case?

According to the Daemonset spec, it can be an int or a string and also supports being set as a percentage.

```
maxUnavailable	<IntOrString>
    The maximum number of DaemonSet pods that can be unavailable during the
    update. Value can be an absolute number (ex: 5) or a percentage of total
    number of DaemonSet pods at the start of the update (ex: 10%). Absolute
    number is calculated from percentage by rounding up. This cannot be 0 if
    MaxSurge is 0 Default value is 1. Example: when this is set to 30%, at most
    30% of the total number of nodes that should be running the daemon pod (i.e.
    status.desiredNumberScheduled) can have their pods stopped for an update at
    any given time. The update starts by stopping at most 30% of those DaemonSet
    pods and then brings up new DaemonSet pods in their place. Once the new pods
    are available, it then proceeds onto other DaemonSet pods, thus ensuring
    that at least 70% of original number of DaemonSet pods are available at all
    times during the update.
```

ArgoCD v2.8.0+804d4b8
K8S 1.24.9

### Additional Information

```shell
Somewhat related, and perhaps may make my problem go away, it would perhaps be more efficient to short circuit the if conditional like the following because the sanity check is only really necessary if `hostNetwork` is enabled:


{{- if and (.Values.hostNetwork) (lt .Values.updateStrategy.rollingUpdate.maxUnavailable 1.0)  }}
```
```

hrivera-ntap:
…string

### What does this PR do?

Fixes  #906

### Motivation

Allow tools like ArgoCD to be able to override `updateStrategy.rollingUpdate.maxUnavailable` with strings (or percentage values)

### More

- [ X ] Yes, I updated the tests accordingly
- [ X ] Yes, I ran `make test` and all the tests passed


mloiseleur (approved):

mloiseleur:
It's complete and well tested.
Thanks for this PR, @hrivera-ntap !

Issue #911, opened 2023-08-17
Traefik V3 new or modified Middlewares not supported


gcaracuel:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

24.0.0

### What version of Traefik are you using?

v3.0

### What did you do?

Migrated ipWhitelist Middleware to ipAllowList but `ipAllowList` is not supported at CRDs.

Similar issues will happen to new Middlewares like `grpcWeb `

### What did you see instead?

When using old type ipWhitelist controller fails, when trying to use ipAllowList CRDs do not allow it.

### What is your environment & configuration?

Not needed

gcaracuel:
Hello @mloiseleur 

Thanks for your reply. I am afraid I won't be able to apply that branch as I am using GitOps approach and this Helm Chart so it would be overwritten. So I could just discard usage of V3 until all this is king of sorted out. 
A new groupversion and keep it aligned in this Helm Chart would have solved this and will add users to V3 beta but I could understand why is not done.

Gonna close this and wait for a stable V3

sjmiller609:
To clarify for anyone else that ends up here - Traefik version 2 and Traefik version 3 have different API specifications for the API version `traefik.io/v1alpha1`. More info here https://github.com/traefik/traefik/pull/9600#issuecomment-1753313008

PR #912, opened 2023-08-17
Add Middlewares ipAllowList and grpcWeb to V3 CRDs

gcaracuel:
### What does this PR do?

Adds new V3 introduced Middlewares to CRDs:
* `grpcWeb`
* `ipAllowList` which is a rename of current `ipWhiteList`

### Motivation

V3 changed `ipWhiteList` Middleware by `ipAllowList`. Not being present on CRD blocks migration for `ipWhiteList` users.

### More

### DIFF
--- a/traefik/crds/traefik.io_middlewares.yaml
+++ b/traefik/crds/traefik.io_middlewares.yaml
@@ -632,6 +632,45 @@ spec:
                       type: string
                     type: array
                 type: object
+              ipAllowList:
+                description: 'IPAllowList holds the IP allowlist middleware configuration. It was introduced by V3.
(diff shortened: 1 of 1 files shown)


mloiseleur:
Hello @gcaracuel,

Thanks for this PR. 
I understand why you did it but I don't see how we can accept it.
With this PR, it means that users of current stable version of Traefik Proxy, v2.10, will have new (non-working) fields on their version. 
Considering that (a) Helm is not allowing us to manage CRDs upgrade (b) the CRDs on Traefik Proxy v3 will still evolve, I guess it's better to stick with stable CRDs from stable version of Traefik Proxy. 

See my comment on #911 for more info.

PR #913, opened 2023-08-22
fix: 🐛 remove old CRDs using traefik.containo.us

### LINKED ISSUE #895: Allow an easy way to disable legacy CRDs in Helm chart
### Welcome!

The docs says that legacy API group (traefik.containo.us) is deprecated and will be removed in v3.
Would it be possible, until v3 is live, to have a true/false switch in the official Helm chart Values to be able to skip applying the legacy CRDs? 

I have two motivations for such request:
- should any of my old helm charts still accidentally use legacy API group, it should fail to deploy so that I immediately know that it should be upgraded
- I'll get rid of duplicated resource aliases with the same name

mloiseleur:
### What does this PR do?

Remove old CRDs using traefik.containo.us.

### Motivation

Fixes #895


HeyHugo:
Bumped into this and couldn't understand why I couldn't list traefik CRD resources after I had upgraded.

To anyone having the same problem delete the legacy crds
```
kubectl get crd | grep 'traefik.containo.us' | awk '{ print $1 }' | xargs kubectl delete crd
```

Then get a crd resource explicitly using the full crd name like this
`kubectl get ingressroute.traefik.io --all-namespaces`

For me it seems some client state wasn't updated until I did. After that using just `kubectl get ingressroute` worked

darkweaver87 (approved):
LGTM

jnoordsij (approved):

davidpanic:
This was reverted in bf75a5df Can anyone provide any information as to why?

mloiseleur:
It's explained in the PR:
* #944

PR #914, opened 2023-08-23
feature ✨ (traefik): added healthcheck ingressRoute

muandane:
### What does this PR do?

This PR introduces an IngressRoute for health checks, addressing the specific requirements of running in a constrained environment. The proposed changes are in alignment with the project guidelines and seamlessly integrate with the existing codebase. By incorporating this IngressRoute.

### Motivation

I'm Deploying private clusters that are behind an Application gateway that has backend probes to check the health of the LoadBalancer (internal load balancers in azure), due to constraints and personally wanting to make it easier for me and everyone to easily get a healthCheck up and running on the desired entryPoint (in the values.yaml I left the entryPoint as traefik so the healthCheck isn't exposed externally). 

### More

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -45,60 +45,60 @@ deployment:
   podLabels: {}
   # -- Additional containers (e.g. for metric offloading sidecars)
   additionalContainers: []
-    # https://docs.datadoghq.com/developers/dogstatsd/unix_socket/?tab=host
-    # - name: socat-proxy
-    #   image: alpine/socat:1.0.5
-    #   args: ["-s", "-u", "udp-recv:8125", "unix-sendto:/socket/socket"]
-    #   volumeMounts:
-    #     - name: dsdsocket
-    #       mountPath: /socket
+  # https://docs.datadoghq.com/developers/dogstatsd/unix_socket/?tab=host
+  # - name: socat-proxy
+  #   image: alpine/socat:1.0.5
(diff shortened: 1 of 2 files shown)


mloiseleur:
Hello @muandane,

Thanks for this PR. 

FTM, I have two questions:
1. Why do you think it should be enabled _by default_ ?
2. Why did you not use `extraObjects` ?

muandane:
Hello @mloiseleur,
 Yes it's not enabled by default since it has the same value capabilities as the Traefik dashboard but also has the same easy way to create an ingressRoute for a healthCheck exposed on whatever entryPoint you need as for the second question tbh i didn't check the extraObjects part since i'm using Ansible tasks to deploy Traefik sometimes it gets constraining on what values i can deploy.

btw the AKS cluster is a private one. 
image

traefik/values.yaml
```diff
@@ -159,6 +159,22 @@ ingressRoute:
     middlewares: []
     # -- TLS options (e.g. secret containing certificate)
     tls: {}
+  healthcheck:
+    # -- Create an IngressRoute for the healthcheck probe
+    enabled: true
```
mloiseleur, line 164:
```suggestion
    enabled: false
```

muandane: Alright that was my bad i didn't notice it

darkweaver87 (approved):
LGTM

PR #916, opened 2023-08-31
fix: add missing separator in crds

bmagic:
### What does this PR do?

Add a missing separator in traefik.io_serverstransporttcps.yaml. 

### Motivation

I apply my crds with the "helm show crds" command. Without the separator, it generates an incorrect yaml file that doesn't allow all crds to be applied via kubectl.


mloiseleur (approved):

mloiseleur:
That's correct. It's present in upstream, too.
Thanks for this fix @bmagic

PR #920, opened 2023-09-11
add: add internal-service ingress

rayshoo:
### What does this PR do?
PR to route internal services to another Ingress controller. By exposing only the dashboard service as an internal service and routing the service to an another Ingress controller (e.g. nginx) equipped with a private load balancer, each service can be used separately without service port forwarding.

```yaml
# Example
# values.yaml#749L
  ## -- An additionnal and optional internal Service.
  ## Same parameters as external Service
  internal:
    type: ClusterIP
    ports:
      traefik:
        host: dashboard.traefik.com
        path: /
        port: 9000
        expose: true
        exposedPort: 9000
        protocol: TCP
    # labels: {}
    # annotations: {}
    # spec: {}
    # loadBalancerSourceRanges: []
    # externalIPs: []
    # ipFamilies: [ "IPv4","IPv6" ]
    ingress:
      enable: true
      annotations:
        # The nginx ingress controller in the example is an internal(private) load balancer.
        kubernetes.io/ingress.class: nginx
        cert-manager.io/cluster-issuer: letsencrypt-prod
      tls:
        enable: true
        hosts:
        - dashboard.traefik.com
        secretName: dashboard.traefik.com
      # -- Additional entries here will be added to the ingress spec.
      # -- Cannot contain tls, rules entries.
      spec: {}
```

### Motivation
I modified the chart for my own use, but I thought it would be helpful to those who want to configure the same environment as me.

### More

> This is my first time seeing the test grammar, so I think it will take some time to write it. If anyone can help, I'd appreciate it, but if not, I'll take the time to commit more.

### DIFF
--- a/traefik/templates/_service-internal-ingress.tpl
+++ b/traefik/templates/_service-internal-ingress.tpl
@@ -1,0 +1,41 @@
+{{- define "traefik.service-internal-ingress-metadata" }}
+  labels:
+  {{- include "traefik.labels" . | nindent 4 -}}
+  {{- with .Values.service.internal.ingress.labels }}
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
+{{- end }}
+
+{{- define "traefik.service-internal-ingress-tls" }}
+  - hosts:
+    {{- range $host := .Values.service.internal.ingress.tls.hosts }}
+    - {{ $host }}
+    {{- end }}
+    secretName: {{ .Values.service.internal.ingress.tls.secretName }}
+{{- end }}
+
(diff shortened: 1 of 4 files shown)


mloiseleur:
You may also use `extraObjects` for this kind of use case, as explained in [Philosophy section](https://github.com/traefik/traefik-helm-chart#philosophy).
It's adding ~ 20 lines in the values for what seems a quite specific use case.

mloiseleur:
I'm closing this PR. 
Feel to re-open it if I missed something.

Issue #925, opened 2023-10-04
Values ignored when installed as a sub chart


at88mph:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

24.0.0

### What version of Traefik are you using?

2.10.4

### What did you do?

If I try to install `traefik` directly with my `myvalues.yaml` to disable the SSL entrypoint:

```yaml
# myvalues.yaml
ports:
  websecure:
```

Then look at the output YAML:

```bash
$ helm template --values myvalues.yaml traefik traefik/traefik | grep entrypoints
          - "--entrypoints.metrics.address=:9100/tcp"
          - "--entrypoints.traefik.address=:9000/tcp"
          - "--entrypoints.web.address=:8000/tcp"
```
Which is what I expect to see.

### What did you see instead?

If, however, I declare Traefik as a Helm dependency in my `Chart.yaml`:

```yaml
dependencies:
  - name: traefik
    version: "24.0.0"
    repository: "https://traefik.github.io/charts"
    condition: traefik.install
```
Then declare my `service-values.yaml file with `traefik` options:

```yaml
# service-values.yaml
traefik:
  ports:
    websecure:
```

And run `template` on my service:

```sh
$ helm template --values service-values.yaml myservice myrepo/myservice | grep entrypoints
          - "--entrypoints.metrics.address=:9100/tcp"
          - "--entrypoints.traefik.address=:9000/tcp"
          - "--entrypoints.web.address=:8000/tcp"
          - "--entrypoints.websecure.address=:8443/tcp"
```

The `websecure` entrypoint is still declared.  Setting the `exposed` and `tls.enabled` options did nothing.  

Is this an intentional feature or a bug in Helm or in Traefik?

### What is your environment & configuration?

Kubernetes 1.27.4
Helm 3.11.3
MacOS 13.5.2 Intel

mloiseleur:
It's neither a bug in this Chart or Helm. 

It works like as expected following those instructions:
```bash
helm create foo
cd foo
[...]

PR #926, opened 2023-10-05
docs: 📚️ how to use this chart as a dependency

mloiseleur:
### What does this PR do?

Show to users how it can work when trying to integrate this Chart into its own specific Chart.

### Motivation

Users are unsure of how it works. See #925 for instance.


darkweaver87 (approved):
LGTM

PR #927, opened 2023-10-06
feat: enable support for update Listener's allowedRoute inside Gateway

renebarbosafl:
### What does this PR do?

Enable support for configuring allowedRoutes within a gateway Listener

### Motivation

When using this helm chart and integrating with kubeVela we found that It's not working properly (was breaking http-route traits) because the default policy is hardcoded as "Same" and the helm chart does not have the option to configure it as convenience so I've added this option.

### More

### DIFF
--- a/traefik/templates/gateway.yaml
+++ b/traefik/templates/gateway.yaml
@@ -18,7 +18,11 @@ spec:
     - name: web
       port: {{ .Values.ports.web.port }}
       protocol: HTTP
-
+      {{- if .Values.experimental.kubernetesGateway.namespacePolicy }}
+      allowedRoutes:
+        namespaces:
+          from: {{ .Values.experimental.kubernetesGateway.namespacePolicy }}
+      {{- end }}
     {{- if .Values.experimental.kubernetesGateway.certificate }}
     - name: websecure
       port: {{ $.Values.ports.websecure.port }}
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -126,6 +126,7 @@ experimental:
   kubernetesGateway:
     # -- Enable traefik experimental GatewayClass CRD
     enabled: false
+    namespacePolicy: All
     gateway:
       # -- Enable traefik regular kubernetes gateway
       enabled: true
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -126,7 +126,7 @@ experimental:
   kubernetesGateway:
     # -- Enable traefik experimental GatewayClass CRD
     enabled: false
-    namespacePolicy: All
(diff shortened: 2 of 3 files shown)


traefik/templates/gateway.yaml
```diff
@@ -18,7 +18,11 @@ spec:
     - name: web
       port: {{ .Values.ports.web.port }}
       protocol: HTTP
-
+      {{- if .Values.experimental.kubernetesGateway.namespacePolicy }}
+      allowedRoutes:
+        namespaces:
+          from: {{ .Values.experimental.kubernetesGateway.namespacePolicy }}
+      {{- end }}
```
mloiseleur, line 25:
```suggestion
      {{- with .Values.experimental.kubernetesGateway.namespacePolicy }}
      allowedRoutes:
        namespaces:
          from: {{ . }}
      {{- end }}
```

mloiseleur (changes requested):
Interesting, thanks for this !

My suggestion is for convenience and is not mandatory.
For merging this PR, test on this support are needed (and mandatory).

renebarbosafl:
Ok, @mloiseleur I pushed a new commit with your suggestion and enabled test support. 
Thank you!

darkweaver87 (approved):
LGTM

traefik/values.yaml
```diff
@@ -126,6 +126,7 @@ experimental:
   kubernetesGateway:
     # -- Enable traefik experimental GatewayClass CRD
     enabled: false
+    # namespacePolicy: All
```
mloiseleur, line 129:
```suggestion
    ## Routes are restricted to namespace of the gateway by default.
    ## https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.FromNamespaces
    # namespacePolicy: All
```

mloiseleur:
@renebarbosafl Since it's a parameter of `Gateway`, shouldn't we move `experimental.kubernetesGateway.namespacePolicy` to `experimental.kubernetesGateway.gateway.namespacePolicy` ?

renebarbosafl:
> @renebarbosafl Since it's a parameter of `Gateway`, shouldn't we move `experimental.kubernetesGateway.namespacePolicy` to `experimental.kubernetesGateway.gateway.namespacePolicy` ?
> 
> **EDIT**: or move all `experimental.kubernetesGateway.gateway.*` to `experimental.kubernetesGateway.` ?


Sure, I pushed a new commit moving everything from `experimental.kubernetesGateway.gateway.*` to `experimental.kubernetesGateway`. Tests were updated as well.

PR #933, opened 2023-10-12
chore: remove label whitespace at TLSOption

fty4:
### What does this PR do?

There is some whitespace next to the labels section inside the Helm chart.

This PR just removes these spaces

### More

- [x] Yes, I updated the tests accordingly -> not required


mloiseleur (approved):
That's right, we missed those ones.
Thanks @fty4.

PR #934, opened 2023-10-12
fix: http redirections breaking cert-manager challenges

renebarbosafl:
### What does this PR do?

We found that when using the HTTP redirections feature (to force usage of HTTPS) all cert-manager challenges to issue new certificates were failing, this PR adds a new priority parameter that actually enables Traefik to work with cert-manager even if using HTTP redirections.

This helm chart does not have any tests on enabling this feature either so I added it.
 
### Motivation

The solution was found here:
https://community.traefik.io/t/solved-http-to-https-redirect-ends-in-an-https-url-with-8443-appended/10136/5

(Thanks to [matthiasbaldi](https://community.traefik.io/u/matthiasbaldi))

### More

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -566,6 +566,7 @@
             {{- $toPort := index $.Values.ports $config.redirectTo }}
           - "--entrypoints.{{ $entrypoint }}.http.redirections.entryPoint.to=:{{ $toPort.exposedPort }}"
           - "--entrypoints.{{ $entrypoint }}.http.redirections.entryPoint.scheme=https"
+          - "--entrypoints.{{ $entrypoint }}.http.redirections.entryPoint.priority=10"
             {{- end }}
             {{- if $config.middlewares }}
           - "--entrypoints.{{ $entrypoint }}.http.middlewares={{ join "," $config.middlewares }}"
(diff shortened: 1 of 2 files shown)


darkweaver87 (commented):
Thanks for your contribution to Traefik helm chart.

It's more common with Traefik to use TLS-ALPN-01 with Traefik ACME provider or DNS challenge of cert-manager.

Having that said, this implementation is too specific for many users. 

What do you think about changing 
```yaml
ports:
  web:
      redirectTo: websecure
```

to:
```yaml
ports:
  web:
    redirectTo:
      port: websecure
      priority: 10
```

?

renebarbosafl:
@darkweaver87 
It makes sense! I pushed a new commit with this change.
Probably it's a breaking change for users of this feature (they'll need to update their values.yaml files).

mloiseleur (changes requested):
@renebarbosafl For users using dns challenges and this redirect, it will fail unless they set a priority. 
=> priority should be an optional settings

renebarbosafl:
Ok @mloiseleur!
I pushed a new commit changing this feature to Optional and updating tests accordingly.

mloiseleur (approved):

darkweaver87 (approved):
LGTM

Issue #936, opened 2023-10-13
Add file provider from a ConfigMap


ebuildy:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

## Feature

Add a value field `extraStaticConfig` to add extra static configuration. 

Under the hood:

- Create a configMap

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: traefik-config
data:
  config.yml: |
    {{ .Values.extraStaticConfig | nindent 2 }}
```

- Mount the configMap
- Tell traefik to use the file `- "--providers.file.filename=/mnt/traefik/config/config.yml"`

## Use case

Better TLS store configuration in case of multiple Traefik instances

## Related

https://github.com/traefik/traefik-helm-chart/issues/69

I can do the PR ASAP if the feature is accepted!

mloiseleur:
That's a good idea.
You're welcome to open a PR about it.

mloiseleur:
:thinking: Re-reading https://doc.traefik.io/traefik/getting-started/configuration-overview/
=> It will be an extra _dynamic_ configuration.

ebuildy:
hey merci @mloiseleur le projet est super bien setup j'ai pu faire la PR en 1h !

PR #937, opened 2023-10-14
feat: add file provider

ebuildy:
### What does this PR do?

As discussed at https://github.com/traefik/traefik-helm-chart/issues/936 , this PR add the option to configure a traefik file provider, as an extra configuration.

### Motivation

* Make migration to kubernetes easier: just copy/paste your traefik config
* Allow to cover the case of multiple traefik instances, fix the multiple TLSStore issue

### More


traefik/templates/_podtemplate.tpl
```diff
@@ -560,6 +564,14 @@
           - "--providers.kubernetesingress.namespaces={{ template "providers.kubernetesIngress.namespaces" $ }}"
           {{- end }}
           {{- end }}
+          {{- with .Values.providers.file }}
+          {{- if .enabled }}
+          - "--providers.file.filename=/opt/traefik/extra-config/config.yml"
```
ldez, line 569: we discourage the usage of `filename` related to problems with some file systems.
We recommend using `directory` instead.

mloiseleur (approved):

darkweaver87 (changes requested):
LGTM but you might wanna add a test for `values.yaml` defaults. The test should check that:
- no configMap is created
-  no extra volume is mounted
-  no CLI parameters nor annotations are added

mloiseleur:
@ebuildy Thanks for this. I just need more details on 

> Allow to cover the case of multiple traefik instances, fix the multiple TLSStore issue

And we should be good to go.


Since the TLSStore is configurable either with `values` or an extra resources, what is the problem when configuring multiple traefik instances ?

ebuildy:
Well I didnt manage to "scope" TLSStore object (or any other CRD like Middleware) to a specific Traefik.

To me this is not possible to run multiple Traefik instances, since there is no "selector". To give you an example, I know prometheus ServiceMonitor have NamespaceSelector field: "Selector to select which namespaces the Endpoints objects are discovered from."

traefik/values.yaml
```diff
@@ -241,6 +241,23 @@ providers:
       # By default this Traefik service
       # pathOverride: ""
 
+  file:
+    # -- Create a file provider
+    enabled: false
+    # -- Allows Traefik to automatically watch for file changes
+    watch: true
+    # -- File content (YAML format, go template supported) (see https://doc.traefik.io/traefik/providers/file/)
+    content: ""
```
mloiseleur, line 250:
```suggestion
    content: 
```

mloiseleur (changes requested):
Ok. I get it, thanks. 
Thanks for this PR, too, it looks good and tested.

Would you please follow the FHS and use a path like `/etc/traefik/dynamic` instead of `/opt/traefik/extra-conf` ?

darkweaver87 (approved):
LGTM

mloiseleur (approved):

darkweaver87 (approved):
LGTM

traefik/values.yaml
```diff
@@ -252,6 +252,23 @@ providers:
       # By default this Traefik service
       # pathOverride: ""
 
+  file:
+    # -- Create a file provider
+    enabled: false
+    # -- Allows Traefik to automatically watch for file changes
+    watch: true
+    # -- File content (YAML format, go template supported) (see https://doc.traefik.io/traefik/providers/file/)
+    content:
```
mloiseleur, line 261:
```suggestion
    content:
```

PR #939, opened 2023-10-17
Add Prometheus scraping annotations only if serviceMonitor not created

SebPlv:
### What does this PR do?

There are two ways to enable Prometheus scraping :

- If using Prometheus Operator, you have to create ServiceMonitors or PodMonitors crds.
- If using Prometheus without Operator, you have to define scraping configuration in pods (or service) annotation.

This PR intend to add Prometheus scraping Annotations only if 'metrics.prometheus' exists but 'metrics.prometheus.serviceMonitor' does not..

### Motivation

Adding scraping annotations **and** ServiceMonitor (or podMonitor) for the same pod can result to get metrics being collected twice by Prometheus.

### More

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -5,7 +5,7 @@
       {{- toYaml . | nindent 8 }}
       {{- end }}
       {{- if .Values.metrics }}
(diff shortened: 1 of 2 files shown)


mloiseleur (approved):

mloiseleur:
Thanks for this PR @SebPlv !

PR #942, opened 2023-10-19
chore(ci): 🔧 👷 add e2e test when releasing

mloiseleur:
### What does this PR do?

Launch a kind cluster and test default install on Traefik for this Chart.
Supersede #941 on which I have a strange issue.

### Motivation

Ensure it works before releasing.

### More


.github/workflows/test.yml
```diff
@@ -28,12 +28,42 @@ jobs:
       - name: Test
         run: make test
 
-      - name: Track docs
+      - name: Test if it's a release PR
```
darkweaver87, line 31: I would rather use an other workflow and filter on path. Instead, it forces you to use bash script tricks.
At least I would use conditional jobs instead of conditional steps.

.github/workflows/test.yml
```diff
@@ -28,12 +28,42 @@ jobs:
...
-            make docs
-            if ! [ -z "$(git status | grep traefik/VALUES.md)" ]; then
+          set +e
+          changed=$(git diff --name-only -r HEAD^1 HEAD | grep Chart.yaml)
+          if [[ -n $changed ]] ; then
+            echo "release=true" >> "$GITHUB_OUTPUT"
+          fi
+          set -e
+
+      - name: Check docs
```
darkweaver87, line 41: I would use an auto-commit action.

.github/workflows/test.yml
```diff
@@ -28,12 +28,42 @@ jobs:
...
+          exit 0
+
+      - name: Create kind cluster
+        if: steps.check.outputs.release
+        uses: helm/kind-action@v1.8.0
+
+      - name: Install LB on Kind
+        if: steps.check.outputs.release
+        run: |
+          kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/v0.13.11/config/manifests/metallb-native.yaml
```
darkweaver87, line 62: We may want to add a renovate config on this.

darkweaver87 (approved):
I approve this despite the fact I don't like the style at all.

.github/workflows/test.yml
mloiseleur: Yes, but probably in an other PR :).

darkweaver87 (approved):

PR #944, opened 2023-10-19
Revert "fix: 🐛 remove old CRDs using traefik.containo.us"

mloiseleur:
### What does this PR do?

This reverts commit 6484a401b6dd49ffac23dc90043d2bfd23ff226f.

### Motivation

It's not supported by current version of Traefik Proxy. One can see in logs:

```shell
E1019 09:53:46.879618       1 reflector.go:140] k8s.io/client-go@v0.26.3/tools/cache/reflector.go:169: Failed to watch *v1alpha1.IngressRoute: failed to list *v1alpha1.IngressRoute: the server could not find the requested resource (get ingressroutes.traefik.containo.us)
```

### More


darkweaver87 (approved):
LGTM

mloiseleur:
cc @HeyHugo @mathewpeterson @mythral @drewburr @jocke-I @jnoordsij

jnoordsij:
I see that, altough I removed all implementations from my cluster, I indeed still have the CRD definitions themselves installed within the cluster; hence I did not encounter this issue. But it looks like reverting this is the only solution for now.

However it may be wise to try and see if there is any way to handle this gracefully within the current version of Traefik Proxy, e.g. check for the presence of the CRD definition before trying to query any manifests implementing this. Given that I'm not familiar with the Traefik Proxy codebase, I'm not sure on where to start on this, but it sounds like it should be achievable.

PR #949, opened 2023-10-28
feat: 💥 rework and allow update of namespace policy for websecure listener

renebarbosafl:
### What does this PR do?

Complement https://github.com/traefik/traefik-helm-chart/pull/927

### Motivation

Allow updating namespace policy when defining certificates on kubernetesGateway (enabling a listener for websecure entrypoint).

### More


darkweaver87 (approved):
LGTM

BTW I think we should allow some users to specify `Selector` policy. 
As of now, we can only specify `All` or `Same` with this code ...
What do you think about it ? It's not related to your PR but it it makes sense for you (how about you @mloiseleur ?) I'll open a PR after yours will be merged.

mloiseleur (approved):

mloiseleur:
@darkweaver87 I agree with you that this Chart should allow to specify selector.

Issue #953, opened 2023-11-02
Import plugins from helm values


ndeitch:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Plugins documentation and helm values are leading to confusion regarding how to enable a plugin.

Currently to enable some external plugin is required to add the following config:

```yaml
experimental:
  plugins:
    enabled: true
```

It is possible as well to declare a list of plugins in helm values, eg:

```yaml
experimental:
  plugins:
    enabled: true
    fail2ban:
      moduleName: "github.com/acouvreur/sablier"
      version: "v1.4.1-beta.3"
```

This is suggested in plugins page:

<img width="559" alt="image" src="https://github.com/traefik/traefik-helm-chart/assets/10992150/34039650-f65e-4857-b589-a628bfa61db7">

[ref](https://plugins.traefik.io/plugins/633b4658a4caa9ddeffda119/sablier)

Although `experimental.plugins[index][pluginName]` is not used by helm to pass as additional argument to traefik container.

I believe the solution would be:

- Update plugin pages by add some note to avoid this confusion;
- Check `{{ .Values.experimental.plugins }}` in `_podtemplate.tpl` and then pass the list as args to container eg:
```yaml
args:
  - "--experimental.plugins.sablier.modulename=github.com/acouvreur/sablier"
  - "--experimental.plugins.sablier.version=v1.4.1-beta.3"
```

mloiseleur:
The second option is clearly a good idea. 
Maybe it's also possible to avoid _enabled_ and so be aligned with current plugins documentation.

Feel free to open a PR

darkweaver87:
@ndeitch, the [v26.0.0](https://github.com/traefik/traefik-helm-chart/pull/975) should resolve your issue. Feel free to reopen it if necessary ;-)

PR #961, opened 2023-11-09
Bugfix/plugin import 953

mysiki:
### What does this PR do?

Fix issue https://github.com/traefik/traefik-helm-chart/issues/953

### Motivation

need it

### More

Only syntax test with https://helm-playground.com
Result not tested...


mloiseleur:
Thanks for this PR !
It needs also some tests, to be sure it will be kept in the future.

mloiseleur (approved):
LGTM. Thanks @darkweaver87 for adding the tests.

mloiseleur (approved):
LTTM

PR #948, opened 2023-10-26
Release namespace for Prometheus Operator resources

nlamirault:
### What does this PR do?

Set namespace for `ServiceMonitor` and `PrometheusRule` resources to the release namespace if no namespace is setted into the values file, like all others resources

### More

- [ ] Yes, I updated the tests accordingly
- [ ] Yes, I ran `make test` and all the tests passed

PR #967, opened 2023-11-17
Release namespace for Prometheus Operator resources

bennesp:
### What does this PR do?

Set namespace for ServiceMonitor and PrometheusRule resources to the release namespace if no namespace is setted into the values file, like all others resources

### Motivation

Closes https://github.com/traefik/traefik-helm-chart/pull/948 which seems to be stale

### More

### DIFF
--- a/traefik/templates/servicemonitor.yaml
+++ b/traefik/templates/servicemonitor.yaml
@@ -9,8 +9,10 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ template "traefik.fullname" . }}
-  {{- with .Values.metrics.prometheus.serviceMonitor.namespace }}
+  {{- if .Values.metrics.prometheus.serviceMonitor.namespace }}
   namespace: {{ . }}
+  {{- else }}
+  namespace: {{ template "traefik.namespace" . }}
   {{- end }}
   labels:
     {{- if (.Values.metrics.prometheus.service).enabled }}
--- a/traefik/templates/servicemonitor.yaml
+++ b/traefik/templates/servicemonitor.yaml
@@ -9,7 +9,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ template "traefik.fullname" . }}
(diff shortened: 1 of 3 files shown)


traefik/templates/servicemonitor.yaml
```diff
@@ -11,6 +11,8 @@ metadata:
   name: {{ template "traefik.fullname" . }}
   {{- with .Values.metrics.prometheus.serviceMonitor.namespace }}
   namespace: {{ . }}
+  {{- else }}
+  namespace: {{ template "traefik.namespace" . }}
   {{- end }}
```
mloiseleur, line 16: `with ... else ... end` does not make sense. [with](https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with) is for setting scope,

mloiseleur (changes requested):
Thanks for this PR !

In order keep good readability, we try avoid having

```yaml
{{- if x }}
field: value
{{- else }}
field: othervalue
{{- end }}
```

Would you please use a template or a ternary instead ?

You should also start with the test, on both objects. The tests should fail without your changes on the template.

traefik/templates/servicemonitor.yaml
bennesp: Actually I understand that it seems less readable but it works. According to the documentation, `with` works exactly like `if` but it also get the scope, so `with ... else` is completely valid. This is also attested by tests. Apart from technicalities though, I understand it is not very readable so it's ok to change it

bennesp:
Can you clarify this sentence please?

> You should also start with the test, on both objects. The tests should fail without your changes on the template.

traefik/templates/servicemonitor.yaml
mloiseleur: thanks

mloiseleur (approved):
LGTM

mloiseleur:
> Apologies, I got now that you were referring to the other template that the original author modified, will write tests also for that template

No problem. Thanks for rebasing and updating the PR. It looks better now.

darkweaver87 (approved):
LGTM

PR #872, opened 2023-06-27
Fixes for traefik v3

davhdavh:
### What does this PR do?

* feat: ✨ better support for traefik v3 (beta2 and earlier no longer supported, use beta3+)
* feat: add support for openTelemetry tracing for traefik v3
* feat: add support for serverstransporttcps crd in traefik v3
* fix: no longer listen on obsolete crds for traefik v3

PR #973, opened 2023-11-27
Fix v3 CRDs. Docs to v3 and some issues on changed Middleware properties like `ipAllowList`

gcaracuel:
### What does this PR do?

Updates CRDs linkd to documentation on ApiVersion traefik.io/v1 related CRDs.
Fixes some Middleware issues, for example `ipAllowList` not being allowed in current CRD which prevents migration from `ipWhiteList` to `ipAllowList`.

This is basically a less aggressive version of #766  done by copying CRDs from https://github.com/traefik/traefik/blob/v3.0/integration/fixtures/k8s/01-traefik-crd.yml

### Motivation

I've started this to address this issue with `ipAllowList` not be valid in the CRDs but Traefik V3 not supporting depreacted `ipWhiteList`

This was also addressed by #766 but that PR was done prior to recent CRDs api version deduplication by #944 so I am just let's say merging both of them because we need CRDs to match V3 config params schema, being waiting for this more than 3 months already....

### More

- [x] Yes, I ran `make test` and all the tests passed\

### DIFF
--- a/traefik/crds/traefik.io_ingressroutes.yaml
+++ b/traefik/crds/traefik.io_ingressroutes.yaml
@@ -39,7 +39,7 @@ spec:
               entryPoints:
                 description: 'EntryPoints defines the list of entry point names to
                   bind to. Entry points have to be configured in the static configuration.
-                  More info: https://doc.traefik.io/traefik/v2.10/routing/entrypoints/
+                  More info: https://doc.traefik.io/traefik/v3.0/routing/entrypoints/
                   Default: all.'
                 items:
                   type: string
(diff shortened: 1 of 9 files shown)


mloiseleur:
Hello @gcaracuel,

Thanks for your interest in Traefik and this PR. 

Currently, this Chart follows _stable_ version of Traefik Proxy. 

There is no way in helm to specify different CRDs following an appVersion. 
=> Merging this PR would mean current user of _stable_ version of Traefik won't be able to use it. 

This chart has been updated to be compatible with Traefik v3, see #872. 
FTM users of this chart with Traefik v3 needs to upgrade their CRDs outside of this Chart, as documented in the README.

For `ipAllowList` and `ipWhiteList`, [a PR has been opened](https://github.com/traefik/traefik/pull/10249) to support both in the next stable version, Traefik v2.11

mloiseleur:
Sorry, it seems I was not clear.
The CRDs in this PR are the CRDs for Traefik Proxy v3 beta.

Were you able to start Traefik Proxy v2 with YAML generated from this PR ?

AFAIK, Traefik Proxy v2 cannot start with those CRDs.

gcaracuel:
> Sorry, it seems I was not clear. The CRDs in this PR are the CRDs for Traefik Proxy v3 beta.
> 
> Were you able to start Traefik Proxy v2 with YAML generated from this PR ?
> 
> AFAIK, Traefik Proxy v2 cannot start with those CRDs.

What's the plan then, instruct CRD to support both `ipWhitelist` & `ipAllowlist`?
All this mess prevents this Helm Chart users to use that Middleware when running V3.

That said and as lesson for the future Traefik team may start thinking in advance how to manage this migrations and revisit what the purpose of Kubernetes ApiVersions. Docs targeting main does not help either.

mloiseleur:
I'm closing this PR, since it cannot work with current Helm feature set.
Feel free to reopen it if needed.

PR #981, opened 2023-12-15
fix: remove generated fields from serverstransporttcps crd

davidalger:
### What does this PR do?

Removes generated metadata and status fields from `serverstransporttcps` CRD manifest.

### Motivation

`creationTimestamp: null` causes perma-drift when attempting to apply the manifest via a terraform `kubernetes_manifest` resource and the presence of `status` fails validation resulting in a plan error

I am working around this by stripping the generated metadata and status field from the manifest before attempting to install it on the cluster, but would be nice to be able to pass the YAML for the CRD directly to the `kubernetes_manifest` resource.


mloiseleur (approved):

mloiseleur:
Thanks.
This CRD is a specific case: it can only be used with Traefik v3. 
I updated your PR with the latest version of this CRD on Traefik v3 branch. It does not contains anymore timestamp or status field. 
Does that work for you ?

darkweaver87 (approved):
LGTM

davidalger:
@mloiseleur Works great. Thank you!

PR #988, opened 2024-01-04
Allow exposure of ports on internal service only

jnoordsij:
### What does this PR do?

Allow exposing ports to the internal service, while not exposing them on the default service at the same time.

### Motivation

Currently the exposed ports of the internal service always match those of the default service. However, it may be useful to expose certain ports to the internal service (e.g. the `metrics` or `traefik` port created by default), as the internal service primary usage is most likely some kind of exposure of endpoints to more trusted clients only.

### More


darkweaver87 (approved):
LGTM, just added one fix for a typo I noticed while reviewing your PR :-)

mloiseleur (commented):
In the current form of this PR, it looks like an hidden feature.
Wdyt about adding an example in either EXAMPLE.md or in values.yaml ?

jnoordsij:
> In the current form of this PR, it looks like an hidden feature. Wdyt about adding an example in either EXAMPLE.md or in values.yaml ?

Added them to `values.yaml` to make them more visible to make them more visible! Adding them to the examples is possible, but might be slightly too much given that I think this is a very niche feature that won't be that relevant to much people.

mloiseleur (approved):

Issue #991, opened 2024-01-14
add possibility to set runtimeClassName


legalgig:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

I would like to use gvisor runtime with traefik. Can you please add a possibility to change runtimeClassName in pod spec?

eddykaya:
I created a draft PR: https://github.com/traefik/traefik-helm-chart/pull/994

@legalgig is this what you wished for?

legalgig:
yeah it's exactly what I need but the runtimeClassName should be specified in pod spec not in container spec

source: https://kubernetes.io/docs/concepts/containers/runtime-class/#usage

eddykaya:
@legalgig when applying the template, the containerspec will be used to create a podspec, so this should work for your case. I just put the PR to review.

legalgig:
I just tried to deploy your changes and it seems that `runtimeClassName` must be specified in pod spec

```
▶ helm template traefik-test -n test . | k apply -f -
serviceaccount/traefik-test created
service/traefik-test created
ingressclass.networking.k8s.io/traefik-test created
ingressroute.traefik.io/traefik-test-dashboard created
Error from server (BadRequest): error when creating "STDIN": Deployment in version "v1" cannot be handled as a Deployment: strict decoding error: unknown field "spec.template.spec.containers[0].runtimeClassName"
```

legalgig:
I just tried to deploy traefik from master branch, sadly, same issue
```
temp/traefik-helm-chart/traefik  master ✗
▶ helm template traefik-test -n test . | k apply -f - 
serviceaccount/traefik-test created
clusterrole.rbac.authorization.k8s.io/traefik-test-test created
clusterrolebinding.rbac.authorization.k8s.io/traefik-test-test created
service/traefik-test created
ingressclass.networking.k8s.io/traefik-test created
ingressroute.traefik.io/traefik-test-dashboard created
Error from server (BadRequest): error when creating "STDIN": Deployment in version "v1" cannot be handled as a Deployment: strict decoding error: unknown field "spec.template.spec.containers[0].runtimeClassName"
```

PR #994, opened 2024-01-29
set runtimeClassName on pod spec

eddykaya:
### What does this PR do?

This PR introduces the ability to set a runtimeClassName attribute for pod spec.
It fixes https://github.com/traefik/traefik-helm-chart/issues/991

### Motivation

I want to fix https://github.com/traefik/traefik-helm-chart/issues/991

### More


traefik/values.yaml
```diff
@@ -261,7 +263,8 @@ providers:
     # -- Allows Traefik to automatically watch for file changes
     watch: true
     # -- File content (YAML format, go template supported) (see https://doc.traefik.io/traefik/providers/file/)
-    content: ""
+    content:
```
darkweaver87, line 266: It's not very consistent with the changes you did on tests line 291 for instance. Don't you think ?

darkweaver87 (commented):
Is it expected this PR is still a draft ? Also cf. my comment

traefik/values.yaml
eddykaya: @darkweaver87 I reverted the changes introduced by my autoformatter, they were not intended.

eddykaya:
> Is it expected this PR is still a draft ? Also cf. my comment

I wanted to wait for a response in the original issue if this is the intended fix which seems to be. I push the PR forward now.

traefik/templates/_podtemplate.tpl
```diff
@@ -53,6 +53,9 @@
       - image: {{ template "traefik.image-name" . }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         name: {{ template "traefik.fullname" . }}
+        {{- if .Values.deployment.runtimeClassName }}
+        runtimeClassName: {{ .Values.deployment.runtimeClassName }}
+        {{- end }}
```
mloiseleur, line 58:
```suggestion
        {{- with .Values.deployment.runtimeClassName }}
        runtimeClassName: {{ . }}
        {{- end }}        
```

mloiseleur (approved):
Thanks for your PR !
It looks good. I have a little non-blocking suggestion.

darkweaver87 (approved):
LGTM

traefik/templates/_podtemplate.tpl
darkweaver87: @eddykaya do you want to add those suggestions before we merge your PR ? 
For consistency purpose only I would add @mloiseleur's suggestions but that's up to you :-)

mloiseleur (approved):

mloiseleur (approved):

darkweaver87 (approved):
LGTM

Issue #1006, opened 2024-02-20
Restrict access to secrets in K8S cluster


jnonino:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Currently RBAC permissions created for Traefik gives access to all secrets with no restrictions. 
As defined in [rbac/clusterrole.yaml#L31](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/rbac/clusterrole.yaml#L31) and [rbac/role.yaml#L21](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/rbac/role.yaml#L21).

```
  - apiGroups:
      - ""
    resources:
      - services
      - endpoints
      - secrets
    verbs:
      - get
      - list
      - watch
```

That generates security concerns as Traefik should not need access to every secret in the cluster. 
The idea should be to have a parametrized field in the secrets permissions to use `resourceNames` and give Traefik access only to the secrets it might require to work.

mloiseleur:
Hello @jnonino,

That's right. It can be improved on this point. 
Thanks for this issue.

mloiseleur:
:warning: Don't forget that TLS certificates can be stored in `Secrets`, for instance when using `cert-manager`.

jnonino:
Sure, we are actually using it that way!

jnonino:
New PR opened: https://github.com/traefik/traefik-helm-chart/pull/1009

mloiseleur:
Fixed with 
- #1009

PR #1009, opened 2024-02-21
feat: restrict access to secrets

jnonino:
### What does this PR do?

This PR aims to allow to users to restrict the secrets accesible by Traefik on the Kubernetes cluster. For this reason, I have added a new field into RBAC configuration `secretResourceNames`. By default, that value is an empty list (`[]`) and the behaviour is the same as before this change, all secrets are accesible from Traefik.

When that field contains one or more secret names, then only those secrets are the ones accessible from Traefik.

### Motivation

Access to secrets without limits raises a security concern and Traefik should have access only to the secrets it requires to work properly, for example TLS certificates stored in secrets and used by Traefik. All other secrets in the cluster should not be accessible.
This was raised in the issue [1006](https://github.com/traefik/traefik-helm-chart/issues/1006).

### More


traefik/templates/rbac/clusterrole.yaml
```diff
@@ -28,7 +28,17 @@ rules:
...
       - endpoints
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
       - secrets
+    {{- if gt (len $.Values.rbac.accessibleSecrets) 0 }}
```
ivankatliarchuk, line 39: Can you explain why a dollar is required?

ivankatliarchuk: `$.Values.rbac.accessibleSecrets`

ivankatliarchuk (approved):
lgtm. I would assume `$` is required for safe null checks

traefik/templates/rbac/clusterrole.yaml
jnonino: Nice catch @ivankatliarchuk,  I removed that dollar symbol from `traefik/templates/rbac/clusterrole.yaml`. It is required in `traefik/templates/rbac/role.yaml` because rules are created inside a `range` loop for namespaces and `$` allows to access values from root.
It is not required in for ClusterRole, just something I forgot to cleanup

ivankatliarchuk (approved):

darkweaver87 (commented):
Hello @jnonino,

Thank you for your contribution.
The PR looks good. I'll test it. 
In the mean time WDYT about renaming `accessibleSecrets` by `secretResourceNames` ?

jnonino:
Hi @darkweaver87, I'll change the name of the field. I was actually expecting feedback about it :smile:

ivankatliarchuk:
Not sure how to remove approval. `list` and `watch` should be unrestricted. Tested recently

Screenshot 2024-02-23 at 12 41 05

traefik/templates/rbac/role.yaml
```diff
@@ -18,7 +18,17 @@ rules:
...
+      - watch
+  - apiGroups:
+      - ""
+    resources:
       - secrets
+    {{- if gt (len $.Values.rbac.secretResourceNames) 0 }}
+    resourceNames: {{ $.Values.rbac.secretResourceNames }}
+    {{- end }}
     verbs:
       - get
```
ivankatliarchuk, line 33: should be only `get`. and `list` and `watch` unrestricted

jnonino:
Linking with issue https://github.com/traefik/traefik/issues/7097 in Traefik repo as the discussion is relevant to the change added in this PR.

darkweaver87 (approved):
LGTM

mloiseleur (approved):

darkweaver87:
@jnonino , it looks like our bot is not able to merge this PR because it can't rebase on master and push to your branch. Would you mind to rebase it ?

mloiseleur:
@jnonino any chance that you can rebase this PR ?

jnonino:
> @jnonino any chance that you can rebase this PR ?

Done!!! Sorry for the delay

mloiseleur (approved):
LGTM

Alestrix:
With
```
  - apiGroups:
      - ""
    resources:
      - secrets
    verbs:
      - list
```
still present in roles.yaml, this
```
  - apiGroups:
      - ""
    resources:
      - secrets
    {{- if gt (len $.Values.rbac.secretResourceNames) 0 }}
    resourceNames: {{ $.Values.rbac.secretResourceNames }}
    {{- end }}
    verbs:
      - get
      - list
```
poses no restriction, since the `list` verb is a superset of `get`. With this, traefik still has access to every resource in that namespace. I understand that traefik is "broken" in a way that it needs that permission, but being able to define a list of `resourceNames` gives the user a false feeling of security.

Or what am I missing?

PR #1012, opened 2024-02-27
fix(rbac): do not create clusterrole for namespace deployment on Traefik v3

ChandonPierre:
A namespace scoped deployment should not create cluster scoped rbac.

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -562,6 +562,9 @@
           {{- if .Values.providers.kubernetesIngress.ingressClass }}
           - "--providers.kubernetesingress.ingressClass={{ .Values.providers.kubernetesIngress.ingressClass }}"
           {{- end }}
+          {{- if and .Values.providers.kubernetesIngress.disableIngressClassLookup (semverCompare ">=3.0.0-0" (include "imageVersion" $) ) }}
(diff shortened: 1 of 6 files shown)


mloiseleur:
@ChandonPierre Thanks. You found an interesting way for rbac namespaced. 

Would you please add test and some documentation in values.yaml ?

darkweaver87 (changes requested):
This PR only works with traefik v3.0.0.
Please update your if statements and your tests accordingly by adding something like:
`{{- if semverCompare "<3.0.0-0" (include "imageVersion" $) }}`

mloiseleur:
@ChandonPierre In its current form, this PR will be breaking for v3 users, using spec.ingressClassname with `rbac.namespaced` enabled. They will be forced to add annotations.

Wdyt about:
1. Introducing this new `provider.kubernetesIngress.disableIngressClassLookup` in values for v3 users
2. When both `provider.kubernetesIngress.disableIngressClassLookup` and `rbac.namespaced` are set, then it won't generate `ClusterRole`

mloiseleur (approved):
Thanks :+1: LGTM

darkweaver87 (approved):
LGTM.
I tested it, it works well, thanks for your contribution :-)

PR #1013, opened 2024-02-28
feat: allow excluding entrypoints from internal service

acolombier:
### What does this PR do?

This PR allows to specifically exclude some entrypoints from the default service. This can be useful if you want to expose different entrypoint on the same set of ports across the default and internal service.

### Motivation

This change was inspired from the following requirement:
- `web` configured with `redirectTo: websecure`, exposed on port 80
- `websecure` configured, exposed on port 443
- `webinsecure` not exposed

This above usecase allows to make sure that the default behavior of an ingress is to redirect to encrypted web traffic and this means, assuming that the default service uses a LoadBalancer or NodePort that all external traffic will be forced to be secured.

Now if you want to allow insecure traffic in the cluster, you will need to use a different port, say 8080, which is sub-optimal.

Thus this feature, hopefully non breaking, which allow to:

Externally:

- `web` configured with `redirectTo: websecure`, exposed on port 80
- `websecure` configured, exposed on port 443

Internally: 

- `webinsecure` configured, exposed on port 80
- `websecure` configured, exposed on port 443

using the following config:

```yaml
ports:
  web:
    onlyExternal: true
  webinsecure:
    port: 10080
    exposedPort: 80
    expose: false
    exposeInternal: true
```

### More

### DIFF
--- a/traefik/templates/_service-internal.tpl
+++ b/traefik/templates/_service-internal.tpl
@@ -35,7 +35,7 @@
 
 {{- define "traefik.service-internal-ports" }}
   {{- range $name, $config := . }}
-  {{- if (or $config.expose $config.exposeInternal) }}
+  {{- if (and (or $config.expose $config.exposeInternal) (not $config.onlyExternal))  }}
   - port: {{ default $config.port $config.exposedPort }}
     name: {{ $name | quote }}
     targetPort: {{ default $name $config.targetPort }}
--- a/traefik/tests/service-internal-config_test.yaml
+++ b/traefik/tests/service-internal-config_test.yaml
@@ -203,3 +203,27 @@ tests:
             port: 3000
             protocol: TCP
             targetPort: internal
+  - it: should not expose ports that are exposed on default service only
+    set:
+      service:
+        internal:
+          enabled: true
+      ports:
+        web:
+          onlyExternal: true
+        websecure:
+          onlyExternal: true
+        internal:
+          expose: false
+          exposeInternal: true
+          port: 4000
+          exposedPort: 3000
+          protocol: TCP
+    asserts:
+      - equal:
+          path: spec.ports
+          value:
+            - name: internal
+              port: 3000
+              protocol: TCP
+              targetPort: internal
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -635,6 +635,10 @@ ports:
     # note that ports exposed on the default service are exposed on the internal
     # service by default as well.
     exposeInternal: false
+    # -- Defines whether the port should be exposed only on the default service;
+    # this can be useful if you want to define different entrypoints for external 
+    # and internal service but expose them on competing ports
+    onlyExternal: false
   web:
     ## -- Enable this entrypoint as a default entrypoint. When a service doesn't explicitly set an entrypoint it will only use this entrypoint.
     # asDefault: true
@@ -654,6 +658,10 @@ ports:
     # note that ports exposed on the default service are exposed on the internal
     # service by default as well.
     exposeInternal: false
+    # -- Defines whether the port should be exposed only on the default service;
+    # this can be useful if you want to define different entrypoints for external 
+    # and internal service but expose them on competing ports
+    onlyExternal: false
     # Port Redirections
     # Added in 2.2, you can make permanent redirects via entrypoints.
     # https://docs.traefik.io/routing/entrypoints/#redirection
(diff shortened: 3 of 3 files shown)


mloiseleur:
Hello @acolombier,

Thanks for this PR. 

This chart is already quite long and complex. I'm not at ease to have `onlyExternal` + `exposedPort` + `exposeInternal`.

Wdyt of re-working this section with something like:

```yaml
ports:
  # http on both
  web:
    expose:
      internal: 80
      external: 80
  # only internal
  webinsecure:
     expose:
        internal: 10080
  # only external
  websecure:
     expose:
        external: 443
  # not exposed
  traefik: {}
```

cc @jnoordsij , since you did the last PR on this subject.

darkweaver87:
Hello @mloiseleur , @acolombier and @jnoordsij,

Since the [PR](https://github.com/traefik/traefik-helm-chart/pull/988) of @jnoordsij I think the implementation is too complex.
WDYT about simply breaking here and configuring services following the service spec and just let ports configure deployment arguments. Something like that for the default:

```yaml
ports:
  traefik:
    port: 9000
    protocol: TCP
  web:
    port: 8000
    protocol: TCP
  websecure:
    port: 8443
    protocol: TCP
    http3:
      enabled: false
    tls:
      enabled: true
  metrics:
    port: 9100
    protocol: TCP
```

and services:

```yaml
services:
  traefik:
    type: LoadBalancer
    spec:
      ports:
      - port: 80
        name: "web"
        targetPort: web
        protocol: TCP
      - port: 443
        name: "websecure"
        targetPort: websecure
        protocol: TCP
```

For those who want to customize their deployment with multiple services for different purposes:

```yaml
services:
  traefik-external:
    type: LoadBalancer
    spec:
      ports:
      - port: 443
        name: "websecure"
        targetPort: websecure
        protocol: TCP
  traefik-internal:
    type: ClusterIP
    spec:
      ports:
      - port: 80
        name: "web"
        targetPort: web
        protocol: TCP
  traefik-other-specific:
    type: ClusterIP
    spec:
      clusterIP: 42.42.42.42
      ports:
      - port: 8080
        name: "web"
        targetPort: web
        protocol: TCP
```

I think it's more customizable on the long term.

jnoordsij:
Ah yeah, I agree things have gotten a bit more complex once again with my PR, but that's mainly because I aimed for backwards compatibility. I think the easiest "breaking" implementation for this that does not require yet another flag is changing the `(or $config.expose $config.exposeInternal)` check I introduced to just `$config.exposeInternal` and explicitly requiring people to set it to true on all ports they want on their internal service. Given that probably a very limited subset of users actually make use of the internal service, this will be a breaking change in technical terms, but probably quite less in practice.

I think @mloiseleur's suggestion is very similar, allowing a little more flexibility on port numbers, which is good on one hand but will result in some value duplication for end-users, which I personally tend to avoid a bit (although it will be so closely grouped in this case it's hardly an issue). However I do like it syntax-wise, maybe a variant with just using booleans instead of actual port numbers is both more resembling of the current chart values and still an improvement in readability? E.g.
```yaml
ports:
  # http on both
  web:
    port: 80
    expose:
      internal: true
      external: true
  # only external
  websecure:
     port: 443
     expose:
        internal: false # or left out
        external: true
```

Making everything entirely customisable as @darkweaver87 suggested is definitely something that will reduce any further alterations on this subject, as basically end-users are near free to alter anything they want. The main issue in my eyes would again be possibly splitting/duplicating port configuration, which would make it a little harder to see the relations between ports and services, and introduce some risk with numbered and/or named port references.

It would also require more knowledge of actual Kubernetes manifests to properly make use of it, however people using multiple services in this manner probably already have sufficient knowledge on that account, and for the default behavior we'd still be providing the appropriate values. A final thing I worry about slightly in such a case is that it might be harder to account for API changes in Kubernetes Services, although I think it's highly unlikely we'll see any sort of major shift there.

An alternative would be something in between: leave the already defined services as is, but add some kind of possibility to define "additional" services through some chart values, that can still access the current port configuration in some appropriate manner. But that would probably require some advanced Helm hacking...

mloiseleur:
One (breaking) PR is fine.

I think I agree with @jnoordsij improvement on my proposal. This is currently how it works and no one asked for a change.

@darkweaver87 wdyt ?

jnoordsij:
@darkweaver87 I do agree allowing more broad service changes makes sense, but I think most of that will boil down to the 'default' service. So probably we should prefer the key 'default' over 'external' in the above suggestions, given that a single customisable service will probably already cover lots of usecases.

Also as an intermediate suggestion, I've created https://github.com/traefik/traefik-helm-chart/compare/master...jnoordsij:traefik-helm-chart:service-refactoring to show how we can possibly consolidate all of this into a single service template which can render multiple services, just by providing arbitrary service config and adding appropriate port expose flags to the newly suggested expose syntax. It's still slightly hackish, and tests probably fail for internal services and maybe elsewhere, but it might be worth exploring such an option.

darkweaver87:
Thanks @jnoordsij. It's really closed to what I was thinking about. I like the idea of having a map making possible to have as many services as we want and above all the fact that the exposure boolean would be dynamic.

acolombier:
Looking good @jnoordsij.
I'm closing this PR so you can get yours in as it will solve the usecase above.

ebuildy:
ddd

PR #1020, opened 2024-03-16
feat: refactor custom services and port exposure

jnoordsij:
### What does this PR do?

This PR introduces the option to easily create custom Service templates, using the same value structure as the default service. It also changes the port `exposure` value to be a map, that maps service names to booleans indicating wether or not the port should be exposed on the service. This allows for ports to be easily repeated across services.

As discussed, this is done in a non-backwards compatible manner, as the new simplified syntax is preferred over keeping changes BC.

### Motivation

The discussion on https://github.com/traefik/traefik-helm-chart/pull/1013, to allow hiding ports from an internal service.

### More

### To Do / Discuss
- [ ] Write some migration notes detailing:
  - how the `expose` flag on ports is now a map (relevant to possibly quite a substantial group of users)
  - how the `internal` service has been removed and can be recreated through values (relevant to most likely a very small number of people)
- [ ] Create an example in the docs about custom services; probably some values that are more or less identical to the 'old' internal service
- [ ] Check for possible additional relevant things to test


traefik/templates/service.yaml
```diff
@@ -1,9 +1,22 @@
...
+{{- if $service.enabled -}}
+
+{{- $fullname := printf "%s-%s" (include "traefik.fullname" $) $name }}
+{{- if eq $name "default" -}}
+{{- $fullname = include "traefik.fullname" $ }}
+{{- end -}}
+
+{{- if ge (len $fullname) 60 }} # 64 - 4 (udp-postfix) = 60
+  {{- fail "ERROR: Cannot create a service whose full name contains more than 60 characters" }}
+{{- end }}
```
mloiseleur, line 14: Wdyt about moving this block in a `traefik.servicename` helper function ?

traefik/values.yaml
```diff
@@ -626,22 +626,20 @@ ports:
     # -- You SHOULD NOT expose the traefik port on production deployments.
     # If you want to access it from outside your cluster,
     # use `kubectl port-forward` or create a secure ingress
-    expose: false
+    expose:
+      default: false
```
mloiseleur, line 630: mmh :thinking: I understand why you name it `default`, but I'm unsure about it. 
Just looking at _values_, it feels like "would you like to expose it _by default_ ?"

Wdyt about :
Option a:

```yaml
expose:
  external: true|false
```

Option b:

```yaml
expose:
  service: true|false
```

Option c:

```yaml
expose:
  default-service: true|false
```

darkweaver87 (commented):
LGTM. 
I would maybe warn the user if he exposes a port refering a non-exisiting service.

traefik/values.yaml
darkweaver87: It comes from our previous discussions. It don't surprises me much as `expose` is now a map of `service_name: boolean`

traefik/templates/service.yaml
jnoordsij: Yeah that makes sense. Just pushed a commit to add this. Things are getting a bit complicated with all the dicts passing roots and other arguments involved, but it should be fine I think.

traefik/values.yaml
jnoordsij: Expose already was a boolean tied to service, so assuming that "service" is implicit context I think `default` makes sense, although `default-service` looks fine to me as well. Moreover, exposing it by "default" is also correct for "basic" chart usage, as that's what the default service does.

I'd argue against `external`, given that that would somehow dictate how one should use the "default" service with this chart, while one can easily use the "default" service as an internal service or have multiple external services, so it seems a bit too narrow for me.

mloiseleur: Ok, It works for me.

mloiseleur (dismissed):

jnoordsij:
Also I'll see if I can add the bits of documentation and upgrade notes later this week!

jnoordsij:
Added examples and some upgrade notes to the changelog; if you think another location is more suitable let me know!

> LGTM. I would maybe warn the user if he exposes a port refering a non-exisiting service.

I've thought about this for a bit, but could not think of an elegant/efficient way to achieve this, and given that it's not really hurtful to have users do this, I'd say we'll leave it as an idea for a possible future improvement.

darkweaver87:
> I've thought about this for a bit, but could not think of an elegant/efficient way to achieve this, and given that it's not really hurtful to have users do this, I'd say we'll leave it as an idea for a possible future improvement.

We can even think about failing at some point like we do for many other things in that chart. But yes, let's think more about it and we will open a new PR if needed.

EXAMPLES.md
```diff
@@ -99,7 +99,7 @@ extraObjects:
 
 To expose the dashboard without IngressRoute, it's more complicated and less
 secure. You'll need to create an internal Service exposing Traefik API with
-special _traefik_ entrypoint.
+special _traefik_ entrypoint. This can be achieved using [custom services](#add-custom-internal-services) as described below.
```
mloiseleur, line 102:
```suggestion
special _traefik_ entrypoint. This internal Service can be created from an other tool, with the `extraObjects` section or using [custom services](#add-custom-internal-services).
```

EXAMPLES.md
```diff
@@ -473,6 +473,64 @@ spec:
...
+```
+
+Ports can then be exposed on this service by using the port name to boolean mapping `expose` on the respective port;
+e.g. to expose the `traefik` API port on your internal service so pods within your cluster can use it, you can do:
+
+```yaml
+ports:
+  traefik:
+    expose:
+      default: false # keep this disabled!
```
mloiseleur, line 501:
```suggestion
    expose:
      # Sensitive data should not be exposed on the internet
      # => Keep this disabled !
      default: false
```

EXAMPLES.md
```diff
@@ -473,6 +473,64 @@ spec:
...
+metadata:
+  name: traefik-internal
+  namespace: traefik
+  labels:
+    app.kubernetes.io/name: traefik
+    app.kubernetes.io/instance: traefik-traefik
+    helm.sh/chart: traefik-26.1.0
+    app.kubernetes.io/managed-by: Helm
+    traefik-service-label: internal
+  annotations:
```
mloiseleur, line 521:
```suggestion
[...]
```

EXAMPLES.md
```diff
@@ -473,6 +473,64 @@ spec:
...
+
+```yaml
+ports:
+  traefik:
+    expose:
+      default: false # keep this disabled!
+      internal: true
+```
+
+This will then provide the following additional service manifest:
```
mloiseleur, line 505:
```suggestion
This will then provide an additional Service manifest, looking like this:
```

mloiseleur:
Thanks for all of this, @jnoordsij, it's really good !

I left some minor suggestions.

darkweaver87 (approved):
LGTM

traefik/Changelog.md
```diff
@@ -1,5 +1,28 @@
 # Change Log
 
+## 27.0.0 (unreleased)
```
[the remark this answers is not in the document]
jnoordsij: Yeah that was exactly my idea although I did not spell it out ;)

jnoordsij:
Thanks for the reviewing!

Incorporated the suggestions and added one more change: remove the need to explicitly add the `enabled: true` flag, which seems superficial to me for the new services and the default service already has a `true` value in the default `values.yaml`, so this should not alter any existing behavior.

mloiseleur (approved):

PR #1027, opened 2024-04-03
feat!: :boom: initial support of Traefik Proxy v3

mloiseleur:
### What does this PR do?

Follows [migration guide](https://doc.traefik.io/traefik/v3.0/migration/v2-to-v3/) on this Chart, each commit for a specific step:

- **v3: http3 & experimental**
- **v3: influxdb v1 removed**
- **v3: core.defaultRuleSyntax**
- **v3: remove preferServerCipherSuites in TLSOption**
- **v3: remove traefik.containo.us API group**
- **v3: removeAPI Group networking.k8s.io/v1beta1**
- **v3: tracing is powered exclusively with otel**
- **fix: tlsoption**
- **v3: new addInternals option**
- **v3: otlp support on metrics and tracing**
- **v3: supports only v3**

Since it's not possible to support both v2 and v3, this PR is a breaking PR. Those who wants to stay on Traefik v2 should not go further that v27.x.y of this Chart.

### Motivation

Traefik Proxy V3 rc3 has been released a few weeks ago. The GA is [coming](https://github.com/traefik/traefik/issues?q=is%3Aissue++milestone%3A3.0+is%3Aopen+).

### More


traefik/tests/service-config_test.yaml
```diff
@@ -13,6 +13,13 @@ tests:
     asserts:
       - hasDocuments:
           count: 0
+  - it: should fail when trying to use this Chart with Traefik Proxy v2
```
darkweaver87, line 16: OK, now I see, but it should not be specifically checked in services.

darkweaver87 (commented):
Awesome PR!
cf. my comments
Additionally, with the logging revamp, it looks to see anything at startup. This might confuse users. WDYT about setting `logs.general.level=INFO` by default ?

Did you see this warning while running tests ? 
```warning: cannot overwrite table with non table for traefik.experimental.plugins (map[])```

File: traefik/tests/service-config_test.yaml (hunk 1 above)
Comment (mloiseleur) [reply traefik/tests/service-config_test.yaml]: Fixed with 10f9102

File: traefik/templates/service.yaml [hunk 2]
```diff
@@ -1,3 +1,7 @@
+{{- if semverCompare "<3.0.0-0" (include "imageVersion" $) }}
```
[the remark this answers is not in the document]
mloiseleur: Fixed with 10f9102

SantoDE (approved):
Gave it a test this morning in one of my environments with Traefik V3. Works pretty well. All manual edits I had to do are now gone :) 

Great job!

darkweaver87 (approved):
LGTM

jnoordsij:
Bit late to the party I guess, but great to see this landing! Looking forward to the actual stable release.

Just one suggestion: we should probably update the Installing section of the `README` to note that for using Traefik v2, only Chart versions up to 27.x.y are officially supported and any version above that might be compatible, but no guarantees nor active support is provided for v2 app + >= v28 chart.

mloiseleur:
> Bit late to the party I guess, but great to see this landing! Looking forward to the actual stable release.
> 
> Just one suggestion: we should probably update the Installing section of the `README` to note that for using Traefik v2, only Chart versions up to 27.x.y are officially supported and any version above that might be compatible, but no guarantees nor active support is provided for v2 app + >= v28 chart.

Would you please review #1035 then ? :)

PR #1029, opened 2024-04-06
feat(podtemplate): set GOMEMLIMIT, GOMAXPROCS when limits are defined

ChandonPierre:
### What does this PR do?

Always define `GOMEMLIMIT` and `GOMAXPROCS` env vars when memory limits and CPU limits are defined.

### Motivation

Traefik is written in GoLang, and is thus [subject to CPU throttling](https://www.riverphillips.dev/blog/go-cfs/) when CPU limits are defined in a container without GOMAXPROCS set. We see this reflected in the `container_cpu_cfs_throttled_seconds_total` metric in our environment; where CPU throttle rate exceeds 1000%

### More

### DIFF
--- a/traefik/tests/pod-config_test.yaml
+++ b/traefik/tests/pod-config_test.yaml
@@ -29,6 +29,16 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].env[0].valueFrom.resourceFieldRef.resource
           value: limits.memory
+  - it: should have not set GOMAXPROCS without cpu limit
+    asserts:
+      - notEqual:
+          path: spec.template.spec.containers[0].env[0].name
+          value: GOMAXPROCS
+  - it: should have not set GOMEMLIMIT without cpu limit
+    asserts:
+      - notEqual:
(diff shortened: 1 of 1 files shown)


mloiseleur:
\o. that's right @ChandonPierre !

It's confirmed with [other](https://medium.com/inlocotech/running-go-application-on-kubernetes-6fb55f908258) [blog](https://blog.howardjohn.info/posts/gomaxprocs/) [post](https://medium.com/@jettycloud/making-sense-of-kubernetes-cpu-requests-and-limits-390bbb5b7c92) and the existence of [automaxprocs](https://github.com/uber-go/automaxprocs).

mloiseleur (commented):
Thanks for this PR :+1: 
Would you please ensure within a test to ensure there is no GOMAXPROCESS & GOMEMLIMIT when resources are not set (since it's the default behavior) ?
Your PR is clearly not doing that : I'm suggesting to add that, just in case, as a safety net.

mloiseleur (approved):

darkweaver87 (approved):
LGTM

PR #1037, opened 2024-04-17
feat: :rocket: add initial support for traefik-hub api gateway

darkweaver87:
### What does this PR do?

This PR adds the initial support for new traefik-hub API gateway.

### Motivation

This will help traefik proxy v3 users to migrate smoothly to traefik-hub API gateway.

### More


traefik/templates/_podtemplate.tpl
```diff
@@ -110,15 +110,21 @@
...
           hostPort: {{ default $config.hostPort $config.http3.advertisedPort }}
-        {{- end }}
+           {{- end }}
           protocol: UDP
+          {{- end }}
+         {{- end }}
         {{- end }}
-        {{- end }}
+        {{- if .Values.hub.token }}
+         {{- $listenAddr := default ":9943" .Values.hub.admission.listenAddr }}
```
darkweaver87, line 124: Looks repeated. You can avoid it by setting it a bit earlier.

traefik/templates/_podtemplate.tpl
```diff
@@ -640,32 +643,32 @@
...
-               {{ with $value }}
+               {{- with $value }}
           - "--hub.metrics.opentelemetry.tls.{{ $field }}={{ $value }}"
                {{- end }}
               {{- end }}
              {{- end }}
             {{- end }}
-            {{ with .platformurl }}
-          - "--hub.platformurl={{ . }}"
+            {{- with .platformUrl }}
```
darkweaver87, line 652: This was correct. This should be aligned with command line arguments.

traefik/values.yaml
```diff
@@ -980,19 +980,20 @@ extraObjects: []
...
 hub:
-  # Set a license token to enable Traefik Hub support. Default: "".
-  token: ""
+  # Name of Secret with key 'token' set to a valid license token.
+  # It enables API Gateway.
+  token:
   admission:
     # -- WebHook admission server listen address. Default: "0.0.0.0:9943".
-    listenaddr:
+    listenAddr:
```
darkweaver87, line 988: Same thing here. Traefik arguments are not camel cased.

traefik/hub-crds/kustomization.yaml
```diff
@@ -1,12 +0,0 @@
----
```
darkweaver87: I don't agree with this commit. People not using hub doesn't need to install CRDs.

darkweaver87 (commented):
cf. my comments

traefik/templates/_podtemplate.tpl
mloiseleur: All parameters of Traefik are camel cased in the documentation

mloiseleur (approved):

mloiseleur (approved):

PR #1040, opened 2024-04-18
fix: 🐛 only expose http3 port on service when TCP variant is exposed

jnoordsij:
### What does this PR do?

Currently when you enable http3 for a defined port, e.g. `websecure`, the corresponding http3 variant will be exposed on all services, regardless of wether or not the corresponding TCP port is exposed.

This PR ensures that the http3 variant is only exposed on those services which also expose the corresponding TCP variant.

### Motivation

Fix a (mostly harmless) bug I noticed when looking into my Service port exposure.

### More


mloiseleur:
The PR looks technically ok. 
It's lacking some information about this, at least in _values_
:thinking: Are we sure this behavior is expected ? There is no use case with internal http3 ?

jnoordsij:
Currently http3 is never considered a separate port in terms of _values_, but rather a setting available on other ports (which makes sense, as http3 can only work on top of a TCP http/2 port). This PR makes sure that the `expose` flag is used for both variants at the same time: if you disable (not expose) the TCP port, you won't get the http3 variant. If you do expose the base port, the http3 variant will be there as well.

We could even go further and allow toggling http3 individually per service, to allow only exposing http/2 but not http/3 on some services but not all, but I doubt that has any usecase.

mloiseleur (approved):
LGTM

PR #1041, opened 2024-04-19
Ability to override liveness and readiness probe paths

matkam:
### What does this PR do?

Provides the option to override the liveness and readiness probe HTTP GET paths.

### Motivation

I wrote a Traefik Plugin that adds some functionality to be used by the readiness probe only. In order to use it, the readiness probe path needs to be different from the liveness probe path. Here is the plugin: https://github.com/livekit/traefik-readiness-plugin

### More

- [ x] Yes, I updated the tests accordingly
- [ x] Yes, I ran `make test` and all the tests passed


mloiseleur:
Hello @matkam,

Thanks for your interest in Traefik. You wrote a plugin, that's cool. 

FTM, there is no plugin supported with this chart.

At first glance, I'm not sure to understand why would someone consider Traefik _not_ ready when a threshold on CPU is reached. It's usually managed with HPA or VPA. As explained in the README:

> To keep this HelmChart as generic as possible we tend to avoid integrating any third party solutions nor any specific use cases.

So => Would you please detail the use case of this plugin ? (and so why we should consider include it ?)

matkam:
Hi @mloiseleur!
Nice to see a familiar face. You also reviewed my PR in https://github.com/kubernetes-sigs/external-dns.

>  I'm not sure to understand why would someone consider Traefik not ready when a threshold on CPU is reached. It's usually managed with HPA or VPA.

This is true under the normal circumstance of short lived HTTP requests, but does not hold true with long running WebSocket connections (multiple hours). The problem is that even after scaling up Traefik Pods, older saturated Pods will continue to receive new connections, and can become overloaded. To get around this problem, we are currently over provisioning Traefik. But if Traefik gets taken out of rotation once it has reached a usage threshold, over provisioning wouldn't be necessary.  Weighted load balancing would probably work better, but not all cloud providers have load balancers that can do that.

mloiseleur:
There is an open issue on _readiness_ : https://github.com/traefik/traefik/issues/10458

You can also tune your probe using api/rawdata content.

matkam:
> There is an open issue on _readiness_ : [traefik/traefik#10458](https://github.com/traefik/traefik/issues/10458)
> 
> You can also tune your probe using api/rawdata content.

thanks for the tip @mloiseleur! I've added a rawdata check to my readiness plugin: https://github.com/livekit/traefik-readiness-plugin/commit/af7e8f86e8867c7c84db25fc7431f5a29e505ef7

kind of amazing this isn't built in already.

darkweaver87 (approved):
LGTM.
To me, nothing is specific here. Some people might want to tune it using @matkam plugin or not.

mloiseleur (approved):
LGTM

PR #1043, opened 2024-04-22
Consistent use of --entryPoints CLI flag capitalization

jnoordsij:
### What does this PR do?

Consistently use capitalization "`entryPoints`" within CLI flags.

### Motivation

Noted some inconsistencies while checking docs and implementing things. Given the Traefik docs mostly use this pattern, and after https://github.com/traefik/traefik/pull/10650 the docs should consistently use this pattern, it's probably neater to use it here as well.

### More

- [x] Yes, I updated the tests accordingly
- [x] Yes, I ran `make test` and all the tests passed

PR #1044, opened 2024-04-22
Allow setting transport settings for ports

jnoordsij:
### What does this PR do?

Allows setting various transport options on ports as defined in the [docs](https://doc.traefik.io/traefik/v3.0/routing/entrypoints/#transport).

Note: based on #1043 so that should be merged and then I hope there'll be no merge conflicts...

### Motivation

Wanting to set one myself with an interface slightly more convenient then using the `additionalArguments` value.

### More


jnoordsij:
As a side note: I personally created this as to set `requestAcceptGraceTimeout` to a non-zero value for all ports by default, given that with how the Kubernetes API handles endpoints, there might in practice exist some interval where the pod has started terminating but the endpoint has not been deregistered and/or this has not been processed by other consumers yet. In such a case, consumers may attempt to send requests to the terminating pod, causing failures.

I think it might be reasonable to set a default non-zero value in this chart by default; let me know if you agree!

mloiseleur (changes requested):
It's missing some documentation on this. 
Wdyt about updating the _values_ ?

jnoordsij:
Copied the values from the test to `values.yaml` as example.

traefik/values.yaml
```diff
@@ -648,15 +648,28 @@ ports:
...
+    # transport:
+    #   respondingTimeouts:
+    #     readTimeout: 60
+    #     writeTimeout: 0
+    #     idleTimeout: 180
+    #   lifeCycle:
+    #     requestAcceptGraceTimeout: 0
+    #     graceTimeOut: 10
+    #   keepAliveMaxRequests: 0
+    #   keepAliveMaxTime: 0
```
mloiseleur, line 672:
```suggestion
    #   keepAliveMaxTime: 42s
```

traefik/values.yaml
```diff
@@ -648,15 +648,28 @@ ports:
...
+    # -- Set transport settings for the entrypoint; see also
+    # https://doc.traefik.io/traefik/routing/entrypoints/#transport
+    # transport:
+    #   respondingTimeouts:
+    #     readTimeout: 60
+    #     writeTimeout: 0
+    #     idleTimeout: 180
+    #   lifeCycle:
+    #     requestAcceptGraceTimeout: 0
+    #     graceTimeOut: 10
```
mloiseleur, line 670:
```suggestion
    #     requestAcceptGraceTimeout: 0s
    #     graceTimeOut: 10s
```

traefik/values.yaml
mloiseleur, line 672:
```suggestion
    transport:
      respondingTimeouts:
        readTimeout: 
        writeTimeout:
        idleTimeout: 
      lifeCycle:
        requestAcceptGraceTimeout:
        graceTimeOut:
      keepAliveMaxRequests:
      keepAliveMaxTime:
```

mloiseleur (commented):
I suggest to make them appear in values, so third party tools are able to use it.

jnoordsij:
> Wdyt of making them appear in values, so third party tools are able to use it ?

Just tried it, but appearently my `if (toString .X)` check fails on nil/null, as it is converted to the string `<nil>` rather than empty string. I figured using just "if" is also not the way to go, as then numeric 0 is ignored (it's falsey).

So there's two choices here AFAICS:
- Change the checks to `if ne (quote .X) "\"\""`, which looks even uglier to me, but does seem to be more accurate in result;
- Don't support empty values, and use literal `""` values in the examples, which are not passing the existing test.

mloiseleur:
> So there's two choices here AFAICS:

I have a third choice to suggest:

With values like this:

```yaml
    transport:
      respondingTimeouts:
```

should work with template code like this:

```yaml
                {{- with .readTimeout }}
          - "--entryPoints.{{ $entrypoint }}.transport.respondingTimeouts.readTimeout={{ . }}"
                {{- end }}
```

jnoordsij:
The `with` operator would not allow passing `0` (numeric) as argument, as it's falsey, so e.g.
```yaml
    transport:
      respondingTimeouts:
        readTimeout: 60
        writeTimeout: 0
        idleTimeout: 42s
```

results in

```yaml
                                - --entryPoints.websecure.transport.respondingTimeouts.readTimeout=60
                                - --entryPoints.websecure.transport.respondingTimeouts.idleTimeout=42s
```
(with no `writeTimeout` argument printed).

mloiseleur:
\o you're right. Sorry, the correct template code should be:

```yaml
                {{- with .readTimeout | toString }}
          - "--entryPoints.{{ $entrypoint }}.transport.respondingTimeouts.readTimeout={{ . }}"
                {{- end }}
```

jnoordsij:
Ah yeah, an explicit `nil` check works too. So 3 styles to pick from I guess:

1:
```yaml
              {{- if ne .readTimeout nil }}
                {{- with .readTimeout | toString }}
          - "--entryPoints.{{ $entrypoint }}.transport.respondingTimeouts.readTimeout={{ . }}"
                {{- end }}
              {{- end }}
```

2:
```yaml
                {{- if and (ne .readTimeout nil) (toString .readTimeout) }}
          - "--entryPoints.{{ $entrypoint }}.transport.respondingTimeouts.readTimeout={{ .readTimeout }}"
                {{- end }}
```

3:
```yaml
                {{- if ne (quote .readTimeout) "\"\"" }}
          - "--entryPoints.{{ $entrypoint }}.transport.respondingTimeouts.readTimeout={{ .readTimeout }}"
                {{- end }}
```

mloiseleur:
Style 1. or 2. is fine. Choose which one you prefer.
Style 3. seems a door open to more not-yet-discover problems.

mloiseleur (approved):
LGTM

PR #1055, opened 2024-04-30
Update DO PROXY Protocol example

jnoordsij:
### What does this PR do?

Update the DO PROXY Protocol example to
- Use chart logic instead of manual `additionalArguments` for better forward compatibility
- Introduce a YAML anchor to dedupe IP lists

### Motivation

Matches my own setup and I thing it's slightly more suitable.

### More


mloiseleur (approved):
LDTM !

PR #1067, opened 2024-05-15
fix: don't add unsupported otlp tracing options

geigerzaehler:
The options `--tracing.otlp.http=true` and `--tracing.oltp.grpc=true` [are not supported][1] (albeit documented). Before this change, setting the values `tracing.otlp.http.enabled` or `tracing.otlp.grpc.enabled` to `true` resulted in the prod crashing.

[1]: https://github.com/traefik/traefik/issues/10721

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -320,7 +320,6 @@
            {{- end }}
            {{- with .http }}
             {{- if .enabled }}
-          - "--metrics.otlp.http=true"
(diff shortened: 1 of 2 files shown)


geigerzaehler:
Cool, thanks. Another thing that confused me is that Traefik itself enables Otel HTTP collection by default when `tracing.otlp` is set to true—even if `tracing.otlp.http.enabled` is set to `false`. You may want to address this as well.

PR #1069, opened 2024-05-17
fix(Traefik Hub): do not deploy mutating webhook when enabling only API Gateway

mloiseleur:
### What does this PR do?

Remove mutating webhook when deploying Traefik Hub API Gateway

### Motivation

It's needed only for Traefik Hub API Management

### More


darkweaver87 (approved):
LGTM

PR #1077, opened 2024-05-24
feat: add deletecollection right on secrets

jspdown:
### What does this PR do?

This PR adds the `deletecollection` right on Secrets. 

### Motivation

The goal is to allow Traefik Hub controller to delete Secrets more efficiently when cleaning up outdated resources.

### More

- [ ] ~Yes, I updated the tests accordingly~


mloiseleur:
Thanks for this PR, @jspdown !
This chart allows to choose between namespace rbac or cluster-wide rbac. It seems this PR covers only cluster role.
Would you please update `role.yaml` accordingly ?

jspdown:
@mloiseleur Thanks for your review. I've updated the PR accordingly.

Please note that I made some cleanup as well:

- Removing duplicated resource rights (services and endpoints)
- Removed resources that are no longer supported (edgeingresses, deployment, statefulsets, and daemonsets)
- Removed write access on ingresses

mloiseleur (approved):

youkoulayley (approved):

PR #1083, opened 2024-06-04
fix: 🐛 namespaced rbac with kubernetesIngress provider disabled

mloiseleur:
### What does this PR do?

Fix template logic in `clusterrole`.

### Motivation

This case should be supported.

### More


darkweaver87 (approved):
LGTM

PR #1093, opened 2024-06-15
Jarush/fix secret slice

### LINKED ISSUE #1090: rbac secretResourceNames is templated incorrectly as a string
### Welcome!

28.2.0

v3.0

### What did you do?

As written the template code for scoping the generated `ClusterRole` to specific secrets only allows for a single value.

```sh
cat << EOF |helm template -f - traefik traefik/traefik
rbac:
  secretResourceNames:
    - secret1
    - secret2
EOF
```

I expected this to create a `ClusterRole` that had the secret permission for:

```yaml
resourceNames:
  - secret1
  - secret2
 ```

### What did you see instead?

The templated `ClusterRole` has this:

```
resourceNames: [secret1 secret2]
```

This results in traefik being granted permission for a secret with the name of `secret 1 secret 2` rather than permission for two separate secrets called `secret1` and `secret2`.

helm

### Additional Information

```shell
This should be an easy change [in the clusterrole.yaml](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/rbac/clusterrole.yaml#L41)

from:


  - apiGroups:
      - ""
    resources:
      - secrets
    {{- if gt (len .Values.rbac.secretResourceNames) 0 }}
    resourceNames: {{ .Values.rbac.secretResourceNames }}
    {{- end }}
    verbs:
      - get
      - list
      - watch


to:

```yaml
  - apiGroups:
      - ""
    resources:
      - secrets
    {{- if gt (len .Values.rbac.secretResourceNames) 0 }}
    {{ with .Values.rbac.secretResourceNames }}
    resourceNames: {{ toYaml . }}
    {{- end }}
    {{- end }}
    verbs:
      - get
      - list
      - watch
```

I'd be open to submitting a PR for this if it would be accepted.
```

### PR DESCRIPTION
**Pull request by justinrush**
### What does this PR do?

The `.Values.rbac.secretResourceNames` setting treated the value as a string which only allowed for a single secret name to be specified. Now it allows multiple values to be specified.

### Motivation

I wanted to restrict traefik's access to multiple secrets but the value was being set as `"[secret1 secret2]"`.

### More

- [ x] Yes, I updated the tests accordingly
- [ x] Yes, I ran `make test` and all the tests passed

closes #1090

### CODE REVIEW

File: traefik/templates/rbac/clusterrole.yaml [hunk 1]
```diff
@@ -37,8 +37,8 @@ rules:
       - ""
     resources:
       - secrets
-    {{- if gt (len .Values.rbac.secretResourceNames) 0 }}
-    resourceNames: {{ .Values.rbac.secretResourceNames }}
+    {{- if .Values.rbac.secretResourceNames }}
+    resourceNames: {{ toYaml .Values.rbac.secretResourceNames |nindent 6 }}
```
mloiseleur, line 41:
```suggestion
    {{- with .Values.rbac.secretResourceNames }}
    resourceNames: {{ toYaml . |nindent 6 }}
```

justinrush: the reason I didn't use `with` is because that adds a `resourceNames: []` stanza. I'm not sure how k8s interprets an empty array versus a nil array and leaving the if in place allows the tests to pass with no changes.

mloiseleur: ah yes. I believed default was `  secretResourceNames: `, but in fact, it's `  secretResourceNames: []`.
Thanks. :+1:

darkweaver87 (approved):
LGTM

traefik/templates/rbac/clusterrole.yaml
mloiseleur: @justinrush I tested it and it works using `with`.

mloiseleur (approved):
Thanks !

justinrush:
@darkweaver87 @mloiseleur what are the next steps for merging this? does the `contributor/waiting-for-corrections` label still need to be there? happy to squash the commits if that's desired.

mloiseleur:
We are good. Thanks @justinrush !

Issue #962, opened 2023-11-14
Provide OCI Artifact chart release


endersonmaia:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

It would be nice to have a OCI Artifact release for the helm charts.

It could be provided via Docker Hub, GitHub Continer Registry, or any other OCI Artifact compatible registry.

```
helm install traefik oci://docker.io/traefik/traefik-chart
```

See: https://helm.sh/docs/topics/registries/

mloiseleur:
That's right. It would be nice :).

x0ddf:
@mloiseleur, as for today, does this [`CHARTS_TOKEN`](https://github.com/traefik/traefik-helm-chart/blob/master/.github/workflows/release.yml#L77) have permission to create a package in the ghcr?

darkweaver87:
@x0ddf and @endersonmaia, you can now install traefik with the command:

```bash
helm install traefik oci://ghcr.io/traefik/helm/traefik
```

Does it solve you issue or you need extra things ?

nlamirault:
works fine @darkweaver87 . You could close i think

PR #1097, opened 2024-06-22
feat(workflow): add oci push

x0ddf:
### What does this PR do?

This PR solves #962 by adding OCI push to the release workflow

### Motivation

Provide the ability to pull the chart from the OCI registry.

### More

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,3 +85,16 @@ jobs:
           commit_username: traefiker
           commit_email: 30906710+traefiker@users.noreply.github.com
         if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
+
+      - name: Publish Helm chart to the ghcr.io registry
+        uses: appany/helm-oci-chart-releaser@v0.4.2
+        with:
+          name: traefik
+          repository:  traefik
+          tag: ${{ steps.chart_version.outputs.CHART_VERSION }}
+          path: ./traefik
+          registry: ghcr.io
+          registry_username: ${{ secrets.GHCR_USERNAME }}
+          registry_password: ${{ secrets.GHCR_PASSWORD }}
+          update_dependencies: 'true' # Defaults to false
+        if: steps.tag_exists.outputs.TAG_EXISTS == 'false'


x0ddf:
@mloiseleur, this PR **requires** to configure two additional secrets for the repository:

- `GHCR_USERNAME` username for ghcr.io
- `GHCR_PASSWORD` [password(PAT) for ghcr.io](https://medium.com/@deepak1812002/get-started-with-github-ghcr-an-alternative-of-dockerhub-f7d5b2198b9a)
---
#### Proof of work: 

- [completed action](https://github.com/x0ddf/traefik-helm-chart/actions/runs/9624858050)

Commands to pull release:
```bash
helm registry login -u ${GHCR_USER} ghcr.io
helm pull oci://ghcr.io/x0ddf/traefik --version 28.3.1
```

.github/workflows/release.yml
```diff
@@ -85,3 +85,16 @@ jobs:
...
+      - name: Publish Helm chart to the ghcr.io registry
+        uses: appany/helm-oci-chart-releaser@v0.4.2
+        with:
+          name: traefik
+          repository:  traefik
+          tag: ${{ steps.chart_version.outputs.CHART_VERSION }}
+          path: ./traefik
+          registry: ghcr.io
+          registry_username: ${{ secrets.GHCR_USERNAME }}
+          registry_password: ${{ secrets.GHCR_PASSWORD }}
```
mloiseleur, line 98:
```suggestion
          registry_username: traefiker
          registry_password: ${{ secrets.CHART_TOKEN }}
```

.github/workflows/release.yml
```diff
@@ -85,3 +85,16 @@ jobs:
...
+          update_dependencies: 'true' # Defaults to false
```
mloiseleur, line 99:
```suggestion
```

mloiseleur (changes requested):
Thanks for taking the time to investigate this !
It looks good. 
See my comments for small changes.

.github/workflows/release.yml
x0ddf: @mloiseleur I agree with `secrets.CHART_TOKEN` but let's keep `registry_username` configurable

mloiseleur (approved):
We will probably have a new release this week. 
=> We'll see soon enough if there is still something missing.

mloiseleur:
cc @jnoordsij , if you want to take a look.

jnoordsij:
Looks all fine to me! Maybe having two different publishing sources might be confusing though, unless there's some way to ensure they're identical.

So in the long run I'd say it be better to use one and migrate accordingly. Given we require Helm > 3.9.0 for the current release, there should be no compatibility issues if we'd migrate to using just OCI.

darkweaver87:
I agree with you @mloiseleur. It has been promoted as stable 2 years ago so let's try it out, use both systems for now and once we are confident, propagate to the rest of our charts.

PR #1099, opened 2024-06-24
feat: ✨ migrate to endpointslices rbac

mloiseleur:
### More


traefik/values.yaml
```diff
@@ -8,7 +8,7 @@ image:
   # -- Traefik image repository
   repository: traefik
   # -- defaults to appVersion
-  tag: ""
+  tag:
```
x0ddf, line 11: Shouldn't it be better to use `~` instead of empty? Also, please update `VALUES.md`

x0ddf (approved):

traefik/values.yaml
mloiseleur: Thanks for your review. I updated `VALUES.md`.
I'm not sure I understand what you mean by using `~`. Would you please detail what you have in mind ?

darkweaver87 (approved):
LGTM

traefik/VALUES.md
```diff
@@ -38,10 +38,10 @@ Kubernetes: `>=1.22.0-0`
 | deployment.additionalContainers | list | `[]` | Additional containers (e.g. for metric offloading sidecars) |
 | deployment.additionalVolumes | list | `[]` | Additional volumes available for use with initContainers and additionalContainers |
 | deployment.annotations | object | `{}` | Additional deployment annotations (e.g. for jaeger-operator sidecar injection) |
-| deployment.dnsConfig | object | `{}` | Custom pod DNS policy. Apply if `hostNetwork: true` dnsPolicy: ClusterFirstWithHostNet |
+| deployment.dnsConfig | object | `{}` | Custom pod [DNS config](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#poddnsconfig-v1-core) |
```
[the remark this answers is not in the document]
mloiseleur: Yes, sure.

traefik/values.yaml
x0ddf: In the YAML, there are many ways to pass the `null` value; one is using the `~` character.

mloiseleur: \o thanks, I was not aware of it.

We are nullifying many values with empty. 
It's more readable, more common in other charts and is ready to be customized when user needs to.

=> So I think we are gonna stick with empty for this PR. 

Feel free to open a proposal if you think this is important.

Issue #1103, opened 2024-06-26
Unable to use additionalServices to setup internal loadbalancer on AKS


arthar360:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

28.3.0

### What version of Traefik are you using?

3.0.2

### What did you do?

I have edited the values.yaml to use additionalServices. I intend to use AKS internal loadbalancer.

```
service:
  enabled: true
  single: true
  type: LoadBalancer
  annotations: {}
  annotationsTCP: {}
  annotationsUDP: {}
  labels: {}
  spec: {}
  loadBalancerSourceRanges: []
  externalIPs: []

  # My Changes
  additionalServices:
    internal:
      type: LoadBalancer
      annotations:
        service.beta.kubernetes.io/azure-load-balancer-internal: "true"
```
All other values in the file are untouched

It should have created an internal load balancer in azure

### What did you see instead?

It shows an error
```
$ helm upgrade --install traefik traefik/traefik -f values.yaml --version 28.3.0
                                                                                                                                   
Error: UPGRADE FAILED: execution error at (traefik/templates/service.yaml:34:6): You need to expose at least one port or set enabled=false to service
```

### What is your environment & configuration?

macOS 14.3.1
Helm: version.BuildInfo{Version:"v3.15.2", GitCommit:"1a500d5625419a524fdae4b33de351cc4f58ec35", GitTreeState:"clean", GoVersion:"go1.22.4"}

### Additional Information

```shell
Not sure if it is a bug or I am doing something wrong. I am currently using chart version 23.0.1 on prod and it works perfectly. 28.3.0 has problems. I have read the breaking changes and applied the breaking changes CRD's accordingly.
```

mloiseleur:
The error message says:
> You need to expose at least one port or set enabled=false to service

And that's right. As you can see in [EXAMPLES.md](https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#add-custom-internal-services), you have to specify in `ports:` section which one you want to expose on this service.

Does that answer your question ?

mloiseleur:
@arthar360 I've opened a PR to improve error message (#1104). Wdyt ?

arthar360:
[...]

PR #1104, opened 2024-06-26
fix: 🐛 improve error message on additional service without ports

mloiseleur:
### Motivation

See https://github.com/traefik/traefik-helm-chart/issues/1103

### More


darkweaver87 (approved):
LGTM

Issue #1132, opened 2024-07-18
helm executing "traefik/templates/gateway.yaml" at <.Values.ports.web.port>: nil pointer evaluating interface {}.port


qianlongzt:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

29.0.1

### What version of Traefik are you using?

3.1.0

### What did you do?

```shell
helm upgrade --install --namespace=traefik --create-namespace   traefik traefik/traefik   --values=cluster/traefik-helm-spec.yaml
```

### What did you see instead?

```shell
Error: UPGRADE FAILED: template: traefik/templates/gateway.yaml:18:22: executing "traefik/templates/gateway.yaml" at <.Values.ports.web.port>: nil pointer evaluating interface {}.port
```

### What is your environment & configuration?

```yaml
image:
  tag: v3.1.0
ports:
  web: null
  websecure:
    port: 443
    expose:
      default: true
    exposedPort: 443
    protocol: TCP
    hostPort: 443
    http3:
      enabled: true
    tls:
      enabled: true
service:
  enabled: false
  type: NodePort
  spec:
    externalTrafficPolicy: Local
hostNetwork: true
additionalArguments:
  - --log.format=json
  - --accesslog
  - --providers.kubernetescrd.allowCrossNamespace=true
deployment:
  kind: Deployment
updateStrategy:
  type: Recreate
  rollingUpdate: null
securityContext:
  capabilities:
    drop: [ALL]
    add: [NET_BIND_SERVICE]
  readOnlyRootFilesystem: true
  runAsGroup: 0 # 
  runAsNonRoot: false
  runAsUser: 0
providers:
  # Disable the Ingress provider (optional)
  # We do not want to use Ingress objects anymore!
  kubernetesIngress:
    enabled: false
  # Enable the GatewayAPI provider
  kubernetesGateway:
    enabled: true
# Allow the Gateway to expose HTTPRoute from all namespaces
gateway:
  namespacePolicy: All

```

darkweaver87:
Hello @qianlongzt ,

Thanks for reporting this.
There are several discussion around gateways, I'll open a PR to satisfy all the needs including yours.

Rémi

Issue #1126, opened 2024-07-15
Default gateway is incompatible with cert-manager


roee88:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

v29.0.1

### What version of Traefik are you using?

3.1.0

### What did you do?

I want to deploy with gateway enabled (default) but integrate it with cert-manager as indicated in the docs:
> Additional gateway annotations (e.g. for cert-manager.io/issuer)

### What did you see instead?

As documented in https://cert-manager.io/docs/usage/gateway/, the hostname and tls.mode fields are required. However, the chart template doesn't specify them.

### What is your environment & configuration?

N/A

darkweaver87:
Hello @roee88 ,

Thanks for reaching out.

Here is a working example.

Create a cluster:

```bash
k3d cluster create traefik --port 80:80@loadbalancer --port 443:443@loadbalancer --k3s-arg "--disable=traefik@server:0"
```

`traefik-values.yaml`:

```yaml
image:
  tag: v3.1.0

ingressRoute:
  dashboard:
    enabled: true

providers:
  kubernetesGateway:
    enabled: true

gateway:
  enabled: true
  annotations:
    cert-manager.io/issuer: selfsigned-issuer
  certificateRefs:
    - name: whoami-tls
```

Install traefik:

```bash
helm repo add traefik https://traefik.github.io/charts  --force-update
helm upgrade --install \
  traefik traefik/traefik\
  --namespace traefik \
  --create-namespace \
  --version v29.0.1 \
 -f values.yaml
```

Install cert-manager:

```bash
helm repo add jetstack https://charts.jetstack.io --force-update
helm upgrade --install \
  cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --create-namespace \
  --version v1.15.1 \
  --set crds.enabled=true \
  --set "extraArgs={--enable-gateway-api}"
```

`whoami-example`:

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whoami
  namespace: traefik
spec:
  replicas: 2
  selector:
    matchLabels:
      app: whoami
  template:
    metadata:
      labels:
        app: whoami
    spec:
      containers:
        - name: whoami
[...]

PR #1133, opened 2024-07-19
fix!: revamp gateway implementation

darkweaver87:
### What does this PR do?

This PR revamps the implementation of gateway object to make it more flexible.

### Motivation

Fixes https://github.com/traefik/traefik-helm-chart/issues/1126.
Fixes https://github.com/traefik/traefik-helm-chart/issues/1132.

### More


roee88:
LGTM. Maybe add a comment in the docs that the port must match ports.web and ports.websecure.

mloiseleur (commented):
Thanks ! This PR looks good :+1:. I will test it.

I agree with @roee88 comment on port matching.

I may add to _fail_ when it happens and to provide an EXAMPLE with both cert-manager and native LE ?

EXAMPLES.md
```diff
@@ -890,3 +890,117 @@ spec:
 Once it's applied, whoami should be accessible on http://whoami.docker.localhost/
 
 </details>
+
+# Use kubernetes Gateway API with Let's encrypt and cert-manager
```
jnoordsij, line 894: I like this example 👍

traefik/templates/gateway.yaml
```diff
@@ -14,25 +17,42 @@ metadata:
...
+      {{- range $portName, $portConfig := $.Values.ports -}}
+        {{- if eq $portConfig.port $config.port -}}
+          {{ $found = true }}
+        {{- end -}}
+      {{- end -}}
+      {{ if not $found }}
+        {{- fail (printf "ERROR: port %0.f is not declared in ports" .port ) }}
+      {{- end -}}
+      port: {{ .port }}
+      {{- if not (has .protocol (list "HTTP" "HTTPS" "TCP" "TLS" "UDP")) }}
```
jnoordsij, line 32: I'm wondering whether or not we should be doing this kind of validations. On one hand, additional validation is good, on the other this requires constant care on our side to match any upstream API changes. Although I do admit this one is unlikely to change and changing it requires little effort, I also wonder how much of a benefit this is in practice.

traefik/templates/gateway.yaml
```diff
@@ -14,25 +17,42 @@ metadata:
...
-      {{- with .Values.gateway.namespacePolicy }}
+      {{- with .namespacePolicy }}
       allowedRoutes:
         namespaces:
           from: {{ . }}
       {{- end }}
+      {{ if or .certificateRefs .mode }}
       tls:
+        {{ with .mode }}
+        {{- if not (has . (list "Passthrough" "Terminate")) }}
```
jnoordsij, line 47: Here I'm even more in favor of removing the validation, given the Gateway API CRD already declares this as an enum, thus resulting in an error when the user tries to set an invalid value. Therefore I'm inclined to say that we should leave validation here to Kubernetes + CRD + API, rather than this chart.

traefik/values.yaml
```diff
@@ -134,14 +134,40 @@ gateway:
...
+      # The port must match a port declared in ports section.
+      port: 8000
+      # -- Optional hostname. See [Hostname](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.Hostname)
+      hostname:
+      # Protocol must be set to HTTP, HTTPS, TCP, TLS or UDP. See [ProtocolType](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.ProtocolType)
+      protocol: HTTP
+      # -- Routes are restricted to namespace of the gateway [by default](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.FromNamespaces
+      namespacePolicy:
+      # -- Add certificates for TLS or HTTPS protocols. See [GatewayTLSConfig](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.GatewayTLSConfig)
+      certificateRefs:
```
jnoordsij, line 154: We should probably drop this value + `mode` value on the `web` port, as it does not intend to use TLS. By adding them to the next one, I think the purpose is clearly documented.

jnoordsij (dismissed):
Great job, I really like the flexibility offered by this! I haven't had the time yet to catch up with all the details of the Gateway API, but I've tried to glance over it for a bit and leave some of my thoughts here!

EXAMPLES.md
mloiseleur, line 894:
```suggestion
# Use Kubernetes Gateway API with cert-manager
```

EXAMPLES.md
```diff
@@ -890,3 +890,117 @@ spec:
...
+--namespace cert-manager \
+--create-namespace \
+--version v1.15.1 \
+--set crds.enabled=true \
+--set "extraArgs={--enable-gateway-api}"
+```
+
+<details>
+
+<summary>With those values, a whoami service can be exposed with a HTTPRoute using HTTPS</summary>
```
mloiseleur, line 936:
```suggestion
<summary>With those values, a whoami service can be exposed with a HTTPRoute on both HTTP and HTTPS</summary>
```

EXAMPLES.md
```diff
@@ -890,3 +890,117 @@ spec:
...
+  tag: v3.1.0
+
+ingressRoute:
+  dashboard:
+    enabled: true
+
+providers:
+  kubernetesGateway:
+    enabled: true
+
```
mloiseleur, line 909:
```suggestion
providers:
  kubernetesGateway:
    enabled: true
```

mloiseleur: 1. We will release in a follow up PR a new release defaulting to v3.1.
2. The dashboard is not related to the subject of this example

EXAMPLES.md
```diff
@@ -890,3 +890,117 @@ spec:
...
+Install cert-manager:
+
+```bash
+helm repo add jetstack https://charts.jetstack.io --force-update
+helm upgrade --install \
+cert-manager jetstack/cert-manager \
+--namespace cert-manager \
+--create-namespace \
+--version v1.15.1 \
```
mloiseleur, line 929:
```suggestion
```

EXAMPLES.md
```diff
@@ -890,3 +890,109 @@ spec:
...
+
+<summary>With those values, a whoami service can be exposed with a HTTPRoute on both HTTP and HTTPS</summary>
+
+```yaml
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: whoami
+  namespace: traefik
```
mloiseleur, line 936:
```suggestion
```

EXAMPLES.md
```diff
@@ -890,3 +890,109 @@ spec:
...
+      containers:
+        - name: whoami
+          image: traefik/whoami
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: whoami
+  namespace: traefik
```
mloiseleur, line 956:
```suggestion
```

EXAMPLES.md
```diff
@@ -890,3 +890,109 @@ spec:
...
+        - name: whoami
+          port: 80
+          weight: 1
+
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: selfsigned-issuer
+  namespace: traefik
```
mloiseleur, line 991:
```suggestion
```

EXAMPLES.md
```diff
@@ -890,3 +890,109 @@ spec:
...
+--namespace cert-manager \
+--create-namespace \
+--version v1.15.1 \
+--set crds.enabled=true \
+--set "extraArgs={--enable-gateway-api}"
+```
+
+<details>
+
+<summary>With those values, a whoami service can be exposed with a HTTPRoute on both HTTP and HTTPS</summary>
```
mloiseleur, line 928:
```suggestion
<summary>With those values, a whoami service can be exposed with HTTPRoute on both HTTP and HTTPS</summary>
```

traefik/values.yaml
```diff
@@ -134,14 +134,36 @@ gateway:
...
   annotations:
+  # -- Define listeners
+  listeners:
+    web:
+      # -- Port is the network port. Multiple listeners may use the same port, subject to the Listener compatibility rules.
+      # The port must match a port declared in ports section.
+      port: 8000
+      # -- Optional hostname. See [Hostname](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.Hostname)
+      hostname:
+      # Protocol must be set to HTTP, HTTPS, TCP, TLS or UDP. See [ProtocolType](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.ProtocolType)
```
mloiseleur, line 149:
```suggestion
      # Specify expected protocol on this listener. See [ProtocolType](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.ProtocolType)
```

traefik/values.yaml
```diff
@@ -134,14 +134,36 @@ gateway:
...
+      protocol: HTTP
+      # -- Routes are restricted to namespace of the gateway [by default](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.FromNamespaces
+      namespacePolicy:
+    websecure:
+      # -- Port is the network port. Multiple listeners may use the same port, subject to the Listener compatibility rules.
+      # The port must match a port declared in ports section.
+      port: 8443
+      # -- Optional hostname. See [Hostname](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.Hostname)
+      hostname:
+      # Protocol must be set to HTTP, HTTPS, TCP, TLS or UDP. See [ProtocolType](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.ProtocolType)
```
mloiseleur, line 159:
```suggestion
      # Specify expected protocol on this listener See [ProtocolType](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.ProtocolType)
```

mloiseleur (approved):
The example worked on my machine.
See my suggestions for non-blocking little improvements to this PR.

PR #1136, opened 2024-07-22
feat: ✨ display release name and image full path in installation notes

mloiseleur:
### What does this PR do?

Display release name and image full path in installation notes

### Motivation

@mmatur discovered it : it always display "Traefik Proxy has been installed", even when using an other image or version compatible with Traefik Proxy, like Traefik Hub.

### More


darkweaver87 (approved):
LGTM

Issue #1147, opened 2024-07-27
Invalid value: "array": tls must be specified for protocols ['HTTPS', 'TLS']


conradwt:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I'm attempting to follow the following walkthrough of using the Gateway API:

https://traefik.io/blog/getting-started-with-kubernetes-gateway-api-and-traefik

When I run 

```zsh
helm upgrade --install --namespace traefik traefik traefik/traefik -f values.yaml
```

### What did you see instead?

```zsh
Release "traefik" does not exist. Installing it now.
Error: 1 error occurred:
	* Gateway.gateway.networking.k8s.io "traefik-gateway" is invalid: spec.listeners: Invalid value: "array": tls must be specified for protocols ['HTTPS', 'TLS']
```

### What version of Traefik are you using?

v3.1.0

### What is your environment & configuration?

```yaml
## File values.yaml ##
providers:
  # Disable the Ingress provider (optional)
  # We do not want to use Ingress objects anymore!
  kubernetesIngress:
	enabled: false
  # Enable the GatewayAPI provider
  kubernetesGateway:
	enabled: true
# Allow the Gateway to expose HTTPRoute from all namespaces
gateway:
  namespacePolicy: All
```

mmatur:
Hi @conradwt,

I moved this issue because it is not related to Traefik but to the helm chart. Someone would take a look to this issue pretty soon.

michalsteyn:
I experienced the same problem for v30 of the helm chart. Downgrading to v29.0.1 resolves the issue. So there is something wrong with v30.

rohanrehman:
It looks like tls value may be missing in the listeners ?
[Here is the traefik-helm-chart values.yaml](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml)

```
gateway:
  # -- When providers.kubernetesGateway.enabled, deploy a default gateway
  enabled: true
  # -- Set a custom name to gateway
  name:
  # -- By default, Gateway is created in the same `Namespace` than Traefik.
  namespace:
  # -- Additional gateway annotations (e.g. for cert-manager.io/issuer)
  annotations:
  # -- Define listeners
  listeners:
    web:
[...]

PR #1154, opened 2024-08-01
fix: disable default HTTPS listener for gateway

darkweaver87:
### What does this PR do?

This PR disables default websecure listener for gateway.

### Motivation

Fixes https://github.com/traefik/traefik-helm-chart/issues/1147.

### More

### DIFF
--- a/traefik/tests/gateway-config_test.yaml
+++ b/traefik/tests/gateway-config_test.yaml
@@ -33,19 +33,6 @@ tests:
       - equal:
           path: spec.listeners[0].allowedRoutes.namespaces.from
           value: "All"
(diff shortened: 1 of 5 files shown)


traefik/tests/gateway-config_test.yaml
```diff
@@ -226,3 +215,40 @@ tests:
     asserts:
       - failedTemplate:
           errorMessage: "ERROR: port 8001 is not declared in ports"
+  - it: should provider listeners by default
```
youkoulayley, line 218:
```suggestion
  - it: should provide listeners by default
```

youkoulayley (approved):

mmatur (approved):
LGTM

PR #1166, opened 2024-08-14
feat: ✨ input validation using schema

mloiseleur:
### More


mloiseleur:
@darkweaver87 @jnoordsij This PR can be drastically improved if we set default empty (and so _typed_) values.

It would mean the comment annotation with type would not be needed anymore.

darkweaver87 (approved):
LVGTM

mloiseleur:
@jnoordsij If you can take a look, it would be great :pray:

jnoordsij (approved):
Oooh I like this a lot, could be very useful to have for validation in IDEs to spot errors more easily. Great work!

Maybe we want to document this for contributors as well? Or even try to have some workflow setup to automatically update the generated schema file? Otherwise probably updating it somewhere in the release process similar to the changelog works fine as well, but still that should be noted somewhere.

mloiseleur:
Ah yes, I totally missed the update / workflow / contributor part. Thanks !

mloiseleur:
@jnoordsij @darkweaver87 I updated Guidelines and added a github action which fails when schema is not updated.

Anything to add / remove / improve before merge ?

jnoordsij (approved):

PR #1171, opened 2024-08-20
fix!: 🐛 set allowEmptyServices to true by default

mloiseleur:
### What does this PR do?

Set `allowEmptyServices` to true by default on Kubernetes CRD & Ingress providers.
It's not needed on Gateway API provider : spec already enforce http error code. 

### Motivation

It will be set by default in the next major version of Traefik Proxy.
See https://github.com/traefik/traefik/issues/10406

### More


darkweaver87 (approved):
LGTM

PR #1180, opened 2024-09-05
fix: 🐛 update CRD to v3.1

mloiseleur:
### What does this PR do?

Update CRD with upstream version for Traefik Proxy v3.1

### Motivation

Even if there is no change of field, this Chart should stay up-to-date with upstream.

### More


darkweaver87 (approved):
LGTM

PR #972, opened 2023-11-24
feat: deployment.podannotations support tpl interpolation

ivankatliarchuk:
### Motivation

Issue https://github.com/traefik/traefik-helm-chart/issues/971

### More

- [x] Yes, I updated the tests accordingly
- [x] Yes, I ran `make test` and all the tests passed

PR #1182, opened 2024-09-10
fix: Allow annotation templates for services that need them #1181

### LINKED ISSUE #1181: Traefik and templating of pod annotations errors with Hashicorp Vault
### Welcome!

31.0.0

v3.1.2

### What did you do?

I'm using Hashicorp Vault injection which uses [annotations](https://developer.hashicorp.com/vault/docs/platform/k8s/injector/annotations) on pods to inject secrets into files on pod startup.  The hope was to use this functionality via the `_FILE` env var in the [credentials](https://go-acme.github.io/lego/dns/cloudflare/#credentials) section as described in the lego docs.  In other words, after putting the secret into Hashicorp Vault, I would like to annotate the traefik pod like so in the helm chart:
```
deployment:
  podAnnotations:
    vault.hashicorp.com/agent-inject: "true"
    vault.hashicorp.com/agent-inject-secret-cloudflare: kv/data/k8s/traefik/cloudflare_token
    vault.hashicorp.com/agent-inject-template-cloudflare: |
      {{- with secret "kv/data/k8s/traefik/cloudflare_token" -}}
      {{ index .Data.data "api-token" }}
      {{- end }}
    vault.hashicorp.com/role: traefik
```
This would inject a file into `/vault/config/cloudflare` that I could then reference via setting an env var, `CF_DNS_API_TOKEN_FILE` to that value and keep it up-to-date with Vault.  The expectation, like other popular charts, is that the above values would render.

### What did you see instead?

The chart errors out during templating:
> Error: template: traefik/templates/_podtemplate.tpl:6:12: executing "traefik.podTemplate" at <tpl (toYaml .Values.deployment.podAnnotations) .>: error calling tpl: cannot parse template "vault.hashicorp.com/agent-inject: \"true\"\nvault.hashicorp.com/agent-inject-secret-cloudflare: kv/data/k8s/traefik/cloudflare_token\nvault.hashicorp.com/agent-inject-template-cloudflare: |\n  {{- with secret \"kv/data/k8s/traefik/cloudflare_token\" -}}\n  {{ index .Data.data \"api-token\" }}\n  {{- end }}\nvault.hashicorp.com/role: traefik": template: gotpl:4: function "secret" not defined

To reproduce:
```
helm pull traefik/traefik --untar
cd traefik 
cat <<EOF>values-annotations.yaml
deployment:
  podAnnotations:
    vault.hashicorp.com/agent-inject: "true"
    vault.hashicorp.com/agent-inject-secret-cloudflare: kv/data/k8s/traefik/cloudflare_token
    vault.hashicorp.com/agent-inject-template-cloudflare: |
      {{- with secret "kv/data/k8s/traefik/cloudflare_token" -}}
      {{ index .Data.data "api-token" }}
      {{- end }}
    vault.hashicorp.com/role: traefik
> EOF
helm template . --values values.yaml --values values-annotations.yaml
``` 

Platform: microk8s  v1.27.16
Arguments: --values
YAML: 
```
deployment:
  podAnnotations:
    vault.hashicorp.com/agent-inject: "true"
    vault.hashicorp.com/agent-inject-secret-cloudflare: kv/data/k8s/traefik/cloudflare_token
    vault.hashicorp.com/agent-inject-template-cloudflare: |
      {{- with secret "kv/data/k8s/traefik/cloudflare_token" -}}
      {{ index .Data.data "api-token" }}
      {{- end }}
    vault.hashicorp.com/role: traefik
```
Chart version: 31.0.0
appVersion: v3.1.2
helm version: v3.15.4

### Additional Information
Right now I see two options: the clean option or the safe option. For the clean option, the `tpl` function [here](https://github.com/traefik/traefik-helm-chart/blob/2dc2253c2e360d2ce2f3fd0ed8407e89b4a0c273/traefik/templates/_podtemplate.tpl#L6) is removed entirely and swapped to `with` but that may have unintended consequences. For the safe option, there is a new boolean introduced and defaulted to current path of using the `tpl` function while providing a switch, `templateAnnotations` below, to disable it:
 ```
# templates/_podtemplate.tpl
  {{- $version := include "imageVersion" $ }}
    metadata:
      annotations:
      {{- if .Values.deployment.podAnnotations }}
        {{- if .Values.deployment.templateAnnotations }}
          {{- tpl (toYaml .Values.deployment.podAnnotations) . | nindent 8 }}
        {{- else }}
          {{- with .Values.deployment.podAnnotations }}
          {{- toYaml . | nindent 8 }}
          {{- end }}
        {{- end }}
```
With an the default set as the existing behavior:
```
# values.yaml
deployment:
   templateannotations: true
```
For the PR I would assume going with the safe option but I would defer to one of the traefik folks if there are opinions.  Thanks in advance.

brow86:
### What does this PR do?
This fixes #1181 by doing the following:
- Adds a  `templateAnnotations` value to default values file and keeps it as `true` to keep existing behavior 
- Updates deployment template to leverage the new boolean
- Add true/false test cases

### Motivation
This should help avoid issues with Hashicorp Vault and agent injection and any other service which uses templates as a part of its annotations.

### More

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -3,7 +3,13 @@
     metadata:
       annotations:
       {{- if .Values.deployment.podAnnotations }}
-        {{- tpl (toYaml .Values.deployment.podAnnotations) . | nindent 8 }}
+        {{- if .Values.deployment.templateAnnotations }}
+          {{- tpl (toYaml .Values.deployment.podAnnotations) . | nindent 8 }}
+        {{- else }}
+          {{- with .Values.deployment.podAnnotations }}
+          {{- toYaml . | nindent 8 }}
+          {{- end }}
+        {{- end }}
       {{- end }}
       {{- if .Values.metrics }}
       {{- if and (.Values.metrics.prometheus) (not (.Values.metrics.prometheus.serviceMonitor).enabled) }}
--- a/traefik/tests/deployment-config_test.yaml
+++ b/traefik/tests/deployment-config_test.yaml
@@ -93,6 +93,33 @@ tests:
       - equal:
           path: spec.template.metadata.annotations.traefik/type
           value: 'internal'
+  - it: should error when annotations themselves contain a template
+    set:
+      deployment:
+        podAnnotations:
+          vault.hashicorp.com/agent-inject-template-test: |
+            {{- with secret 'kv/data/path/to/secret' -}}
+            {{ index .Data.data "secret" }}
+            {{- end }}
+    asserts:
+      - failedTemplate:
+          errorPattern: "[cannot parse template]+"
+  - it: should not error when templated annotations have templating disabled
+    set:
+      deployment:
+        podAnnotations:
+          vault.hashicorp.com/agent-inject-template-test: |
+            {{- with secret 'kv/data/path/to/secret' -}}
+            {{ index .Data.data "secret" }}
+            {{- end }}
+        templateAnnotations: false
+    asserts:
+      - equal:
+          path: spec.template.metadata.annotations["vault.hashicorp.com/agent-inject-template-test"]
+          value: |
+            {{- with secret 'kv/data/path/to/secret' -}}
+            {{ index .Data.data "secret" }}
+            {{- end }}
   - it: should have labels with specified values
     set:
       deployment:
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -52,6 +52,8 @@ deployment:
   # It supports templating. One can set it with values like traefik/name: '{{ template "traefik.name" . }}'
   podAnnotations: {}
   # -- Additional Pod labels (e.g. for filtering Pod by custom labels)
+  templateAnnotations: true
+  # -- Additional Pod labels (e.g. for filtering Pod by custom labels)
   podLabels: {}
   # -- Additional containers (e.g. for metric offloading sidecars)
   additionalContainers: []


mloiseleur:
I understand your needs. I'm not sure of the correct way to fix it and it seems a really specific needs.

The tpl has been introduced last year with https://github.com/traefik/traefik-helm-chart/pull/972 from @ivankatliarchuk . 

Don't you have a way to "print" template code with `tpl` ?

brow86:
@mloiseleur - Point taken on this being a specific use case but I can see this potentially being a problem with any other service that does template interpolation with annotations which-- moving forward-- may be a lot more than just this one case.  In other words, I'd be shocked if this doesn't break something else moving forward.  As for solving it, the PR introduces a boolean to flip it off and defaults it to the default behavior of running `tpl`, does that work? 

The only other thing I'll mention is that we use this on a lot of our helm charts and traefik is the first chart I've run into that templates the values out-of-the-box.  Chart templates on other popular repos like [grafana](https://github.com/grafana/helm-charts/blob/614fe8997c2a2cdd6e63b2c0e8816e76b875711f/charts/grafana/templates/deployment.yaml#L12) and the [nginx](https://github.com/kubernetes/ingress-nginx/blob/0111961e7d79b386c7717fa392d727e8fdadfc2a/charts/ingress-nginx/templates/default-backend-deployment.yaml#L30) don't template out-of-the-box so, while I agree it is helpful and should be kept in, many other charts don't currently do this.

brow86:
@mloiseleur - Wanted to poke this again.  Is the purposed PR not viable?  Is there another suggested approach?

brow86:
> AFAIK, without any PR, one can set its template using a Configmap with Vault and agent injection.

I guess the question then is whether the following changes takes this from being simple to complex:

- one extra variable in the values file
- an if/else in the template value
- two additional test cases for the change

This doesn't feel like a heavy change but do those changes break the philosophy?  More importantly: could that same logic be used to argue that the original change to template annotations broke that mold?  Looking at other popular helm chart annotations in the community, it does not look like this is the case.

mloiseleur:
I added `no-merge` label, in order to see if there is an interest about this use case.

mloiseleur:
There has been no interest from other community members on this feature since last year, so I close this pr.
Feel free to re-open it or open a new one if needed.

PR #1197, opened 2024-09-20
Add loadbalancerip support

ColonelBundy:
### What does this PR do?
Add support for specifying loadblanacerip

### Motivation
Some cni's does not support externalIps field but rather the loadbalancerIp field.


ColonelBundy:
Can be added using spec.loadbalancerip 

No longer necessary.

PR #1199, opened 2024-09-26
Replace `CLF` with `common` in `values.yaml`

WillDaSilva:
I tried using `CLF` as the docs suggested for the access log format, but I received the following error:

```
2024-09-26T21:36:25Z ERR github.com/traefik/traefik/v3/pkg/middlewares/accesslog/logger.go:98 > Unsupported access log format: "CLF", defaulting to common format instead.
```


darkweaver87 (approved):
Thanks for your PR. You're right, for the record, it's defined [here](https://github.com/traefik/traefik/blob/v3.1.4/pkg/middlewares/accesslog/logger.go#L35). 
LGTM

PR #1200, opened 2024-09-27
Change apiVersion to updated group

NeuronButter:
### What does this PR do?

Changes an old reference to the `traefik.containo.us` API Group in EXAMPLES.md.

### Motivation

The new API group has been in use for a long while now, seems like it was just missed in EXAMPLES.md.

### More

(marked as done but irrelevant)


mloiseleur (approved):
\o we missed it ! Thanks :+1:

PR #1210, opened 2024-10-09
Update topology spread constraints comments

jonathanbeber:
### What does this PR do?

This PR updates it to use the `app.kubernetes.io/name` label which is a default label in `traefik.labelselector`.

### Motivation

The help text for the topology spread constraints in the values file was misleading since the `app` label is not a default.

### More

- [ ]Yes, I updated the tests accordingly

I don't think any apply.


darkweaver87 (approved):
LGTM.

Thanks for your contribution @jonathanbeber :+1:

PR #1213, opened 2024-10-10
fix!(Env Variables): allow extending env without overwrite

### LINKED ISSUE #958: Allow extending env value more easily
### Welcome!

Currently the `env` key in the values file is a list. This means that it is not easily extendable, since Helm will overwrite list values when set, rather than merging them. This means that when one wants to pass an additional env variable, e.g. `TZ=US/Alaska`, the entire list contained in this repo has to be copied and passed with it, or else will get lost:
```
  env:
    # This has to be copied (current values.yaml content)
    - name: POD_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.name
    - name: POD_NAMESPACE
      valueFrom:
        fieldRef:
          fieldPath: metadata.namespace
    # Custom additions can then be placed afterwards
    - name: TZ
      value: US/Alaska
```

It would be nice to have another value type, e.g. key-based, for the environment, so Helm can properly merge it. This can then be parsed into the correct format within the relevant templates.

Suggested implementation: the approach on the following StackExchange question might be a good place to start:
https://stackoverflow.com/questions/59394422/helm-charts-with-multiple-lists-in-multiple-values-files

mloiseleur:
### What does this PR do?

Move expected env variable directly into the template.
With this PR, only additional env variable can be set with _values_.

### Motivation

Fixes #958

### More


darkweaver87 (approved):
This also removes the capability for users to remove completely the env variables we set by default. Considering our defaults, it looks reasonable to me.
LGTM

PR #1214, opened 2024-10-10
fix(certificateResolvers)\!: :boom: use same syntax in Chart and in Traefik

mloiseleur:
### What does this PR do?

Align syntax used in this Chart and in Traefik for certificateResolvers.

Before

```yaml
certResolvers:
  myAcmeResolver:
    email: email@example.com
    dnsChallenge:
      provider: myProvider
      resolvers:
        - 1.1.1.1
        - 8.8.8.8
    tlsChallenge: true
```

After

```yaml
certificatesResolvers:
  myAcmeResolver:
    acme: 
      email: email@example.com
      dnsChallenge:
        provider: myProvider
        resolvers:
          - 1.1.1.1
          - 8.8.8.8
      tlsChallenge: true
```

### Motivation

Nowadays, Traefik Proxy and Traefik Hub supports various certificateResolvers, not only ACME. 
Current implementation do not allow to configure `tailscale` or `distributedAcme`, for instance.

### More

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -331,14 +331,15 @@ Here is a more complete example, using also native Let's encrypt feature of Trae
 persistence:
   enabled: true
   size: 128Mi
-certResolvers:
+certificatesResolvers:
   letsencrypt:
-    email: "{{ letsencrypt_email }}"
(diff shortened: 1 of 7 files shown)


darkweaver87 (approved):
LGTU

PR #1217, opened 2024-10-11
fix(schema): :bug: targetPort can also be a string

darkweaver87:
### What does this PR do?

This PR makes it possible to configure `targetPort` with a named port.

### Motivation

Being compliant with [kubernetes specs](https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports).

### More


mloiseleur (approved):
LVTM

PR #1219, opened 2024-10-11
feat(Gateway API): add infrastructure in the values

mloiseleur:
### More


darkweaver87 (approved):
LGTM

ihmels:
I think the new `infrastructure` section has no effect as Traefik does not support `GatewayInfrastructurePropagation`.

Issue #1209, opened 2024-10-08
Don’t bundle Gateway API CRDs


genebean:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

I’m doing some greenfield work and trying to take advantage of all the Gateway API goodness Traefik, Argo CD, Argo Rollouts, Linkerd, and others are doing, but have found that Traefik bundling in the Gateway API CRDs in the helm charts is both an anti pattern and problematic. The other projects I’ve looked at all seem to have you pull them in as a prerequisite. This allows the end user to pull in whatever works best for the combination of tools deployed. 

For example, here’s what prompted this issue: I have a cluster with
- Traefik + external-dns + cert-manager for north/south traffic and 
- Linkerd + cert-manager for east/west traffic and
- Argo CD managing everything

Traefik & Argo CD both have Gateways that use certs from cert-manager, and Argo CD is using HTTPRoutes. 

I’m also working to deploy an application stack that will utilize Linkerd + GRPCRoutes inside the cluster for observability and gRPC load balancing. 

The problem with the bundled CRDs is that [Linkerd still uses Gateway API v1alpha2](https://github.com/linkerd/linkerd2/issues/13032) for compatibility with Google Kubernetes Engine (GKE).

Thus, to get a version of Linkerd with the features I need (including support for [Native Sidecars](https://linkerd.io/2.16/tasks/graceful-shutdown/#graceful-shutdown-of-job-and-cronjob-resources) so that Jobs like the ones in Argo CD itself work correctly) and Traefik to work together, I am currently forced to use a custom build of the Traefik helm chart with the Gateway API manifest replaced with the experimental 1.1.0 version as the stable version doesn’t expose the v1alpha2 variant (just beta and stable ones). 

https://github.com/traefik/traefik-helm-chart/issues/1141 is one possible solution to this if setup like https://github.com/linkerd/linkerd2/tree/main/charts/linkerd-crds as they allow toggling off the bundled Gateway API CRDs by setting `enableHttpRoutes: false` like mentioned at https://linkerd.io/2.16/features/httproute/ 

Lastly, for reference, I posted about this at https://community.traefik.io/t/install-traefik-w-o-gateway-api-crds/24712/ too.

paul-at-cybr:
[...]

Issue #1141, opened 2024-07-25
CRD only Helm chart for Traefik


brian-provenzano:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

It would nice if Traefik had a separate helm chart just for CRDs so they could be easily managed by tooling like ArgoCD. Many other projects have started doing this (e.g. [karpenter](https://github.com/aws/karpenter-provider-aws/tree/main/charts/karpenter-crd)) and I think it would be benficial for other ArgoCD users.  Thanks!

genebean:
Linkerd does this too.

PR #1223, opened 2024-10-14
feat: :package: add separated chart for CRDs

darkweaver87:
### What does this PR do?

This PR introduces a separate chart used by the main one to install CRDs.

### Motivation

https://github.com/traefik/traefik-helm-chart/issues/1141
https://github.com/traefik/traefik-helm-chart/issues/1209

### More


traefik-crds/templates/crds.yaml
```diff
@@ -0,0 +1,42 @@
...
+"crds-files/traefik/traefik.io_ingressroutes.yaml"
+"crds-files/traefik/traefik.io_ingressroutetcps.yaml"
+"crds-files/traefik/traefik.io_ingressrouteudps.yaml"
+"crds-files/traefik/traefik.io_middlewares.yaml"
+"crds-files/traefik/traefik.io_middlewaretcps.yaml"
+"crds-files/traefik/traefik.io_serverstransports.yaml"
+"crds-files/traefik/traefik.io_serverstransporttcps.yaml"
+"crds-files/traefik/traefik.io_tlsoptions.yaml"
+"crds-files/traefik/traefik.io_tlsstores.yaml"
+"crds-files/traefik/traefik.io_traefikservices.yaml"
```
mloiseleur, line 13: reminder: Check if we can avoid this list and use the directory directly

.github/workflows/release.yml
```diff
@@ -27,18 +29,18 @@ jobs:
...
+      - name: Copy LICENSE, EXAMPLES.md and README.md for packaging
         run: |
           cp ./README.md ./traefik/README.md
           cp ./EXAMPLES.md ./traefik/EXAMPLES.md
           cp ./LICENSE ./traefik/LICENSE
 
       - name: Generate default static install
         run: |
-          kustomize build traefik/crds > traefik.yaml
+          kustomize build traefik-crds > traefik.yaml
```
mloiseleur, line 40: Locally, this line exit with this msg:
> Error: must build at directory: not a valid directory: evalsymlink failure on 'traefik-crd' : lstat /home/mloiseleur/git/traefik-helm-chart/traefik-crd: no such file or directory

.github/workflows/release.yml
```diff
@@ -6,12 +6,14 @@ on:
       - master
 
 env:
-  tag_prefix: v
+  traefik_tag_prefix: traefik_v
+  traefik-crds_tag_prefix: traefik-crds_v
```
mloiseleur, line 10:
```suggestion
  tag_prefix: v
  crds_tag_prefix: crds_v
```

.github/workflows/release.yml
```diff
@@ -6,12 +6,14 @@ on:
...
 
 jobs:
   test:
     uses: "traefik/traefik-helm-chart/.github/workflows/test.yml@master"
-  release:
+
+  release-traefik:
```
mloiseleur, line 16:
```suggestion
  traefik:
```

.github/workflows/release.yml
```diff
@@ -97,3 +99,82 @@ jobs:
           registry_username: traefiker
           registry_password: ${{ secrets.GHCR_TOKEN }}
         if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
+
+  release-traefik-crds:
```
mloiseleur, line 103:
```suggestion
  traefik-crds:
```

traefik/crds/kustomization.yaml
```diff
@@ -1,24 +0,0 @@
...
-  - hub.traefik.io_apibundles.yaml
-  - traefik.io_ingressroutes.yaml
-  - traefik.io_ingressroutetcps.yaml
-  - traefik.io_ingressrouteudps.yaml
-  - traefik.io_middlewares.yaml
-  - traefik.io_middlewaretcps.yaml
-  - traefik.io_serverstransports.yaml
-  - traefik.io_serverstransporttcps.yaml
-  - traefik.io_tlsoptions.yaml
-  - traefik.io_tlsstores.yaml
```
mloiseleur, line 23: This file should not be removed. 
It's the current working & documented way to upgrade CRDs.

mloiseleur (changes requested):
1. The CI needs to be tested, with a tag on your fork.
2. The documented CRDs upgrade process should still work. I see (at least) two options:
  a. Keep and copy crd files 
  b. Update CRD upgrade instruction to use them in the new directory

.github/workflows/release.yml
darkweaver87: It works. Just don't do any typo while running the command :-): traefik-crd**s**

traefik/crds/kustomization.yaml
darkweaver87: I documented it.

darkweaver87:
Tested on my own account :-)

.github/workflows/changelog.json
```diff
@@ -0,0 +1,53 @@
+{
+  "template": "## [#{{TO_TAG}}](#{{RELEASE_DIFF}}) (${DATE})\n#{{CHANGELOG}}",
+  "categories": [
+    {
+      "title": "## :boom: Breaking changes",
```
mloiseleur, line 5:
```suggestion
      "title": "## :boom: BREAKING CHANGES",
```

.github/workflows/changelog.json
```diff
@@ -0,0 +1,53 @@
...
+      "rules": [
+        {
+          "pattern": "fix\\((${REGEXP}).*\\):.*",
+          "on_property": "title",
+          "flags": "gu"
+        }
+      ]
+    },
+    {
+      "title": "## :package: Other",
```
mloiseleur, line 35:
```suggestion
      "title": "## :package: Others",
```

.github/workflows/release.yml
```diff
@@ -61,16 +70,40 @@ jobs:
...
         with:
           tag: ${{ steps.tag_version.outputs.new_tag }}
           name: ${{ steps.tag_version.outputs.new_tag }}
-          body: ${{ steps.tag_version.outputs.changelog }}
+          body: ${{ steps.changelog.outputs.changelog }}
           prerelease: ${{ contains(steps.chart_version.outputs.CHART_VERSION, '-') }}
           artifacts: "traefik.yaml"
         if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
 
+      # avoid helm-gh-pages to push both charts
```
mloiseleur, line 101:
```suggestion
      # avoid to push both charts
```

.github/workflows/release.yml
```diff
@@ -97,3 +130,113 @@ jobs:
...
+      - name: Create release
+        uses: ncipollo/release-action@v1
+        with:
+          tag: ${{ steps.tag_version.outputs.new_tag }}
+          name: ${{ steps.tag_version.outputs.new_tag }}
+          body: ${{ steps.changelog.outputs.changelog }}
+          prerelease: ${{ contains(steps.chart_version.outputs.CHART_VERSION, '-') }}
+        if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
+
+      # avoid helm-gh-pages to push both charts
```
mloiseleur, line 211:
```suggestion
      # avoid to push both charts
```

mloiseleur (approved):
I have just minor naming suggestion left.
Thanks, it LVGTM :+1:

mloiseleur:
@jnoordsij If you think we should not do it, it's still time and possible to explain why !

jnoordsij:
I definitely have some mixed thoughts about it. While I totally see the issue with CRDs and Helm and the reason this might call for some different structure, that also implies that probably we should tread carefully in finding a solution that works around Helm limitations.

To be more concrete: I can agree with using a separate chart for CRDs, but I'd rather make it entirely stand-alone than add it as a dependency to the main chart. Reasoning:
1. By installing it as a dependency, it is now subject to all the possible issues that may arise when using Helm for managing CRDs: rollbacks or uninstalls of the chart may cause CRDs to be altered/removed in such a way that global cluster behavior might be impacted. This is particularly problematic for cases where one wants to install multiple instances of Traefik within the same cluster.
2. Creating a separate chart and then adding it as a dependency seems odd, as this makes it harder to manage them separately for end-users, given all operations will influence both. 
3. Managing it as a dependency is somewhat of a burden (minor issue)
4. Helm docs seem to suggest separate charts as a way to deal with CRDs, but from what I understand from the wording they (probably for above reasoning) suggest to have them installed separately: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts.
5. Thus we somewhat go against Helm conventions, which may be confusing to some users, although probably these users know enough about CRDs already to tackle the issue.

Some other alternatives we may opt for are:
1. Keeping things as is and improving documentation on CRDs and upgrading; probably quite tricky given how hard are CRDs to deal with and the amount of complexity involved in explaining the issues (see https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you).
2. Rather than create the overhead of a dependency chart, move the crds from the Helm-default `crd` directory to templates and install from there. Then add a resource-policy to the CRDs to prevent accidental  This allows for upgrading them through the chart without the burden/complexity of an entire new chart. I've seen this pattern to be used in [the `cert-manager` Helm chart](https://cert-manager.io/docs/installation/helm/#uninstalling-with-helm), which might serve as a working example employed elsewhere.
3. Removing CRD management with Helm entirely and let users manually manage them all the way.

I think the most important trade-off here is how much knowledge/action is required from the end-user vs. potential risks by simplifying things for them, but in turn making some seemly harmless operations actually unsafe. The usecase which is probably most impacted by any change is whenever there's multiple instance of the chart being installed in the cluster. Given this is an advanced usecase we can probably expect end-users employing this pattern to generally know quite a bit more about Helm/CRDs and thus focus on the "simple" usecase in terms of end-user convenience, as long as we make sure the choice is documented clearly for those with a more advanced installation.

All in all I think my preference would be the following:
1. Create the separate chart for managing CRDs, but NOT as a dependency
    - For basic usage, the installation instruction would then be install the CRD chart, then install the main chart; then upgrading instructions would be first update your CRD chart, then update the regular chart. This requires some additional action for basic usage, but limited to 'simple' Helm operations so probably ok?
    - This prevents accidents with operations applied while altering/upgrading the main chart. 
    - For advanced usage, we can simply tell people: use the CRD chart if you want to manage it in that way; manually install them if you prefer that, otherwise just don't and manage on your own. This is quite easy and more or less corresponds to current behavior
2. Add the CRDs as templates within the chart like in the cert-manager chart
    - This pattern is used elsewhere already and thus has some real-world proof of being applied
    - Probably the most easy for end-users without knowledge about CRDs, while potential breakages are limited by the resource policy
    - For advanced usage we just need to document the flag that disables this and leaves management to the user. However they can then not really benefit from updates added here.
3. Keep things as is, but find a way to better reach users on how to update CRDs
    - I think CRDs are just slightly too complex as a concept, especially with Helm, and this puts too much burden on end-users with a simple setup
4. Add separate chart for managing CRDs as a dependency
    - Still possible, however I think the options above are just better alternatives
5. Removing CRD management with Helm entirely and let users manually manage them all the way
    - Basically option 3. but even more complex, so this probably helps no-one 

Final concern: for all cases, I've not yet checked how upgrading will go from the current system towards the new one. Have you tested this yet? I'm not sure if/how Helm can deal with CRDs that have previously been installed by the main chart, and might suddenly be claimed by the new CRD chart after applying this. Similarly issues may arise in other options.

(This has become a bit longer that I intended, hopefully I managed to get my main concerns addressed. Make sure to ask me on anything unclear, as I understand this might raise quite a few questions.)

mloiseleur:
Damn it, that's right. It will break many configuration with the dependency :warning: !
And without the dependency, the install instruction needs to be changed, it won't work out of the box. 
Even simple blog post with helm install will fail :(.

I'm quite interested in option 2, with doubts ; I'm unsure if the CRDs can be deployed _before_ being used by the dashboard or the healthcheck `IngressRoute`.
As you said, it is the most easy for the users, and that's what we want. 
We want to keep this and allow advanced users / new features like Gateway API to be more flexible.

=> I'll make some tests on this and open a PR if it looks good.

Many thanks for this review :pray:

darkweaver87:
To be honnest, I didn't test to install the chart with CRDs and then upgrade with the one with dependencies. It's very important to validate it.

For the rest of the concerns, I can understand it but let's summarize the use cases. I see 3 scenarios:
1. people who just want a simple installation and rely on us: this is currently a nightmare to handle CRDs
2. people who have a custom installation and just want to install CRDs in a simpler manner
3. people who have multiple instances of this chart and want to manage CRDs on one hand and the resources

For this, we have multiple solutions, adding your propositions:
- doing nothing. I can't even think about this. As a user I would prefer seeing people developing the software I use trying to solve my issues, failing and iterating rather than doing nothing for me :-)
- having crds in a `crd` directory might work. I've checked the ordering and it's good. It can address all scenarios but it will require from us somehow to add big `{{ if .Values.addCRDs }}` in all our templates which is, I think, not elegant at all
- create a separate chart for CRDs but not as dependency. This will not cover use case 1. Also, the won't ensure people to get the right version of the CRDs along with the current chart version.
- create a separate chart for managing CRDs as a dependency. We will lose people. I've seen to many projects where a lot of charts have been created, without deprecation notices,  and I didn't know which one to choose
- create a separate chart for CRDs as a dependency like this PR propose. This will cover use case 1, 2 and 3. For use case 2 and 3, the dependency will still be here but having all boolean to false will ensure CRDs not being installed. Also, we will have the capability to specify which version goes with a given chart version (like the  `crd` directory solution).

traefik-crds/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: traefik-crds
+description: A Traefik based Kubernetes ingress controller
+type: application
+version: 0.0.1
```
mloiseleur, line 5:
```suggestion
version: 3.2.0-rc1
```

traefik-crds/values.yaml
```diff
@@ -0,0 +1,4 @@
+traefik: false  # @schema type:[boolean]
```
mloiseleur, line 1:
```suggestion
traefik: false 
```

traefik-crds/values.yaml
```diff
@@ -0,0 +1,4 @@
...
+gateway_api: false  # @schema type:[boolean, string]; default: false
+hub: false  # @schema type:[boolean, string]; default: false
+deleteOnUninstall: false  # @schema type:[boolean]
```
mloiseleur, line 4:
```suggestion
deleteOnUninstall: false
```

traefik/Chart.yaml
```diff
@@ -29,3 +29,7 @@ annotations:
     - "feat(deps): update traefik docker tag to v3.1.5"
     - "feat(Traefik Proxy): update rbac following v3.2 migration guide"
     - "chore(release): 🚀 publish v32.1.0"
+#dependencies:
+#  - name: traefik-crds
+#    version: 0.0.1
+#    repository: https://traefik.github.io/charts
```
mloiseleur, line 35: The dependency does not work. I have encountered two issues when testing:
1. schema error on global: 
```less
Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
traefik-crds:
- (root): Additional property global is not allowed
```

2. I allowed additional property in the schema, and another error on values:

```less
helm.go:86: 2024-10-29 16:11:08.371483008 +0100 CET m=+0.316145995 [debug] values don't meet the specifications of the schema(s) in the following chart(s):
traefik:
- traefik-crds.gateway_api: Invalid type. Expected: string, given: boolean
- traefik-crds.hub: Invalid type. Expected: string, given: boolean
```

traefik-crds/values.yaml
darkweaver87: Let's keep the schema validation.

traefik-crds/values.yaml
darkweaver87: Let's keep the schema validation.

traefik/Chart.yaml
darkweaver87: fixed by https://github.com/traefik/traefik-helm-chart/pull/1223/commits/b79e8efbb56815a7e5e22ae8807fc159a9a50323

traefik-crds/Chart.yaml
darkweaver87: I don't know which versioning we should use. But traefik one, doesn't seem to be a good idea as we aren't installing traefik only CRDs.

traefik/Chart.yaml
mloiseleur: I tested the upgrade and it failed with:

```less
Error: UPGRADE FAILED: Unable to continue with update: CustomResourceDefinition "ingressroutes.traefik.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "traefik"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "traefik"
```

To reproduce:

1. Lauch k3d: `k3d cluster create traefik-hub --port 80:80@loadbalancer --port 443:443@loadbalancer --port 8000:8000@loadbalancer --k3s-arg "--disable=traefik@server:0"`
2. Install Traefik with v32.1.0: `helm install traefik -n traefik --wait   --version v32.1.0   --set ingressClass.enabled=false   --set ingressRoute.dashboard.enabled=true   --set ingressRoute.dashboard.matchRule='Host(`dashboard.docker.localhost`)'   --set ingressRoute.dashboard.entryPoints={web}   --set ports.web.nodePort=30000   --set ports.websecure.nodePort=30001    traefik/traefik`
3. :heavy_check_mark: Check GW API version: `kubectl describe crds httproutes.gateway.networking.k8s.io | head`
4. Uncomment dependency in Chart.yaml with `repository: file://../traefik-crds`
5. Uncomment bottom dependency values (L953-959) about this traefik-crds
6. build dependency: helm dependency build
7. Upgrade and enable GW API provider with local new version: `helm upgrade traefik -n traefik --wait  --set ingressClass.enabled=false   --set ingressRoute.dashboard.enabled=true   --set ingressRoute.dashboard.matchRule='Host(`dashboard.docker.localhost`)'   --set ingressRoute.dashboard.entryPoints={web}   --set ports.web.nodePort=30000   --set ports.websecure.nodePort=30001 --set providers.kubernetesGateway.enabled=true   ./traefik`

darkweaver87: Please read upgrade documentation: https://github.com/traefik/traefik-helm-chart/pull/1223/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R113

mloiseleur:
> Add the CRDs as templates within the chart like in the cert-manager chart

This morning, I have explored this option. I have encountered two major issues:

1. On **install**: It fails to deploy dashboard `IngressRoute` (CRD is not loaded at this stage)
2. On **upgrade**: It fails to deploy CRD: invalid ownership on metadata and annotations

We are really fighting the tide, here.
I haven't found a way to switch CRDs from current Helm CRD management to Helm Template management without failure.

This use case of selectively picking CRDs is specific. It's clearly not the common use case.

I think we are left with two options on this:

1. Option _Standalone CRD Chart_: 
   * The common use case works is not modified (current Chart stays as it is today)
   * Users has to _opt-in_ to use this new Chart
   * There is no link between them.
   * Like karpenter, we provide two Charts, the current one with CRDs and a new one with only CRDs. 
2. Option _Documentation_: 
   * Instructions in the README on how to install CRDs separately
   * Users with specifics needs on CRDs deploy them separately and can install the Chart without CRDs after

@jnoordsij @darkweaver87 : Wdyt ?

mloiseleur:
Additional notes: 
Helm maintainers have started to work on Helm v4, see [here](https://github.com/helm/community/blob/main/hips/hip-0012.md).
CRDs management are [discussed in the roadmap](https://docs.google.com/document/d/1WJ3K96fJeldKHoKhejWHDvCOTddEvY-RCtQBUaZ57FM/edit?tab=t.0).
Helm maintainers have also detailed why it's working this way in [HIP-0011](https://github.com/helm/community/blob/main/hips/hip-0011.md).

jnoordsij:
I think I'm still with option 1 like I was before when explaining my thoughts. CRDs in Helm are just quite complicated at the moment, so I don't think there is a solve-all solution without adjustments on the Helm side. Although managing two charts at once might be complicated for users, if we document matching versions it hopefully should not be to complicated and will provide a way of updating CRDs that is a lot easier then users having to do it manually.

darkweaver87:
I'm not strongly against option 1, but I think we will clearly lost users on the way. The approach of having a dependency chart seduced me at first because of managing all this compatibility complexity being a responsibility delegated to us. 
With option 1, we will somehow, have to complete a compatibility matrix for users to know which version of CRD chart comes with which chart version or traefik/traefik-hub version.

The other option is keeping it that way:
- people who don't want to bother with it and let us manage their CRDs for them will just need to set:
```
traefik-crds:
  traefik: true
```
- people who want to manage two or more instances of traefik will just need to install the chart as previously and either manage CRDs by themselves or install the CRD chart (they will be able to find which version of the CRD chart comes with which version of traefik or traefik-hub following the dependencies section of the main chart)
- people who want a specific version of traefik will fall in the previous case

However, even with the option 1, the main points remains how users can migrate from the previous system to the new one ? Adding the annotation works but the command I provided is mainly linux friendly :-)

genebean:
It might be worth checking out how https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack is setup. It bundles https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-operator-crds in but has an option to not install them for when you want to manage them separately. I’ve been using that lately and it works well and isn’t too difficult.

traefik/values.yaml
```diff
@@ -951,3 +951,11 @@ hub:
       insecureSkipVerify: false
   # Enable export of errors logs to the platform. Default: true.
   sendlogs:  # @schema type:[boolean, null]
+
+traefik-crds:
+  # -- Set all the following to false if you want to manage CRDs your-self
+  traefik: true
+  hub: |
+    {{- tpl "and .Values.hub.token .Values.hub.apimanagement.enabled"' . }}
+  gateway_api: |
```
mloiseleur, line 960:
```suggestion
  gatewayAPI: |
```

traefik/values.yaml
```diff
@@ -965,3 +965,14 @@ hub:
       insecureSkipVerify: false
   # Enable export of errors logs to the platform. Default: true.
   sendlogs:  # @schema type:[boolean, null]
+
+traefik-crds:
+  # -- Set it to true to opt-in CRD management
```
mloiseleur, line 970:
```suggestion
  # -- When set to true, it will enable the optional Chart dedicated to CRD management
```

mloiseleur:
Following this last review, the behavior observed when using the dedicated crd chart as a dependency is clearly not satisfactory: 

-  it's not possible to use CR like the Dashboard or a Gateway: the CRDs are not yet deployed.

The only viable option left is to provide an additional optional chart for CRD, without using dependency mechanism. 
User will have to take care of correct versioning between this Traefik Chart and this CRD chart.
The documentation should be updated to explain how to deploy this Chart, with a clean cluster or with a Traefik Chart already installed.

traefik-crds/Chart.yaml
mloiseleur: Yes, that's right. I updated my suggestion accordingly

traefik-crds/values.yaml
```diff
@@ -0,0 +1,4 @@
+traefik: true
+gatewayAPI: false  # @schema type:[string, boolean]; default: false
+hub: false  # @schema type:[string, boolean]; default: false
```
mloiseleur, line 3:
```suggestion
gatewayAPI: false 
hub: false  
```

traefik-crds/values.yaml
```diff
@@ -0,0 +1,4 @@
+traefik: true
```
mloiseleur, line 1: Wdyt of providing comments explaining what it does ?

README.md
```diff
@@ -96,8 +110,41 @@ kubectl apply --server-side --force-conflicts -k https://github.com/traefik/trae
 helm upgrade traefik traefik/traefik
 ```
 
+> [!WARNING]
+> If you want to upgrade from a standard installation to a CRDs chart powered one,
```
mloiseleur, line 114:
```suggestion
> When upgrading from standard installation to the one with additional CRD chart,
```

README.md
```diff
@@ -96,8 +110,41 @@ kubectl apply --server-side --force-conflicts -k https://github.com/traefik/trae
...
+```bash
+kubectl get customresourcedefinitions.apiextensions.k8s.io -o name | grep traefik.io | xargs kubectl patch --type='json' -p='[{"op": "add", "path": "/metadata/labels", "value": {"app.kubernetes.io/managed-by":"Helm"}},{"op": "add", "path": "/metadata/annotations/meta.helm.sh~1release-name", "value":"traefik-crds"},{"op": "add", "path": "/metadata/annotations/meta.helm.sh~1release-namespace", "value":"traefik-crds"}]'
+# If you use gateway API, you might also want to change Gateway API ownership
+kubectl get customresourcedefinitions.apiextensions.k8s.io -o name | grep gateway.networking.k8s.io | xargs kubectl patch --type='json' -p='[{"op": "add", "path": "/metadata/labels", "value": {"app.kubernetes.io/managed-by":"Helm"}},{"op": "add", "path": "/metadata/annotations/meta.helm.sh~1release-name", "value":"traefik-crds"},{"op": "add", "path": "/metadata/annotations/meta.helm.sh~1release-namespace", "value":"traefik"}]'
+helm install traefik-crds traefik/traefik-crds
+```
+
 New major version indicates that there is an incompatible breaking change.
 
+### A CRDs chart powered installation
```
mloiseleur, line 126:
```suggestion
### With additional CRDs chart
```

README.md
```diff
@@ -96,8 +110,41 @@ kubectl apply --server-side --force-conflicts -k https://github.com/traefik/trae
...
+
+To upgrade from an already installed chart:
+```bash
+# Update repository
+helm repo update
+# Change CRDs ownership
+# Upgrade Traefik
+helm install traefik-crds traefik/traefik-crds -n traefik
+```
+
```
mloiseleur, line 147:
````suggestion
````

File: README.md [hunk 22]
```diff
@@ -56,7 +58,9 @@ Due to changes in API Group of Traefik CRDs from `containo.us` to `traefik.io`,
 | Chart v23.0.0 and above |  [x]                        | [x]                    |
 | Chart v28.0.0 and above |                             | [x]                    |
 
-### Deploying Traefik
+### Deploying
```
mloiseleur, line 61:
```suggestion
## Deploying
```

README.md
```diff
@@ -56,7 +58,9 @@ Due to changes in API Group of Traefik CRDs from `containo.us` to `traefik.io`,
...
+
+#### The standard way
```
mloiseleur, line 63:
```suggestion
### The standard way
```

README.md
```diff
@@ -75,15 +79,25 @@ Complete documentation on all available parameters is in the [default file](./tr
 helm install -f myvalues.yaml traefik traefik/traefik
 ```
 
-🛂 **Warning**: Helm v2 support was removed in the chart version 10.0.0.
+#### With additional CRDs chart
```
mloiseleur, line 82:
```suggestion
### With additional CRDs chart
```

README.md
```diff
@@ -75,15 +79,25 @@ Complete documentation on all available parameters is in the [default file](./tr
...
+
+```bash
+helm install traefik-crds traefik/traefik-crds
+helm install traefik traefik/traefik --skip-crds
+helm list # should display two charts installed
+```
 
 ## Upgrading
 
 One can check what has changed in the [Changelog](./traefik/Changelog.md).
```
mloiseleur, line 93:
```suggestion

New major version indicates that there is an incompatible breaking change.
```

README.md
```diff
@@ -96,8 +110,32 @@ kubectl apply --server-side --force-conflicts -k https://github.com/traefik/trae
...
+> you **have** to change ownership on CRDs **before** installing CRDs chart
+
+```bash
+kubectl get customresourcedefinitions.apiextensions.k8s.io -o name | grep traefik.io | xargs kubectl patch --type='json' -p='[{"op": "add", "path": "/metadata/labels", "value": {"app.kubernetes.io/managed-by":"Helm"}},{"op": "add", "path": "/metadata/annotations/meta.helm.sh~1release-name", "value":"traefik-crds"},{"op": "add", "path": "/metadata/annotations/meta.helm.sh~1release-namespace", "value":"traefik-crds"}]'
+# If you use gateway API, you might also want to change Gateway API ownership
+kubectl get customresourcedefinitions.apiextensions.k8s.io -o name | grep gateway.networking.k8s.io | xargs kubectl patch --type='json' -p='[{"op": "add", "path": "/metadata/labels", "value": {"app.kubernetes.io/managed-by":"Helm"}},{"op": "add", "path": "/metadata/annotations/meta.helm.sh~1release-name", "value":"traefik-crds"},{"op": "add", "path": "/metadata/annotations/meta.helm.sh~1release-namespace", "value":"traefik"}]'
+helm install traefik-crds traefik/traefik-crds
+```
+
 New major version indicates that there is an incompatible breaking change.
```
mloiseleur, line 124:
```suggestion
```

mloiseleur (approved):
LGTM

PR #1224, opened 2024-10-14
chore: allow TRACE log level

fty4:
### What does this PR do?

Allow to specify log level `TRACE` which is a valid Traefik log level.
see: https://doc.traefik.io/traefik/observability/logs/#level

### Motivation

I tried to debug something and needed `TRACE` as log.level which was forbidden by this chart:

```log
Error: Failed to render chart: exit status 1: Error: values don't meet the specifications of the schema(s) in the following chart(s):
traefik:
- logs.general.level: logs.general.level must be one of the following: "INFO", "WARN", "ERROR", "FATAL", "PANIC", "DEBUG"
```

### More

- [x] ~Yes, I updated the tests accordingly~

### DIFF
--- a/traefik/values.schema.json
+++ b/traefik/values.schema.json
@@ -597,7 +597,8 @@
                                 "ERROR",
                                 "FATAL",
                                 "PANIC",
-                                "DEBUG"
+                                "DEBUG",
+                                "TRACE"
                             ],
                             "type": "string"
                         },


traefik/values.yaml
```diff
@@ -337,8 +337,8 @@ logs:
     # -- Set [logs format](https://doc.traefik.io/traefik/observability/logs/#format)
     format:  # @schema enum:["common", "json", null]; type:[string, null]; default: "common"
     # By default, the level is set to INFO.
-    # -- Alternative logging levels are DEBUG, PANIC, FATAL, ERROR, WARN, and INFO.
-    level: "INFO"  # @schema enum:[INFO,WARN,ERROR,FATAL,PANIC,DEBUG]; default: "INFO"
+    # -- Alternative logging levels are DEBUG, PANIC, FATAL, ERROR, WARN, TRACE and INFO.
```
mloiseleur, line 340:
```suggestion
    # -- Alternative logging levels are TRACE, DEBUG, INFO, WARN, ERROR, FATAL, and PANIC.
```

traefik/values.yaml
```diff
@@ -337,8 +337,8 @@ logs:
...
+    level: "INFO"  # @schema enum:[INFO,WARN,ERROR,FATAL,PANIC,DEBUG,TRACE]; default: "INFO"
```
mloiseleur, line 341:
```suggestion
    level: "INFO"  # @schema enum:[TRACE,DEBUG,INFO,WARN,ERROR,FATAL,PANIC]; default: "INFO"
```

mloiseleur:
@fty4 Thanks for this PR :+1: .
You're right, we missed it.
I left small suggestion in order to display (and configure) it with severity order, like on Traefik Proxy documentation.
Wdyt ?

fty4:
Hi @mloiseleur 
just saw after I deployed my modified version that no log on level TRACE was printed.

After reviewing the traefik/traefik sourcecode I saw that not one `logger.Trace()` was called - I guess the level is [never used](https://github.com/search?q=repo%3Atraefik%2Ftraefik%20logger.Trace()&type=code).
zerolog allows to specify that level - because of that it might be "allowed".

Knowing this we might just close this PR, right?
Might better to remove the level from the main repository documentation?

darkweaver87 (approved):
LGTM

mloiseleur (approved):
After second thought, I think I agree with @darkweaver87 .
This is a level supported by zerolog and it's documented in Traefik Proxy.
So for me it's good to go.
Wdyt ? Is this ok for you ? can we merge it ?

fty4:
I am fine with that - you can merge it!
Also we get the correct order of the severity as you suggested.

PR #1227, opened 2024-10-18
fix(Traefik Hub): RBAC for distributedAcme

mloiseleur:
### What does this PR do?

Allows to update and delete secret with Traefik Hub API Gateway

### Motivation

It's needed when using [distributedAcme](https://doc.traefik.io/traefik-hub/api-gateway/reference/install/tls/ref-certs-resolvers).

### More


darkweaver87 (approved):
LGTM

PR #1237, opened 2024-10-29
chore(renovate): add regex manager for VALUES.md

logica0419:
### What does this PR do?

Added renovate regex manager to update VALUES.md

### Motivation

https://github.com/traefik/traefik-helm-chart/pull/1235#issuecomment-2443548382

### More


mloiseleur:
@logica0419 Thanks. 
Would you please also add grouping so this regex is done in the same PR where the docker image is updated ?

logica0419:
I guess the updates from the same datasource are automatically grouped.
ref: https://github.com/logica0419/traefik-helm-chart/pull/2

Do I get what you mean correctly?

mloiseleur (approved):

PR #1250, opened 2024-11-12
fix: :bug: support specifying plugins storage

darkweaver87:
### What does this PR do?

This PR adds the support of custom plugins storage.

### Motivation

Adding `plugins` volume to the current chart make duplicates in volumes.

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -176,3 +176,13 @@ Key: {{ $cert.Key | b64enc }}
       {{ printf "- \"%s\"\n" . }}
     {{- end -}}
 {{- end -}}
+
+{{- define "traefik.hasPluginsVolume" -}}
+    {{- $found := false -}}
+    {{- range . -}}
+       {{- if eq .name "plugins" -}}
+           {{ $found = true }}
+       {{- end -}}
+    {{- end -}}
+    {{- $found -}}
+{{- end -}}
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -156,7 +156,7 @@
             mountPath: {{ .mountPath }}
             readOnly: true
           {{- end }}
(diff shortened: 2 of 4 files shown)


traefik/templates/_podtemplate.tpl
```diff
@@ -156,7 +156,7 @@
             mountPath: {{ .mountPath }}
             readOnly: true
           {{- end }}
-          {{- if gt (len .Values.experimental.plugins) 0 }}
+          {{- if and (gt (len .Values.experimental.plugins) 0) (ne (include "traefik.hasPluginsVolume" .Values.deployment.additionalVolumes) "true") }}
```
mloiseleur, line 159:
```suggestion
          {{- if and (gt (len .Values.experimental.plugins) 0) (not (hasKey .Values.deployment.additionalVolumes "plugins")) }}
```

traefik/templates/_helpers.tpl
```diff
@@ -176,3 +176,13 @@ Key: {{ $cert.Key | b64enc }}
...
+
+{{- define "traefik.hasPluginsVolume" -}}
+    {{- $found := false -}}
+    {{- range . -}}
+       {{- if eq .name "plugins" -}}
+           {{ $found = true }}
+       {{- end -}}
+    {{- end -}}
+    {{- $found -}}
+{{- end -}}
```
mloiseleur, line 188:
```suggestion
```

Can this be replaced with `hasKey` :thinking: or did I miss something ?

mloiseleur (approved):
Test and code is working.
See my suggestion to reduce code size and custom code.

traefik/templates/_helpers.tpl
darkweaver87: It's not a map but an object list.

darkweaver87:
```
executing "traefik.podTemplate" at <.Values.deployment.additionalVolumes>: wrong type for value; expected map[string]interface {}; got []interface {}
```

Issue #1254, opened 2024-11-15
Allow changing automountServiceAccountToken in the PodSpec


vidarno:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

There is no way to set automountServiceAccountToken in the current helm chart.

automountServiceAccountToken is set to false in the actual service account definition, but in environments that have policies that require automountServiceAccountToken to be false in the actual Podspec this isn't good enough.

Ideally the same value would change automountServiceAccountToken in both the service account definition and in the podspec, but since they are respectively true and false by default this would effectively change the default behaviour if both were set by this one value so I suggest to only have this value control the attribute in the podspec and not both.

I am ready to provide a PR that provides this feature. It passes the tests and has updated the schema.

vidarno:
To show what I mean, here is a PR for the change - https://github.com/traefik/traefik-helm-chart/pull/1262

mloiseleur:
There is no way because Traefik needs this token. 
Would you please describe with more details your use case ?

mloiseleur:
We are interested in this issue, but we’re unsure about the use case and the traction it will receive. We are going to leave the status as kind/proposal to give the community time to let us know if they would like this idea.
We will reevaluate as people respond.

mloiseleur:
We haven't seen any traction from the community on this since last year.

PR #1262, opened 2024-11-25
Allow changing automountServiceAccountToken in PodSpec via values.yaml

vidarno:
### What does this PR do?

This PR adds serviceAccount.automountServiceAccountToken to values.yaml and thus allows changing automountServiceAccountToken in the PodSpec, which is true by default. This PR does not change the default behavior.

### Motivation

automountServiceAccountToken is set to false in the actual service account definition, but in environments that have policies that require automountServiceAccountToken to be false in the actual Podspec this isn't good enough. Ideally the same value would change automountServiceAccountToken in both the service account definition and in the podspec, but since they are respectively true and false by default this would effectively change the default behaviour if both were set by this one value so I opted to only have this value control the attribute in the podspec.

This is the issue described in https://github.com/traefik/traefik-helm-chart/issues/1254
### More

- [X ] Yes, I updated the tests accordingly

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -23,7 +23,7 @@
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "traefik.serviceAccountName" . }}
-      automountServiceAccountToken: true
+      automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
       terminationGracePeriodSeconds: {{ default 60 .Values.deployment.terminationGracePeriodSeconds }}
       hostNetwork: {{ .Values.hostNetwork }}
       {{- with .Values.deployment.dnsPolicy }}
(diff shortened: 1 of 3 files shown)


vidarno:
The need comes from environments that don't allow automounting the serviceaccount due to policies (OPA, Kyverno or similar).

The solution in such an environment is to disable automounting of the serviceaccount and instead use service account token projection, like described here - https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#launch-a-pod-using-service-account-token-projection

This helm chart already has support for volumes and volumemounts, so this should be a good alternative to automounting the serviceaccount for those who are unable to do so.

darkweaver87:
Thanks @vidarno for the detailed explanation.
Then, could you just add a test and I'll be happy to review and merge it ?

Passie1982 (approved):

mloiseleur:
Since there is no answer from vidarno, I close this PR.
Feel free to re-open this PR or open a new one if you need it.

PR #1268, opened 2024-12-05
feat(Traefik Proxy): add `tracing.sampleRate`

slayne:
### What does this PR do?

This PR adds the option to specify a tracing samplingRate in the chart values

### Motivation

The tracing samplingRate was not exposed in the values, meaning that we have to specify it directly in CLI options if needed.
solves #1269 

### More


mloiseleur (commented):
Thanks for this PR. 
It has been introduced in v3.0 so there's no need to add a version checker.
About implementation, there are two options:

1. Lazy option 

```sh
{{- with .Values.tracing.sampleRate }}
[...]
```

2. Refactor option

```helm
{{- with .Values.tracing }}
[all options starting .Values.tracing]
{{- with .sampleRate }}
```

PR #1286, opened 2024-12-17
fix: support uppercase letters in entrypoint names

uvNikita:
### What does this PR do?

Traefik supports entrypoints with uppercase letters (e.g. webSecure), but helm generated manifest fails to apply because port names can only be lowercase. 

This forces lower case on port names for entrypoints.

### Motivation

We are migrating traefik deployment from raw yaml files to helm chart, but one of our entrypoints has uppercase letter in the name (`webSecure`).

This issue prevents the migration since helm generates invalid pod template with port name based on entrypoint name with uppercase letter.

From [Kubertes Docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#endpointport-v1-discovery-k8s-io):

> EndpointPort
> name  .... must consist of lower case alphanumeric characters ...

### More


jnoordsij:
Hey there! Thanks for your contribution.

I have two questions/suggestions:
1. Although Traefik itself allows uppercase entrypoints, this chart solely defines them in lowercase. For example, the values for `websecure` are found here: https://github.com/traefik/traefik-helm-chart/blob/acd4405e4304637b4a273c03ad4a03fa3f03be3a/traefik/values.yaml#L669-L729
Similarly almost all of the docs examples use lowercase entrypoints. (I'm not exactly sure, but some naming might even be converted to lowercase internally.) Considering all that, it might be easier to migrate your values to use the lowercase name. However I understand there may be particular conditions to your use-case that prevent you from doing so.
2. You propose to change the `name` field on the port on the pod (see [docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#containerport-v1-core) for the exact restrictions on this field). However, the same name is also used on the service, which also has (less strict) restrictions (see [docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#serviceport-v1-core). Therefore, if we were to adopt these changes, I think we should be a little more careful here and introduce some additional helper that can be reused across various places where we reference this.

uvNikita:
> Considering all that, it might be easier to migrate your values to use the lowercase name.

Our entrypoint is currently defined as `webSecure` (note capital S) and we don't have direct control of the Ingress / IngressRoute since they are managed by our users. So it will be a big undertaking for us to coordinate a change with multiple teams to modify an entrypoint for all objects. 

We can of course automate the change on the running cluster and add a mutation hook to fix incoming objects, but ideally we could perform helm migration without this, since it's a valid entrypoint name at least from the traefik's prospective.

As per the second point, adjusting the name to the lower case at least brings the `name` field closer to the requirement from the Kubernetes API side.

Maybe some sort of name sanitizer would be even better, but IMHO lowering the case improves the current state iteratively and simple enough change that can be merged now. Further improvement could build on top of it.

mloiseleur:
@uvNikita thanks for all those details about your use case, which is quite unexpected.

As @jnoordsij said:
> However, the same name is also used on the service,

Would you please add a test on `Service` with an uppercase entrypoint name ?

uvNikita:
@mloiseleur You mean something like this?

I added `| lower` call to a service port name too and a test for it.

mloiseleur:
@uvNikita Yes, thanks. There is the same kind of test required for service, also, isn't it ?

mloiseleur (approved):

darkweaver87 (approved):
LGTM

PR #1300, opened 2025-01-09
fix(Gateway API): status should not use default service when it's disabled

### LINKED ISSUE #1291: kubernetes gateway service probably should not default to the helm chart service if the service is disabled
### Welcome!

33.2.1

3.1.2

### What did you do?

Excerpt from the values file:
```
...
ports:
  web: null
  websecure:
    hostPort: 9443
    port: 9443
    tls:
      enabled: true
service:
  enabled: false
providers:
  kubernetesGateway:
    enabled: true
gateway:
  listeners:
    web: null
    websecure:
      port: 9443
      protocol: HTTPS
      certificateRefs:
        - kind: Secret
          name: gateway-cert
...
```

### What did you see instead?

*Comment with regards to the bug report form: the question "What did you see instead?" only makes sense if you asked "What did you expect to see?" first*

In the logs:
```
ERR Unable to get Gateway status addresses error="could not find a service with name traefik in namespace traefik" providerName=kubernetesgateway
```

It appears to me that those are not relevant, but if you feel some crucial information is missing, please let me know and I'll gladly provide it.

### Additional Information

```shell
If possible, the chart should check if the service is enabled, and only use the default, if this is the case.
```

mloiseleur:
### What does this PR do?

Set empty default on status of kubernetesGateway when service is disabled

### Motivation

Fixes #1291

### More


darkweaver87 (approved):
LGTM

PR #1301, opened 2025-01-10
fix(Traefik Proxy)!: move redirectTo => redirections

### LINKED ISSUE #1288: Improved Documentation for values.yaml
### Welcome!

The documentation around the use of values.yaml should be improved. I was unable to find documentation on what the ports.web.redirectTo value should be, beyond a generic object. It took a ton of searching before finally finding a similar setup that used port: websecure as the value for this key. There may be other configurable parameters in the ports.web.redirectTo that I'm unaware of because there's no documentation.

mloiseleur:
### What does this PR do?

Refactor `redirectTo` section in `ports` with upstream Traefik Proxy syntax.

**Before**

```yaml
ports:
  web:
    redirectTo:
      port: websecure
      permanent: true
```

**After**

```yaml
ports:
  web:
    redirections:
      entryPoint:
        to: websecure
        scheme: https
        permanent: true
```

### Motivation

Fixes #1288

### More


traefik/templates/_podtemplate.tpl
```diff
@@ -598,14 +598,25 @@
           {{- range $entrypoint, $config := $.Values.ports }}
           {{- if $config }}
             {{- if $config.redirectTo }}
-             {{- $toPort := index $.Values.ports $config.redirectTo.port }}
+              {{- fail "ERROR: redirectTo syntax has been removed in v34 of this Chart. See Release notes or EXAMPLES.md for new syntax." -}}
+            {{- end }}
+            {{- if $config.redirections }}
+             {{- with $config.redirections.entryPoint }}
+              {{- if and (eq .to "websecure") (ne .scheme "https") }}
```
darkweaver87, line 605: We can't assume users  didn't specify:

```yaml
ports:
  websecure:
    tls:
      enabled: false
```

darkweaver87: Could you just check if the port on which we redirect has `tls.enabled:  true` ?
We might also want to check if the `to` port exists.

darkweaver87 (approved):
LGTM

PR #1307, opened 2025-01-13
chore(Gateway API): :recycle: remove template from values

darkweaver87:
### More


traefik/values.yaml
```diff
@@ -319,8 +319,9 @@ providers:  # @schema additionalProperties: false
       hostname: ""
       # -- The Kubernetes service to copy status addresses from. When using third parties tools like External-DNS, this option can be used to copy the service loadbalancer.status (containing the service's endpoints IPs) to the gateways. Default to Service of this Chart.
       service:
```
mloiseleur, line 321:
```suggestion
      service:
          # -- The Kubernetes service to copy status addresses from. When using third parties tools like External-DNS, this option can be used to copy the service loadbalancer.status (containing the service's endpoints IPs) to the gateways. Default to Service of this Chart.
```

mloiseleur (approved):

PR #1324, opened 2025-01-28
feat: :sparkles: add missing consulCatalogEnterprise provider for Hub

darkweaver87:
### More


mloiseleur (commented):
Wdyf of having a test with minimal required configuration and a test with all parameters set (and so checked) ?

mloiseleur (approved):

PR #1327, opened 2025-01-29
fix: :bug: fix and test changelog

darkweaver87:
### More


mloiseleur (approved):
LTTM

PR #1333, opened 2025-02-04
aiservices are available in Hub API GW

juliens:
### What does this PR do?

This PR moves the rbac for aiservices to Hub API GW instead of only Hub APIM

### Motivation

Be able to use AIServices in Hub API GW

### More


mloiseleur (approved):

traefik/templates/rbac/clusterrole.yaml
```diff
@@ -202,6 +202,14 @@ rules:
       - delete
   {{- end }}
   {{- if .Values.hub.token }}
```
mloiseleur, line 204:
```suggestion
```

mloiseleur (changes requested):
See my suggestion.

mloiseleur (approved):

Issue #1311, opened 2025-01-15
Json schema validation fails when using traefik-crds as chart dependency


Soortor:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

1.1.0

### What version of Traefik are you using?

34.0.0

### What did you do?

We deploy many of our applications with ArgoCD by defining them as subcharts.
For Traefik this works, but for Traefik-crds this does not as the `additionalProperties` is set to false in the `.schema.yaml`

Example chart.yaml to reproduce:

```
apiVersion: v2
name: traefik-crds
description: Traefik CRDS Helm Chart

version: 1.0.0

dependencies:
  - name: traefik-crds
    version: 1.1.0
    repository: https://traefik.github.io/charts
```

The expectation is that this renders a helm template with default values.

### What did you see instead?

```
$ helm template .
Error: values don't meet the specifications of the schema(s) in the following chart(s):
gateway:
- (root): Additional property global is not allowed
```

### What is your environment & configuration?

ArgoCD

### Additional Information

```shell
Similar issue reported here: https://github.com/helm/helm/issues/10392#issuecomment-1091673838
It appears that helm implicitly injects a global section from the parent chart, which is not allowed by the subchart.
```

darkweaver87:
Hello @Soortor ,

Thanks for reporting this.
Those charts needs to be used separately (cf. https://github.com/traefik/traefik-helm-chart/blob/master/README.md#an-installation-with-additional-crds-chart).
After a lot of tests we intentionally removed the dependency from the original [PR](https://github.com/traefik/traefik-helm-chart/pull/1223).
We will open a PR to detail the reasons so it will be explicit in the README.

Rémi

Soortor:
Instead of loading as a subchart, we ended up loading the chart directly via ArgoCD:

```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: traefik-crds
  namespace: argocd
  finalizers:
    - resources-finalizer.argocd.argoproj.io
spec:
  project: default
  source:
    chart: traefik-crds
    repoURL: https://traefik.github.io/charts
    targetRevision: 1.1.0
  destination:
    server: https://kubernetes.default.svc
  syncPolicy:
[...]

PR #1334, opened 2025-02-07
Update .schema.yaml to allow additionalProperties

portswigger-tim:
### What does this PR do?

additionalProperties is set to false this presents issues to users that might inject a set of properties to be passed to multiple charts. 

### More


mloiseleur:
@portswigger-tim Am I correct to understand you are trying to fix #1311 ?

If so, wdyt about adding a `global` property, and so both solve the issue and keep the benefits of using a schema ?

mloiseleur:
superseded by #1348

Issue #1315, opened 2025-01-20
Skip applying CRDs in main traefik helm chart.


PurseChicken:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Now that traefik-crds exist as a separate chart, we need another way to "skip" applying the CRDs from the main traefik helm chart.

Currently the only option is to use the helm option of --skip-crds. This is not ideal as there may be other ways that the chart is applied beyond using helm directly. ArgoCD specifically allows you to use .spec.source.helm.skipCrds: true but using this in an application set means that it applies to every application manifest derived from the application set. As you can imagine this is not ideal.

There are a few options that I think can be considered:

**1.** Completely remove the CRDs from the traefik chart and require applying CRDs with the separate chart (bonus here is that you only have to manage CRDs in one place):

```
helm install traefik-crds traefik/traefik-crds
helm install traefik traefik/traefik
```

**2.** Convert existing traefik/crds directory to a sub-chart. Reference it as a dependency chart enabled by a key value pair.

Example:

values.yaml:
```values.yaml
crds:
  enabled: true
```

Chart.yaml:
```Chart.yaml
dependencies:
    - name: crds
      version: "1.2.0"
      condition: crds.enabled
```

traefik/crds/Chart.yaml:
```Chart.yaml
apiVersion: v2
name: crds
version: 1.2.0
```

~~**3.** Change CRD manifests to include if block based on key value pair in values:~~
**STRIKED, as this is not a valid option. You can't template manifests in the default CRD folder.**

exampleCrdManifest.yaml:
```exampleCrdManifest.yaml
{{- if .Values.crds.enabled }}
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
........
......
....
..
{{- end }}
```

values.yaml:
```values.yaml
crds:
  enabled: true
```

grantcarthew:
More context from a Google Cloud Platform perspective.

Googles documentation states to enable the Gateway API as either an option on the deployment of a new cluster or executing a command against an existing cluster:

https://cloud.google.com/kubernetes-engine/docs/how-to/deploying-gateways

This gives you the Gateway API CRDs to support the GKE Gateway Controller:

[...]

PR #1340, opened 2025-02-12
Change CRD's to use crds-files pattern like traefik-crds chart

PurseChicken:
### What does this PR do?

This PR changes the way CRD's are applied. Prior to this, CRD's were using helms default "crds" directory which did not allow any templating of the CRD's. Thus the only way to disable the CRD's in this chart was to use the --skip-crds helm flag.

In this PR, the CRDs are now applied in a simliar way to the existing traefik-crds chart. There is now a .Values.crds.enabled key which allows you to enable or disable the traefik charts crds. This is referenced by the crds.yaml file which then uses a helper to apply the CRD's if .Values.crds.enabled is true. This is the same patten as the traefik-crds chart.

### Motivation

Please see: https://github.com/traefik/traefik-helm-chart/issues/1315

We need a way to enable or disable the traefik CRD's beyond using the helm flag of --skip-crds. With this PR, you can now use .Values.crds.enabled and control the CRDs that way. This is extremely useful when you may, for example, apply CRD's separately via the traefik-crds helm chart.

### More


PurseChicken:
If this gets approved \ merged, the documentation for applying the traefik chart along with the traefik-crds chart likely need to be changed. Since --skip-crds wont be required. Instead just making a value change at .Values.crds.enabled

PurseChicken:
> If this gets approved \ merged, the documentation for applying the traefik chart along with the traefik-crds chart likely need to be changed. Since --skip-crds wont be required. Instead just making a value change at .Values.crds.enabled

Added a commit with this addressed:
https://github.com/traefik/traefik-helm-chart/pull/1340/commits/b88f93939cabc7e8879c9be1b5191015daa0435b

mloiseleur:
@PurseChicken I tested this PR. It does not work:

```sh
gh pr checkout 1340
# Install the Helm chart
helm install traefik --wait \
  --set ingressClass.enabled=false \
  --set ingressRoute.dashboard.enabled=true \
  --set ingressRoute.dashboard.matchRule='Host(`dashboard.docker.localhost`)' \
  --set ingressRoute.dashboard.entryPoints={web} \
  --set ports.web.nodePort=30000 \
  --set ports.websecure.nodePort=30001 \
   ./traefik
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "traefik-dashboard" namespace: "traefik" from "": no matches for kind "IngressRoute" in version "traefik.io/v1alpha1"
```

PurseChicken:
> @PurseChicken I tested this PR. I got an error:
> 
> ```shell
> gh pr checkout 1340
> # Install the Helm chart
> helm install traefik --wait \
>   --set ingressClass.enabled=false \
>   --set ingressRoute.dashboard.enabled=true \
>   --set ingressRoute.dashboard.matchRule='Host(`dashboard.docker.localhost`)' \
>   --set ingressRoute.dashboard.entryPoints={web} \
>   --set ports.web.nodePort=30000 \
>   --set ports.websecure.nodePort=30001 \
>    ./traefik
> Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "traefik-dashboard" namespace: "traefik" from "": no matches for kind "IngressRoute" in version "traefik.io/v1alpha1"
> ```

Thanks! I think this is an issue with the CRD's not fully applying before the resources try to reference it. We might have to hook the CRD's to apply first. I will do some additional testing and make sure this is resolved before we move forward.

crouth-redge:
This needs to be reopened and prioritized, as mentioned on https://github.com/traefik/traefik-helm-chart/issues/1315 - There is currently no viable method to install this on GKE Autopilot now that the ability to turn Googles Gateway CRDs off is gone and that the Traefik CRDs chart is documented as _deprecated_.

PR #1348, opened 2025-02-26
Fix: Allow using the traefik-crds chart as a subchart/dependency

### LINKED ISSUE #1311: Json schema validation fails when using traefik-crds as chart dependency

ArlonAntonius:
### What does this PR do?
Fixes #1311
Superseeds #1334
TL:DR Allows the usage of the traefik-crds chart as a subchart/dependency.

### Motivation
To use the traefik-crds Helm Chart as a dependency in your projects, the validation needs to be slightly looser to support the usage of the chart. Helm passes specific values by default (`global` in this case), causing the Schema Validation to give an error.

In line with other Helm Charts (like cert-manager), it would be beneficial to allow users to set:
- `global` to fix the error that will always be caused by Helm passing the `global` values
- `enabled` to provide users with the default (recommended) flag for conditions on Helm Charts

The one thing that I did, that is more opinionated and would like the opinion of Traefik's maintainer on is the usage of `additionalProperties` on the `global` object. Cert-manager specifically provides a set of variables, defeating part of the feature's purpose (in my opinion).
https://github.com/cert-manager/cert-manager/blob/23d3aea079ce88cec6f9394f4a4027720f8d7c2a/deploy/charts/cert-manager/values.schema.json#L686

### More

### DIFF
--- a/traefik-crds/values.schema.json
+++ b/traefik-crds/values.schema.json
@@ -4,6 +4,16 @@
     "additionalProperties": false,
     "description": "The Cloud Native Application Proxy",
     "properties": {
+        "enabled": {
+            "default": true,
+            "description": "Field that can be used as a condition when this chart is a dependency. This definition is only here as a placeholder such that it is included in the JSON schema. See https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags for more info.",
+            "type": "boolean"
+        },
+        "global": {
+            "description": "Global values shared across all (sub)charts",
+            "type": "object",
+            "additionalProperties": true
+        },
         "deleteOnUninstall": {
             "type": "boolean"
         },


mloiseleur:
@ArlonAntonius I guess you mean on #1311 . We can add required properties from some use cases and keep the benefits of the schema.

FYI, the schema is generated from values:

https://github.com/traefik/traefik-helm-chart/blob/6350cb1f068f9f79e3540567742ebf7a0064fef8/Makefile#L30-L35

Would you please update your PR accordingly, so the schema will be kept ?

ArlonAntonius:
@mloiseleur Made the required changes, let me know if this aligns correctly with Traefik's coding and documentation standards 😉 
If it's easier to discuss something in real-time, feel free to let me know.

traefik-crds/values.yaml
```diff
@@ -2,6 +2,14 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into templates
 
+# -- Global values
+# This definition is only here as a placeholder such that it is included in the json schema.
+global: {} # @schema additionalProperties: true
+# -- Set it to true when this chart is used as a dependency to install the chart.
+# Field that can be used as a condition when this chart is a dependency.
```
mloiseleur, line 9:
```suggestion
# -- Field that can be used as a condition when this chart is a dependency.
```

mloiseleur:
@ArlonAntonius See my suggestion for improved documentation. Otherwise, LGTM.

ArlonAntonius:
@mloiseleur Included your change 🚀 Any chance we can include this in the release you have planned in #1351 ?

mloiseleur (approved):

mloiseleur:
@ArlonAntonius You need to rebase this PR (and/or _allow edit from maintainers_)

ArlonAntonius:
@mloiseleur I will have updated the branch and gave your permission to the repository, otherwise I have to recreate the entire PR right now as far as I know.

Does that help for now?

mloiseleur:
@ArlonAntonius It's good. Thanks.

PR #1362, opened 2025-03-13
Do not quote protocol on pod ports

jnoordsij:
### What does this PR do?

No longer quote the protocol value on pod ports.

### Motivation

I found that quoting of ports only happens on pods, not on services, and moreover the http3 port has an unquoted UDP as default value. When trying to mimic the http3 port configuration with some manual values to allow for smoother upgrade of my Traefik deployment, I figured I couldn't achieve the exact same result as by enabling the http3 flag due to this fact.

Note this is not breaking or blocking in any way, but mostly a slightly cleaner and more consistent output.

### More

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -116,7 +116,7 @@
           {{- if $config.hostIP }}
           hostIP: {{ $config.hostIP }}
           {{- end }}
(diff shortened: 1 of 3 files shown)


darkweaver87 (approved):
LGTM

mloiseleur (approved):

PR #1363, opened 2025-03-13
Add port name template functions

jnoordsij:
### What does this PR do?

Create a dedicated template function for parsing port names and references to port names.

It also removes various quotes from port names that no longer appear in template output,
although this should not change behavior as these quotes should be unused already.

Thus in practice, this should change nothing at all in chart functionality.

### Motivation

As a follow-up to #1286, this creates a dedicated helper, that allows for
- reusing more easily
- adding documentation about what we're doing and why
- easier extensibility or validation in the future 
- preventing future errors (e.g. `websecure-http3` exactly matches the allowed length of 15 characters, so we got lucky 😉)

I've not bothered with adding e.g. regex checks to the portnames, given that probably it won't serve any real-world usecase and probably Kubernetes itself will also complain already about invalid ones.

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -93,6 +93,32 @@ are multiple namespaced releases with the same release name.
 {{- end -}}
 
 {{/*
+Change input to a valid name for a port.
+This is a best effort to convert input to a valid port name,
+which only allows lowercase alphanumeric characters and '-',
+and imposes limits on the lenght of the name.
+See also https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#containerport-v1-core.
+*/}}
+{{- define "traefik.portname" -}}
+{{- $portName := . -}}
+{{- $portName = $portName | lower -}}
+{{- $portName = $portName | trunc 15 | trimSuffix "-" -}}
+{{- print $portName -}}
+{{- end -}}
+
+{{/*
+Change input to a valid port reference.
+See also the traefik.portname helper.
+*/}}
+{{- define "traefik.portreference" -}}
+{{- if kindIs "string" . -}}
+    {{- print (include "traefik.portname" .) -}}
+{{- else -}}
+    {{- print . -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Construct the path for the providers.kubernetesingress.ingressendpoint.publishedservice.
 By convention this will simply use the <namespace>/<service-name> to match the name of the
 service generated.
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -736,7 +736,7 @@ spec:
     app.kubernetes.io/instance: traefik-traefik
   ports:
   - port: 8080
-    name: "traefik"
+    name: traefik
     targetPort: traefik
     protocol: TCP
 ```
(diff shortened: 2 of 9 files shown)


traefik/templates/_helpers.tpl
```diff
@@ -92,6 +92,32 @@ are multiple namespaced releases with the same release name.
 {{- (printf "%s-%s" (include "traefik.fullname" .) (include "traefik.namespace" .)) | trunc 63 | trimSuffix "-" }}
 {{- end -}}
 
+{{/*
+Change input to a valid name for a port.
+This is a best effort to convert input to a valid port name,
+which only allows lowercase alphanumeric characters and '-',
+and imposes limits on the lenght of the name.
```
darkweaver87, line 99:
```suggestion
and imposes limits on the length of the name.
```

darkweaver87 (approved):
LGTM

mloiseleur (commented):
@jnoordsij LGTM. It seems to work well. It will also do it silently, user will need to go through the code in order to understand why its name was truncated. 

I understand we should avoid set it as an hard requirements (like those in requirements.yaml), but we can at least inform the user.
=> Wdyt of adding a warning in NOTES.txt when the name has been truncated ?

traefik/templates/_helpers.tpl
```diff
@@ -92,6 +92,32 @@ are multiple namespaced releases with the same release name.
 {{- (printf "%s-%s" (include "traefik.fullname" .) (include "traefik.namespace" .)) | trunc 63 | trimSuffix "-" }}
 {{- end -}}
 
+{{/*
+Change input to a valid name for a port.
+This is a best effort to convert input to a valid port name,
+which only allows lowercase alphanumeric characters and '-',
+and imposes limits on the length of the name.
+See also https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#containerport-v1-core.
```
mloiseleur, line 100:
```suggestion
Kubernetes portname follow this RFC: https://www.rfc-editor.org/rfc/rfc6335#section-5.1
```

traefik/templates/_helpers.tpl
```diff
@@ -92,6 +92,32 @@ are multiple namespaced releases with the same release name.
...
+*/}}
+{{- define "traefik.portname" -}}
+{{- $portName := . -}}
+{{- $portName = $portName | lower -}}
```
mloiseleur, line 104: upper case seems to be valid, according to the RFC.

traefik/templates/_helpers.tpl
```diff
@@ -92,6 +92,32 @@ are multiple namespaced releases with the same release name.
...
+{{- $portName = $portName | trunc 15 | trimSuffix "-" -}}
```
mloiseleur, line 105: Wdyt of trimPrefix on "-" ?

traefik/templates/_helpers.tpl
jnoordsij: See comment above

traefik/templates/_helpers.tpl
mloiseleur: Make sense to me :+1:

jnoordsij:
I got a bit tangled up with newline printing in NOTES.txt, so I've opted to trim everything from the file itself and just add explicit newline print directives for clarity.

traefik/templates/NOTES.txt
```diff
@@ -1,36 +1,62 @@
-
-
+{{/* Print release information */}}
+{{- printf "\n" -}}
+{{- printf "\n" -}}
```
mloiseleur, line 3:
```suggestion
{{- printf "\n\n" -}}
```

PR #1367, opened 2025-03-21
feat: ability to disable Prometheus scraping while keeping the metrics endpoint enabled

ksubileau:
### What does this PR do?

Adds a `metrics.prometheus.scrape` value to be able to configure the value of the `prometheus.io/scrape` annotation, set it as true by default to keep existing behavior

### Motivation

Sometimes it may be useful to be able to enable the metrics endpoint without actually enabling metrics collection.
However, currently, the chart enables both the endpoint and the metrics collection by default, and the only option offered is to disable both at once by setting `metrics.prometheus` to `null`.

In my specific use case, Traefik is deployed on an infrastructure where a monitoring tool is configured to automatically collect all metrics from pods with the `prometheus.io/scrape` annotation set to `true` (which is pretty standard so far). But I need to be able to set the annotation value to `false` to disable this generic collection and use another specific collector that will route the metrics differently.

### More

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -237,6 +237,7 @@ Kubernetes: `>=1.22.0-0`
 | metrics.prometheus.serviceMonitor.namespaceSelector | object | `{}` |  |
 | metrics.prometheus.serviceMonitor.relabelings | list | `[]` |  |
 | metrics.prometheus.serviceMonitor.scrapeTimeout | string | `""` |  |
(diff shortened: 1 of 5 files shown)


darkweaver87:
Hello @ksubileau ,

Thank you for your interest. 
We’re unsure about the use case and the traction it will receive. We are going to leave the status as kind/proposal to give the community time to let us know if they would like this idea.

mloiseleur:
No interest from the community over the last year.
You are welcome to re-open or open a new one if you need it.

PR #1370, opened 2025-03-31
fix(ci): :bug: avoid conflicting pushes on the same repo

darkweaver87:
### More

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,8 @@ jobs:
     uses: "traefik/traefik-helm-chart/.github/workflows/test.yml@master"
 
   traefik:
-    needs: test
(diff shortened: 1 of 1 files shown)


mloiseleur (commented):
In order to wait less, and since the e2e tests are not required for CRDs chart, wdyt about removing `needs [ test ]` on traefik-crds ?

**EDIT**: Since tests are already run and mandatory in the release PR, I suggest to remove it when merge.

mloiseleur:
Shouldn't it block, also, when e2e tests fails ?

darkweaver87:
The test job don't fail. The E2E does. I would keep test dependency.
I don't have any strong opinion a dependency on E2E. It might be a good idea.

mloiseleur (approved):

mloiseleur (approved):

PR #1379, opened 2025-04-07
fix(ci): :bug: fail on test error

darkweaver87:
### More

### DIFF
--- a/hack/test.sh
+++ b/hack/test.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
 
+set -e
+
 /usr/bin/helm unittest --color ./traefik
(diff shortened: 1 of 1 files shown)


traefik/tests/oci-config_test.yaml
```diff
@@ -21,7 +21,7 @@ tests:
     asserts:
       - equal:
           path: spec.template.spec.containers[0].image
-          value: repo.io/proxy:v3.3.5
+          value: repo.io/traefik/traefik:v3.3.5
```
mloiseleur, line 24:
```suggestion
          value: repo.io/traefik:v3.3.5
```

PR #1380, opened 2025-04-07
feat(Traefik Hub): namespaces

jspdown:
### What does this PR do?

This PR allows to configure the namespaces Traefik Hub can watch:

```
--set "hub.namespaces={ns1,ns2}"
```

By setting this field, it enables the `hub.namespaces` flag on the Traefik Hub container and configures the mutating webhook to only notify about the changes happening in the configured namespaces.

The limitation that prevented to enable `rbac.namespaced` for Hub has been removed. 

### Motivation

Give the ability to install Traefik Hub in a multi-tenant environment or simply restrict what it's able to see.

### More

Keep as draft until the new version of Traefik Hub that brings the `hub.namespaces` option is released.

### DIFF
--- a/traefik/templates/NOTES.txt
+++ b/traefik/templates/NOTES.txt
@@ -61,7 +61,7 @@ for more info. 🚨
 {{- end -}}
 
 {{/* Warn about hub not watching namespaces configured in providers */}}
-{{- if and .Values.hub.apimanagement.enabled (and .Values.rbac.enabled .Values.rbac.namespaced) }}
+{{- if and .Values.hub.token (and .Values.rbac.enabled .Values.rbac.namespaced) }}
     {{- if .Values.hub.namespaces -}}
         {{- range (list "kubernetesCRD" "kubernetesGateway" "kubernetesIngress") }}
             {{- $provider := . -}}
@@ -69,9 +69,9 @@ for more info. 🚨
             {{- $providerEnabled := get (get $.Values.providers .) "enabled" -}}
             {{- if $providerEnabled -}}
                 {{- if $providerNamespaces -}}
-                    {{- $intersect := (include "list.intersect" (dict "a" $providerNamespaces "b" $.Values.hub.namespaces)) }}
-                    {{- if $intersect }}
-                        {{- printf "WARNING: %s provider is configured to watch namespaces %s but those ones are not watched by Hub provider.\n" $provider $intersect -}}
+                    {{- $difference := (include "list.difference" (dict "a" $providerNamespaces "b" $.Values.hub.namespaces)) | fromYamlArray }}
+                    {{- if $difference }}
+                        {{- printf "WARNING: %s provider is configured to watch namespaces %s but those ones are not watched by Hub provider.\n" $provider $difference -}}
                     {{- end -}}
                 {{- else -}}
                     {{- printf "WARNING: %s provider is configured to watch all namespaces but Hub provider only watches %s.\n" $provider $.Values.hub.namespaces -}}
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -248,10 +248,14 @@ Key: {{ $cert.Key | b64enc }}
     {{- $found -}}
 {{- end -}}
 
-{{- define "list.intersect" -}}
-     {{ $list := .a }}
-     {{- range .b -}}
-        {{ $list = without $list . }}
-     {{- end -}}
-     {{ $list }}
-{{- end -}}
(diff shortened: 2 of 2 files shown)


traefik/templates/hub-admission-controller.yaml
```diff
@@ -52,159 +52,23 @@ metadata:
...
-  - name: hub-agent.traefik.portal
+  {{- $resources := list
+      (dict "name" "hub-agent.traefik.portal"               "endpoint" "/api-portal"            "resource" "apiportals")
+      (dict "name" "hub-agent.traefik.api"                  "endpoint" "/api"                   "resource" "apis")
+      (dict "name" "hub-agent.traefik.access"               "endpoint" "/api-access"            "resource" "apiaccesses")
+      (dict "name" "hub-agent.traefik.catalog-item"         "endpoint" "/api-catalog-item"      "resource" "apicatalogitems")
+      (dict "name" "hub-agent.traefik.managed-subscription" "endpoint" "/managed-subscription"  "resource" "managedsubscriptions")
+      (dict "name" "hub-agent.traefik.plan"                 "endpoint" "/api-plan"              "resource" "apiplans")
+      (dict "name" "hub-agent.traefik.bundle"               "endpoint" "/api-bundle"            "resource" "apibundles")
+      (dict "name" "hub-agent.traefik.version"              "endpoint" "/api-version"           "resource" "apiversions")
```
mloiseleur, line 63: This is great :+1: !
Do you think you can order this list alphabetically ?

traefik/values.yaml
```diff
@@ -935,6 +935,8 @@ hub:
   # -- Name of `Secret` with key 'token' set to a valid license token.
   # It enables API Gateway.
   token: ""
+  # -- Array of namespaces to watch. If left empty, Traefik watches all namespaces.
+  namespaces: []
```
mloiseleur, line 939: There are already [namespaces in kubernetes CRD](https://doc.traefik.io/traefik/providers/kubernetes-crd/#namespaces), [namespaces in kubernetes Ingress](https://doc.traefik.io/traefik/providers/kubernetes-ingress/#namespaces) and [namespaces in kubernetes Gateway](https://doc.traefik.io/traefik/providers/kubernetes-gateway/#namespaces)

Since hub provider works with all k8s providers, wdyt of adding a warning in NOTES.txt when namespaces does not match ?

jspdown: Yes that's a good idea!

traefik/templates/NOTES.txt
```diff
@@ -59,3 +59,24 @@ for more info. 🚨
...
+        {{- range (list "kubernetesCRD" "kubernetesGateway" "kubernetesIngress") }}
+            {{- $provider := . -}}
+            {{- $providerNamespaces := get (get $.Values.providers .) "namespaces" -}}
+            {{- if $providerNamespaces -}}
+                {{- $intersect := (include "list.intersect" (dict "a" $providerNamespaces "b" $.Values.hub.namespaces)) }}
+                {{- if $intersect }}
+                    {{- printf "WARNING: %s provider is configured to watch namespaces %s but those ones are not watched by Hub. It may be an issue !\n" $provider $intersect -}}
+                {{- end -}}
+            {{- else -}}
+                {{- printf "WARNING: %s provider is configured to watch all namespaces but Hub only watches %s. It may be an issue !\n" $provider $.Values.hub.namespaces -}}
```
mloiseleur, line 75:
```suggestion
                {{- printf "WARNING: %s provider is configured to watch all namespaces but Hub provider only watches %s.\n" $provider $.Values.hub.namespaces -}}
```

traefik/values.yaml
```diff
@@ -935,6 +935,8 @@ hub:
   # -- Name of `Secret` with key 'token' set to a valid license token.
   # It enables API Gateway.
   token: ""
+  # -- Array of namespaces to watch. If left empty, Traefik watches all namespaces.
```
mloiseleur, line 938: I have two comments about this:

1. Current code is also including traefik ns by default in the list. 
2. When rbac.namespaced is set, it's possible to watch all namespaces nonetheless.

darkweaver87: It's like we have already done for RBACs on other providers (Ingress, CRD, Gateway) and I updated documentation accordingly. Even if it's namespaced, users may want to add all namespaces here and restrict it after some tests, so we need to keep the freedom of doing it.

mloiseleur (approved):

darkweaver87 (approved):
LGTM. `no-merge` until `v3.16.0` is not available.

PR #1397, opened 2025-05-05
feat: improve UX for hub token

mmatur:
### What does this PR do?

This PR improves the UX when setting the hub token value.
- If the token is less or equal to 64 characters, then consider as the secret name
- if the token is greater or equal to 65 characters, then consider as the token value

### Motivation

Improve user experience and avoid hidden value

### More


mloiseleur (approved):
LSTM

darkweaver87 (approved):
LGTM

PR #1409, opened 2025-05-13
fix(Traefik Hub): upgrade with --reuse-values

mloiseleur:
### What does this PR do?

When users tries to upgrade with `--reuse-values`, since it does not take into account new values from the new version of the chart, like `hub.namespaces` in **v35.2.0**.

### Motivation

Even if users should not upgrade with `--reuse-values`, this PR improves userXP and allows them to do so for this specific use case.

### More


darkweaver87 (commented):
We won't deal with any values change. How about documenting the usage of `--reset-then-reuse-values` done for this purpose ?

darkweaver87:
Could you add a test to cover this ?

PR #1410, opened 2025-05-15
fix: :bug: ingress route annotations should not be null

darkweaver87:
### More


mloiseleur:
Would you please detail your annotations ?

traefik/tests/dashboard-ingressroute_test.yaml
```diff
@@ -115,17 +114,22 @@ tests:
       path: metadata.annotations
       value:
         kubernetes.io/ingress.class: test-class
-- it: should insert annotations correctly
+- it: should insert annotations correctly (and give precedence to user annotations)
```
mloiseleur, line 117:
```suggestion
- it: should insert annotations correctly and give precedence to user annotations
```

mloiseleur (approved):

PR #1418, opened 2025-05-20
fix(podtemplate): calculate GOMEMLIMIT at 80% of memory limit to reduce OOM

shlomitubul:
### What does this PR do?
chart current pod template injects
GOMEMLIMIT = pod memory limits, it leaves no headroom for Go GC to run before the process hits the cgroup ceiling, leading to OOM kills. In my tests, I show great improvement with lower values, like 80%.

### Motivation

Multiple OOMs that were gone once we changed GOMEMLIMIT value 

### More

### DIFF
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 35.3.0
+version: 35.4.0
 # renovate: image=traefik
 appVersion: v3.4.0
 kubeVersion: ">=1.22.0-0"
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -272,3 +272,42 @@ Key: {{ $cert.Key | b64enc }}
     {{- end }}
     {{- toYaml $diff }}
 {{- end }}
+
+{{/*
+  This helper converts the input value of memory to Bytes.
+  Input needs to be a valid value as supported by k8s memory resource field.
+ */}}
+{{- define "traefik.convertMemToBytes" }}
+  {{- $mem := lower . -}}
+   {{- if hasSuffix "e" $mem -}}
+    {{- $mem = mulf (trimSuffix "e" $mem | float64) 1e18 -}}
+  {{- else if hasSuffix "ei" $mem -}}
+    {{- $mem = mulf (trimSuffix "e" $mem | float64) 0x1p60 -}}
+  {{- else if hasSuffix "p" $mem -}}
+    {{- $mem = mulf (trimSuffix "p" $mem | float64) 1e15 -}}
+  {{- else if hasSuffix "pi" $mem -}}
+    {{- $mem = mulf (trimSuffix "pi" $mem | float64) 0x1p50 -}}
+  {{- else if hasSuffix "t" $mem -}}
+    {{- $mem = mulf (trimSuffix "t" $mem | float64) 1e12 -}}
+  {{- else if hasSuffix "ti" $mem -}}
+    {{- $mem = mulf (trimSuffix "ti" $mem | float64) 0x1p40 -}}
+  {{- else if hasSuffix "g" $mem -}}
+    {{- $mem = mulf (trimSuffix "g" $mem | float64) 1e9 -}}
+  {{- else if hasSuffix "gi" $mem -}}
+    {{- $mem = mulf (trimSuffix "gi" $mem | float64) 0x1p30 -}}
+  {{- else if hasSuffix "m" $mem -}}
+    {{- $mem = mulf (trimSuffix "m" $mem | float64) 1e6 -}}
+  {{- else if hasSuffix "mi" $mem -}}
+    {{- $mem = mulf (trimSuffix "mi" $mem | float64) 0x1p20 -}}
+  {{- else if hasSuffix "k" $mem -}}
+    {{- $mem = mulf (trimSuffix "k" $mem | float64) 1e3 -}}
+  {{- else if hasSuffix "ki" $mem -}}
+    {{- $mem = mulf (trimSuffix "ki" $mem | float64) 0x1p10 -}}
+  {{- end }}
+{{- $mem }}
+{{- end }}
+
+{{- define "traefik.gomemlimit" }}
+{{- $memlimitBytes := include "traefik.convertMemToBytes" . | mulf 0.8 -}}
+{{- printf "%dMiB" (divf $memlimitBytes 0x1p20 | floor | int64) -}}
+{{- end }}
--- a/traefik/tests/pod-config_test.yaml
+++ b/traefik/tests/pod-config_test.yaml
@@ -19,26 +19,23 @@ tests:
               resourceFieldRef:
                 resource: limits.cpu
                 divisor: "1"
-  - it: should have set GOMEMLIMIT with cpu limit
+  - it: should have set GOMEMLIMIT with memory limit
     set:
       resources:
         limits:
-          memory: "4"
+          memory: "4Gi"
     asserts:
       - contains:
           path: spec.template.spec.containers[0].env
           content:
             name: GOMEMLIMIT
-            valueFrom:
-              resourceFieldRef:
-                resource: limits.memory
-                divisor: "1"
+            value: "3276MiB"
   - it: should have not set GOMAXPROCS without cpu limit
     asserts:
       - notEqual:
           path: spec.template.spec.containers[0].env[0].name
           value: GOMAXPROCS
-  - it: should have not set GOMEMLIMIT without cpu limit
+  - it: should have not set GOMEMLIMIT without memory limit
     asserts:
       - notEqual:
           path: spec.template.spec.containers[0].env[0].name
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -857,10 +857,7 @@
           {{- end }}
           {{- if ($.Values.resources.limits).memory }}
           - name: GOMEMLIMIT
-            valueFrom:
-              resourceFieldRef:
-                resource: limits.memory
-                divisor: '1'
+            value: {{ include "traefik.gomemlimit" .Values.resources.limits.memory | quote }}
           {{- end }}
           {{- with .Values.hub.token }}
           - name: HUB_TOKEN


traefik/templates/_helpers.tpl
```diff
@@ -272,3 +272,42 @@ Key: {{ $cert.Key | b64enc }}
     {{- end }}
     {{- toYaml $diff }}
 {{- end }}
+
+{{/*
```
shlomitubul, line 276: I bascily copy-paste this from https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_helpers.tpl#L172

traefik/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traefik
 description: A Traefik based Kubernetes ingress controller
 type: application
-version: 35.3.0
+version: 35.4.0
```
mloiseleur, line 5:
```suggestion
version: 35.3.0
```
Please do not touch the version. Release are done following a specific process.

traefik/templates/_helpers.tpl
```diff
@@ -272,3 +272,42 @@ Key: {{ $cert.Key | b64enc }}
...
+  This helper converts the input value of memory to Bytes.
+  Input needs to be a valid value as supported by k8s memory resource field.
+ */}}
+{{- define "traefik.convertMemToBytes" }}
```
darkweaver87, line 280: This method doesn't cover all K8S valid values, cf. https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

Let's take K8S example:

- 128974848 ✅
- 129e6 ✅
- 129M  ✅
- 128974848000m ❌
- 123Mi ✅

With my suggestions and my tests it works :-)

traefik/templates/_helpers.tpl
```diff
@@ -272,3 +272,42 @@ Key: {{ $cert.Key | b64enc }}
...
+  {{- $mem := lower . -}}
+   {{- if hasSuffix "e" $mem -}}
```
darkweaver87, line 282:
```suggestion
  {{- if hasSuffix "e" $mem -}}
```

traefik/templates/_helpers.tpl
```diff
@@ -272,3 +272,42 @@ Key: {{ $cert.Key | b64enc }}
...
+  {{- else if hasSuffix "t" $mem -}}
+    {{- $mem = mulf (trimSuffix "t" $mem | float64) 1e12 -}}
+  {{- else if hasSuffix "ti" $mem -}}
+    {{- $mem = mulf (trimSuffix "ti" $mem | float64) 0x1p40 -}}
+  {{- else if hasSuffix "g" $mem -}}
+    {{- $mem = mulf (trimSuffix "g" $mem | float64) 1e9 -}}
+  {{- else if hasSuffix "gi" $mem -}}
+    {{- $mem = mulf (trimSuffix "gi" $mem | float64) 0x1p30 -}}
+  {{- else if hasSuffix "m" $mem -}}
+    {{- $mem = mulf (trimSuffix "m" $mem | float64) 1e6 -}}
```
darkweaver87, line 299:
```suggestion
  {{- else if hasSuffix "m" . -}}
    {{- $mem = divf (trimSuffix "m" $mem | float64) 1e3 -}}
  {{- else if hasSuffix "M" . -}}
    {{- $mem = mulf (trimSuffix "m" $mem | float64) 1e6 -}}
```

traefik/tests/pod-config_test.yaml
```diff
@@ -19,26 +19,23 @@ tests:
               resourceFieldRef:
                 resource: limits.cpu
                 divisor: "1"
-  - it: should have set GOMEMLIMIT with cpu limit
+  - it: should have set GOMEMLIMIT with memory limit
```
darkweaver87, line 22:
```suggestion
  - it: should have set GOMEMLIMIT with bytes memory limit
```

traefik/tests/pod-config_test.yaml
```diff
@@ -19,26 +19,23 @@ tests:
...
     set:
       resources:
         limits:
-          memory: "4"
+          memory: "4Gi"
```
darkweaver87, line 26:
```suggestion
          memory: "128974848"
```

traefik/tests/pod-config_test.yaml
```diff
@@ -19,26 +19,23 @@ tests:
...
     asserts:
       - contains:
           path: spec.template.spec.containers[0].env
           content:
             name: GOMEMLIMIT
-            valueFrom:
-              resourceFieldRef:
-                resource: limits.memory
-                divisor: "1"
+            value: "3276MiB"
```
darkweaver87, line 32:
```suggestion
            value: "98MiB"
```

traefik/tests/pod-config_test.yaml
```diff
@@ -19,26 +19,23 @@ tests:
...
   - it: should have not set GOMAXPROCS without cpu limit
```
darkweaver87, line 33: This should cover pretty much every memory input value:
```suggestion
  - it: should have set GOMEMLIMIT with exponential  memory limit
    set:
      resources:
        limits:
          memory: "129e6"
    asserts:
      - contains:
          path: spec.template.spec.containers[0].env
          content:
            name: GOMEMLIMIT
            value: "98MiB"
  - it: should have set GOMEMLIMIT with SI memory limit
    set:
      resources:
        limits:
          memory: "129M"
    asserts:
      - contains:
          path: spec.template.spec.containers[0].env
          content:
            name: GOMEMLIMIT
            value: "98MiB"
  - it: should have set GOMEMLIMIT SI (small unit) with memory limit
    set:
      resources:
        limits:
          memory: "128974848000m"
    asserts:
      - contains:
          path: spec.template.spec.containers[0].env
          content:
            name: GOMEMLIMIT
            value: "98MiB"
  - it: should have set GOMEMLIMIT with IEC memory limit
    set:
      resources:
        limits:
          memory: "123Mi"
    asserts:
      - contains:
          path: spec.template.spec.containers[0].env
          content:
            name: GOMEMLIMIT
            value: "98MiB"
  - it: should have not set GOMAXPROCS without cpu limit
```

jnoordsij (changes requested):
Hey there!

Thanks a lot for this contribution. I'm personally a great fan of the performance and stability improvements that setting `GOMEMLIMIT` introduces, which is one of the reasons [I've submitted various PRs elsewhere](https://github.com/pulls?q=is%3Apr+author%3Ajnoordsij+GOMEMLIMIT) to add this functionality to Helm charts.

As such, I think it's great to allow more flexibility on this from the chart side. Although it does come with some maintenance burden and slightly complicated helpers, I still consider this beneficial to improve a lot of usecases.

However, I do wonder a bit what would be the most suitable 'default' here to pick. The current implementation is of course the simplest, but as you've clearly seen in your use-case, it may not be the best choice for everyone. But I think that using 80% will effectively be the same: it will be suitable to some use-cases, but might also impact some use-cases which are currently doing fine in a negative way. Given that decreasing the limit may result in more gc runs, which may result in more CPU consumption and thus eventually potential slower response times, I think it is essential that we pick a number that is both generic enough, yet still customisable for people aiming to tweak their own setup.

For the choice of default value, in https://github.com/cert-manager/cert-manager/pull/6977#issuecomment-2180602404 I was provided with some great feedback on it. Based on those references, with the (indirectly referenced) [Go documentation itself](https://tip.golang.org/doc/gc-guide#Suggested_uses) being the most relevant in my opinion, I would suggest that probably a value between 90 and 95 would be more accurate as 'default'.

I think a default deviating from the recommended 90-95% or the simple current 100% is something we should only do if we actually have some very clear benchmarks for a representative use-case that suggests another value is more suitable.

So in that regard I would suggest:
- allowing tuning as @darkweaver87 suggested is essential to introduce this change;
- using a default value of 90% to 95% is more suitable based on current knowledge.

shlomitubul:
Hey @jnoordsij, I appreciate your comment and review. I applied @darkweaver87 and your suggestion

traefik/values.yaml
```diff
@@ -889,6 +889,11 @@ serviceAccountAnnotations: {}
 # -- [Resources](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) for `traefik` container.
 resources: {}
 
+# -- Percentage of memory limit to set for GOMEMLIMIT
+# -- set as decimal (0.9 = 90%, 0.95 = 95% etc)
+# -- only takes effect when resources.limits.memory is set
+goMemLimitPercentage: 0.9
```
mloiseleur: @darkweaver87 @jnoordsij Would it make sense for you to move it under `deployment:` ?

jnoordsij: Yeah that sounds reasonable. We could even do `resources` as it's also heavily tied to that, but I think `deployment` makes a little more sense.

traefik/templates/_helpers.tpl
```diff
@@ -275,3 +275,48 @@ Hash: {{ sha1sum ($cert.Cert | b64enc) }}
...
+    {{- $mem = mulf (trimSuffix "ki" $mem | float64) 0x1p10 -}}
+  {{- end }}
+{{- $mem }}
+{{- end }}
+
+{{- define "traefik.gomemlimit" }}
+{{- $percentage := .percentage -}}
+{{- $memlimitBytes := include "traefik.convertMemToBytes" .memory | mulf $percentage -}}
+{{- printf "%dMiB" (divf $memlimitBytes 0x1p20 | floor | int64) -}}
+{{- end }}
```
jnoordsij:
```suggestion
{{- end }}

```
Add EOL here for neatness

traefik/tests/pod-config_test.yaml
```diff
@@ -19,26 +19,67 @@ tests:
...
       - contains:
           path: spec.template.spec.containers[0].env
           content:
             name: GOMEMLIMIT
-            valueFrom:
-              resourceFieldRef:
-                resource: limits.memory
-                divisor: "1"
+            value: "98MiB"
+  - it: should have set GOMEMLIMIT with exponential  memory limit
```
jnoordsij:
```suggestion
  - it: should have set GOMEMLIMIT with exponential memory limit
```

jnoordsij (changes requested):

darkweaver87 (approved):
LGTM

mloiseleur:
@shlomitubul You need to update the schema. See `make schema`

shlomitubul:
> @shlomitubul You need to update the schema. See `make schema`

@mloiseleur done

mloiseleur:
@shlomitubul And also "make docs", to update VALUES.md

shlomitubul:
> @shlomitubul And also "make docs", to update VALUES.md

Done

jnoordsij:
```
  FAIL  Pod configuration	traefik/tests/pod-config_test.yaml
	- should have set GOMEMLIMIT with bytes memory limit

		- asserts[0] `contains` fail
			Template:	traefik/templates/deployment.yaml
			DocumentIndex:	0
			ValuesIndex:	0
			Path:	spec.template.spec.containers[0].env
			Expected to contain:
				- name: GOMEMLIMIT
				  value: 98MiB
			Actual:
				- name: POD_NAME
				  valueFrom:
				    fieldRef:
				      fieldPath: metadata.name
				- name: POD_NAMESPACE
				  valueFrom:
				    fieldRef:
				      fieldPath: metadata.namespace
				- name: GOMEMLIMIT
				  value: 110MiB
```
https://github.com/traefik/traefik-helm-chart/actions/runs/15945655864/job/45047294267?pr=1418#step:8:55

It seems a number of tests are failing; I assume this is due to still using the old `0.8` modifier instead of the new `0.9` default. I think you should probably update the tests to include a fixed value, to prevent it from being susceptible to changes of the default.

shlomitubul:
> ```
>   FAIL  Pod configuration	traefik/tests/pod-config_test.yaml
> 	- should have set GOMEMLIMIT with bytes memory limit
> 
> 		- asserts[0] `contains` fail
> 			Template:	traefik/templates/deployment.yaml
> 			DocumentIndex:	0
> 			ValuesIndex:	0
> 			Path:	spec.template.spec.containers[0].env
> 			Expected to contain:
> 				- name: GOMEMLIMIT
> 				  value: 98MiB
> 			Actual:
> 				- name: POD_NAME
> 				  valueFrom:
> 				    fieldRef:
> 				      fieldPath: metadata.name
> 				- name: POD_NAMESPACE
> 				  valueFrom:
> 				    fieldRef:
> 				      fieldPath: metadata.namespace
> 				- name: GOMEMLIMIT
> 				  value: 110MiB
> ```
> 
> 
>     
>   
> 
> https://github.com/traefik/traefik-helm-chart/actions/runs/15945655864/job/45047294267?pr=1418#step:8:55
> 
> 
> It seems a number of tests are failing; I assume this is due to still using the old `0.8` modifier instead of the new `0.9` default. I think you should probably update the tests to include a fixed value, to prevent it from being susceptible to changes of the default.

@jnoordsij  updated the tests, it seem all tests are passed now, thank you for taking the time and review this !

PR #1407, opened 2025-05-13
feat(CRDs): ✨ update CRDs for Traefik Proxy v3.4.x

mloiseleur:
### What does this PR do?

Update Traefik Proxy CRDs with the required content for v3.4.x

### Motivation

Traefik Proxy v3.4.0 has been released.

### More

- [ ] Yes, I updated the tests accordingly
- [ ] Yes, I updated the schema accordingly
- [ ] Yes, I ran `make test` and all the tests passed

PR #1422, opened 2025-05-21
fix(CRDs): validation check on RootCA for both servertransport

LinAnt:
### What does this PR do?

Fixes faulty validation check

https://github.com/traefik/traefik/issues/11774

Fixes a bug introduced by this PR: https://github.com/traefik/traefik-helm-chart/pull/1407


rtribotte:
Hi,

JSYK, we also fixed this on Traefik side: https://github.com/traefik/traefik/pull/11775.

mloiseleur:
@LinAnt I see a diff between the content of this PR and the content of [upstream PR](https://github.com/traefik/traefik/pull/11775).  

Would you please update this PR to use the same rule (and so the same CRD) ?

LinAnt:
Will do :+1:

mloiseleur (approved):

PR #1425, opened 2025-05-24
fix(Traefik Proxy): supported `ingressRoute.*.annotations` breaks templating

SIMULATAN:
### What does this PR do?

Fix a regression introduced by #1410 causing templating to fail when the `annotations` field is unset in `ingressRoute` children:
```
Error: Error: template: traefik/templates/ingressroute.yaml:7:57: executing "traefik/templates/ingressroute.yaml" at <$config.annotations>: wrong type for value; expected map[string]interface {}; got interface {}
```

Example `values.yml` reproducer:
```yml
ingressRoute:
  api:
    enabled: true
```
(of course, this is nonsensical without additional fields, adding them would just introduce more boilerplate without affecting this particular issue)

### Motivation

[Renovate-created bump PR failed the CI workflow](https://github.com/SIMULATAN/k8s-ops/pull/171), after further investigation, this was caused by [my separate `api` ingress](https://github.com/SIMULATAN/k8s-ops/blob/a2b423e2c43837dbc08a23bca9b9c70f22c5a6d0/traefik/traefik.values.yml#L45-L58) missing the `annotations` field (which is set to `{}` by default for the ingresses present in the default `values.yaml`)

### More

- [ ] Yes, I updated the tests accordingly *(there is no test for additional `ingressRoute`s yet, should I create one?)*

### DIFF
--- a/traefik/tests/unsupported-ingressroute_test.yaml
+++ b/traefik/tests/unsupported-ingressroute_test.yaml
@@ -1,0 +1,15 @@
+suite: unsupported IngressRoute configuration
+templates:
+- ingressroute.yaml
+tests:
+- it: should not fail when enabling an unsupported IngressRoute
+  set:
+    ingressRoute:
+      api:
+        enabled: true
+  asserts:
+  - hasDocuments:
+      count: 1
+  - equal:
+      path: metadata.name
+      value: RELEASE-NAME-api
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -185,6 +185,7 @@ gatewayClass:  # @schema additionalProperties: false
   # -- Additional gatewayClass labels (e.g. for filtering gateway objects by custom labels)
   labels: {}
 
+# -- Only dashboard & healthcheck IngressRoute are supported. It's recommended to create workloads CR outside of this Chart.
 ingressRoute:
   dashboard:
     # -- Create an IngressRoute for the dashboard


jnoordsij:
Hey there! Thanks for your contribution. I haven't had the full time to dive into the details, but the kind of error you are mentioning immediately made me think of some similar issues I've seen for the most recent Helm release (3.18.0): https://github.com/helm/helm/issues/30878.

Just for reference and to help us ensure this bug is actually caused by the mentioned chart change, can you:
- list the Helm version you are using when you observe this error (`helm version`)
- if you are using 3.18.0 and it is possible for you, downgrade to 3.17.x and see if the problem still persists?

I saw that at https://github.com/traefik/traefik-helm-chart/issues/1427#issuecomment-2910026537 someone has already experienced differing behavior on Helm versions, so probably we should wait till Helm releases 3.18.1 (which they aim to do soon according to above issue) with some fixes before we try and find a solution to fix problems here.

jnoordsij:
Thanks for the quick update and the time you spend debugging this! It seems indeed your issue happens independent of the Helm version, where the issue described in #1427 happens partially due to changes in Helm behavior and partially due to reasons similar to your use-case.

We as chart maintainers will further discuss on how to approach dealing with this PR and hope to give you an update soon!

mloiseleur:
@SIMULATAN Following discussion with all the maintainers, we are ready to consider this PR. 

I pushed a review commit where I:

1. added a documentation tip in values explaining it's unsupported 
2. added a test to ensure your use case won't fail again in next releases

Wdyt ? Anything we missed ?

mloiseleur (approved):

PR #1430, opened 2025-05-30
fix(Traefik Hub): add strict check on admission cert

zalbiraw:
Add strict cert check to prevent issues when lookup fails because of server side mode or dry run

### What does this PR do?

Ensures that the value returned from the secret look up has the correct keys inside of it

### Motivation

Cause issues for ArgoCD deployments that tried to ignore changes on the cert. The issue happens because the helm template is not run on the server and is not aware of the secrets available. Therefore, it tries to re-create certs which causes ArgoCD to go out of sync and cause an infinite cycle of unnecessary restarts 

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -219,7 +219,10 @@ Key: {{ index $.Values.hub.apimanagement.admission.customWebhookCertificate "tls
 Hash: {{ sha1sum (index $.Values.hub.apimanagement.admission.customWebhookCertificate "tls.crt") }}
 {{- else -}}
     {{- $cert := lookup "v1" "Secret" (include "traefik.namespace" .) $.Values.hub.apimanagement.admission.secretName -}}
-    {{- if and $cert (hasKey $cert.data "tls.crt") (hasKey $cert.data "tls.key") -}}
+    {{- if $cert }}
+        {{ if or (not (hasKey $cert.data "tls.crt")) (not (hasKey $cert.data "tls.key")) -}}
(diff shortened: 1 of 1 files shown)


mloiseleur (approved):

traefik/templates/_helpers.tpl
```diff
@@ -219,7 +219,7 @@ Key: {{ index $.Values.hub.apimanagement.admission.customWebhookCertificate "tls
 Hash: {{ sha1sum (index $.Values.hub.apimanagement.admission.customWebhookCertificate "tls.crt") }}
 {{- else -}}
     {{- $cert := lookup "v1" "Secret" (include "traefik.namespace" .) $.Values.hub.apimanagement.admission.secretName -}}
-    {{- if $cert -}}
+    {{- if and $cert (hasKey $cert.data "tls.crt") (hasKey $cert.data "tls.key") -}}
```
darkweaver87: While I would tend to agree on this, I think this is very intrusive.
With this test, we know for sure this secret doesn't contain any certificate and if somehow the user provide an existing secret name by mistake for `hub.apimanagement.admission.secretName`, this code will simply erase this secret.
I think the helm chart should just fail at this point.
```suggestion
    {{- if $cert }}
        {{ if or (not (hasKey $cert.data "tls.crt")) (not (hasKey $cert.data "tls.key")) -}}
            {{- fail (printf "ERROR: secret %s/%s exists but doesn't contain any certificate data. Please remove it or change hub.apimanagement.admission.secretName." (include "traefik.namespace" .) $.Values.hub.apimanagement.admission.secretName) }}
        {{- end -}}
```

mloiseleur:
@zalbiraw We may make a release today. Do you think you can take into account @darkweaver87 review today, so we can include this PR can be included in this release ?

zalbiraw:
Sorry @mloiseleur I was away last week. No rush on this but its good to go now.

mloiseleur (approved):

darkweaver87 (approved):
LGTM

PR #1441, opened 2025-06-10
fix(schema): 🐛 allow additional properties on `global`

### LINKED ISSUE #1439: New global settings + schema validation prevent installation
### Welcome!

36.0.0

default

Being able tp upgrade

```
Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
traefik:
- global: Additional property uniqueRelease is not allowed
- global: Additional property uniqueReleaseCustomValues is not allowed
```

### What are your values ?

Our charts use a lot the global settings, and changes linked to #1419 and pull request #1436 blocks the deployment of our charts.

### Additional Information

```shell

```

mloiseleur:
### What does this PR do?

Allow additional properties on `global`

### Motivation

Fixes #1439

### More


mloiseleur:
It's possible to rename `global.checkNewVersion` and `global.sendAnonymousUsage`.

For `global.azure.xxx`, AFAIK, we cannot. It's an hard requirement from Microsoft. See [here](https://learn.microsoft.com/en-us/partner-center/marketplace-offers/azure-container-technical-assets-kubernetes?tabs=linux%2Clinux2#update-the-helm-chart).

jnoordsij:
Ahhh that makes a lot of sense, I only looked at https://github.com/traefik/traefik-helm-chart/pull/1436 and didn't realize the `global` option was already introduced prior to that.

Then this is all fine and we'll keep Azure keys as is, but maybe it's still a little neater to rename the `global.checkNewVersion` and `global.sendAnonymousUsage` keys at some point.

mloiseleur:
`global.checkNewVersion` and `global.sendAnonymousUsage` are the also the name used in Traefik Proxy configuration. 
Since we are trying to align helm chart values with upstream configuration as much as we can, so IMHO it's better to keep it like this, **or** to rename it on both sides, in Proxy configuration and in this chart.

jnoordsij:
Yeah I agree it's a hard trade-off. It feels a bit scary to possibly intertwine with the magic `global` key, but probably it shouldn't break anyways, and people are probably not likely to use the same exact subkeys for some other purpose.

We may also opt to put it under e.g. `deployment.global`, but that also feels wrong.

All-in-all I have a slight preference for changing it, but also fully fine with leaving it as is.

mloiseleur:
I'll let @darkweaver87 review this PR and share its though, then.

darkweaver87 (approved):
LGTM

PR #1447, opened 2025-06-11
feat: handle hub version with Traefik v3.4

lbenguigui:
### What does this PR do?

This PR updates the Traefik Hub / Traefik Proxy version mapping.

### Motivation

Keep the mapping up to date.

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -185,7 +185,9 @@ It requires a dict with "Version" and "Hub".
         {{- $hubProxyVersion = "v3.1" }}
      {{- else if semverCompare "<v3.10.2-0" $version -}}
         {{ $hubProxyVersion = "v3.2" }}
-     {{- else if semverCompare "<v3.17.0-0" $version -}}
+     {{- else if semverCompare "<v3.15.3-0" $version -}}
+        {{ $hubProxyVersion = "v3.3" }}
+     {{- else if or (semverCompare "=v3.16.0" $version) (semverCompare "=v3.16.1" $version) -}}
         {{ $hubProxyVersion = "v3.3" }}
      {{- end -}}
    {{- end -}}


traefik/templates/_helpers.tpl
```diff
@@ -185,7 +185,9 @@ It requires a dict with "Version" and "Hub".
         {{- $hubProxyVersion = "v3.1" }}
      {{- else if semverCompare "<v3.10.2-0" $version -}}
         {{ $hubProxyVersion = "v3.2" }}
-     {{- else if semverCompare "<v3.17.0-0" $version -}}
+     {{- else if semverCompare "<v3.15.3-0" $version -}}
+        {{ $hubProxyVersion = "v3.3" }}
+     {{- else if or (semverCompare "=v3.16.0" $version) (semverCompare "=v3.16.1" $version) -}}
```
mloiseleur, line 190:
```suggestion
     {{- else if semverCompare "<v3.16.2-0" -}}
```

lbenguigui: It is done on purpose. If I apply your suggestion, `v3.15.3` will be considered based on `v3.3` which is false.

mloiseleur: You're right :+1: . 
It seems it's possible to compare multiple version in the same expr, like ">= 1.2 < 3.0.0 || >= 4.2.3" ([source](https://helm.sh/docs/chart_template_guide/function_list/#basic-comparisons))
I'll close this suggestion and make a new one. 

Do you think you can add tests, too ? It seems required now that we are introducing this level of complexity.

mloiseleur, line 190:
```suggestion
     {{- else if semverCompare "<v3.15.3-0 || >= v3.16.0 < v3.16.2" $version -}}
```

lbenguigui: I don't get why this version is better, I prefer my version actually

darkweaver87, line 190:
```suggestion
     {{- else if semverCompare "<=v3.16.1" $version -}}
```

darkweaver87 (approved):
LGTM

traefik/templates/_helpers.tpl
mloiseleur: I'll try to explain it. 
FTM, one condition == one version. So it's quite simple.
With multiple conditions for the same version, it makes it more complex.

nmengin: @mloiseleur I agree with @lbenguigui, having one condition per Hub minor version to check seems more understandable for maintenance.

mloiseleur: ok, let's start with this. We will see if we need to refactor it or not.

PR #1452, opened 2025-06-14
feat(deployment): Allow omitting replicas via null and support zero replicas

### LINKED ISSUE #1451: Add an option to disable rendering the 'replicas' field in Deployment
### Welcome!

My goal is to manage the Traefik `Deployment` using **Argo Rollouts** for progressive delivery.

To do this effectively, the Argo Rollouts controller must have exclusive control over the pod replicas. The ideal practice for this is to completely delegate replica management to the `Rollout` resource. This means the `replicas` field should not be present in the original `Deployment` manifest, allowing the Rollout controller to manage it without any conflicting instructions.

However, the Traefik Helm chart currently always renders the `replicas` field in the `Deployment` manifest. This can lead to conflicts or unpredictable behavior when an external controller like Argo Rollouts tries to manage the same field.

This feature request is to **add a switch to optionally disable the rendering of the `replicas` field**, providing a cleaner integration path.

### The Context: The Problem in Detail

The core issue is that the `replicas` field is always rendered, due to this line in `traefik/templates/deployment.yaml`:

https://github.com/traefik/traefik-helm-chart/blob/8873fd68ec36561b8853fb357848f2423e597fc9/traefik/templates/deployment.yaml#L39

Even if we could set the value to `0`, the presence of the `replicas: 0` field itself is an explicit instruction. A cleaner approach is to omit the field entirely when its management is delegated.

### The Idea: Proposed Solution

I propose introducing a new boolean value in the chart's `values.yaml`, for example `deployment.manageReplicas`.

  * **`deployment.manageReplicas` (default: `true`):** When `true` or unset, the chart behaves exactly as it does now, maintaining backward compatibility.
  * **`deployment.manageReplicas: false`:** When set to `false`, the `replicas` field is completely omitted from the rendered `Deployment` manifest.

This could be implemented with a simple conditional block in the template:

```diff
# proposed change in deployment.yaml
{{- if not .Values.autoscaling.enabled }}
{{- if .Values.deployment.manageReplicas | default true }}
replicas: {{ default 1 .Values.deployment.replicas }}
{{- end }}
{{- end }}
```

This approach provides the necessary flexibility for GitOps and progressive delivery tools without affecting existing users.

### How to Reproduce (Example Use Case)

With the proposed change, the following Argo CD `Application` manifest would work as intended. By setting `deployment.manageReplicas: false`, we signal that Argo Rollouts is now in charge of the replica count.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: traefik
  namespace: argocd
spec:
  destination:
    namespace: traefik
    server: https://kubernetes.default.svc
  source:
    repoURL: https://traefik.github.io/charts
    targetRevision: 36.1.0
    chart: traefik
    helm:
      parameters:
        - name: deployment.manageReplicas # New parameter
          value: "false"
      values: |-
        extraObjects:
          - apiVersion: argoproj.io/v1alpha1
            kind: Rollout
            metadata:
              name: traefik-rollout
            spec:
              workloadRef:
                apiVersion: apps/v1
                kind: Deployment
                name: traefik
              strategy:
                canary:
                  steps:
                  - setWeight: 10
                  - pause:
                      duration: 5m
          - apiVersion: autoscaling/v2
            kind: HorizontalPodAutoscaler
            metadata:
              name: traefik-rollout-hpa
            spec:
              scaleTargetRef:
                apiVersion: argoproj.io/v1alpha1
                kind: Rollout
                name: traefik-rollout
              minReplicas: 5
              maxReplicas: 50
              metrics:
              - type: Resource
                resource:
                  name: cpu
                  target:
                    type: Utilization
                    averageUtilization: 80
  project: traefik
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
      - CreateNamespace=true
      - ServerSideApply=true
```

Thank you for considering this enhancement. This flexibility would be a great help for users adopting modern GitOps and progressive delivery patterns.

na2na-p:
closes  #1451.

## Summary

This PR enhances the `deployment.replicas` logic to support two key use cases without introducing a new parameter:

1.  **Delegating Replica Management:** Setting `replicas: null` now omits the `replicas` field from the rendered `Deployment` manifest. This is essential for users who manage pod replicas with external controllers like **Argo Rollouts**.
2.  **Scaling to Zero:** Setting `replicas: 0` is now correctly handled, rendering `replicas: 0` in the manifest to allow scaling down to zero pods.

## Background and Motivation

As originally described in \#1451, there was a need to omit the `replicas` field to allow progressive delivery tools like Argo Rollouts to manage the replica count without conflict.

The initial proposal in the issue was to add a new `deployment.manageReplicas` boolean flag. However, a simpler approach was suggested in the issue discussion: **allowing `deployment.replicas` to be set to `null`**.

This PR adopts that simpler approach. The key challenge was to implement it in a way that differentiates `null` (omit the field) from `0` (render `replicas: 0`), as a simple truthiness check (`if .Values.deployment.replicas`) would treat both as falsy.

## Implementation Details

To solve this, this PR changes the condition for rendering the `replicas` field from the original `default` function to an explicit non-`nil` check.

**Original code in `deployment.yaml`:**

```yaml
{{- if not .Values.autoscaling.enabled }}
replicas: {{ default 1 .Values.deployment.replicas }}
{{- end }}
```

**Proposed change in this PR:**

```yaml
{{- if and (not .Values.autoscaling.enabled) (not (eq .Values.deployment.replicas nil)) }}
replicas: {{ .Values.deployment.replicas }}
{{- end }}
```

This implementation correctly handles both use cases while maintaining backward compatibility for existing users and avoiding the introduction of a new values parameter.


na2na-p:
I have tested the following scenarios and confirmed they work as intended:

## A standard Deployment managed by an HPA.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: traefik-testing
  namespace: argocd
  labels:
    ManagedBy: CRD
spec:
  destination:
    namespace: traefik-testing
    server: https://kubernetes.default.svc
  source:
    repoURL: https://github.com/na2na-p/traefik-helm-chart.git
    targetRevision: feat/deployment-allow-null-replicas
    path: traefik
    helm:
      parameters:
        - name: deployment.replicas
          value: "null"
      values: |-
        autoscaling:
          enabled: true
          maxReplicas: 50
          minReplicas: 5
  project: default
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
      - CreateNamespace=true
      - ServerSideApply=true
```
image


## Setting deployment.replicas: 0 while using Argo Rollouts.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: traefik-testing-with-rollouts
  namespace: argocd
  labels:
    ManagedBy: CRD
spec:
  destination:
    namespace: traefik-testing-with-rollouts
    server: https://kubernetes.default.svc
  source:
    repoURL: https://github.com/na2na-p/traefik-helm-chart.git
    targetRevision: feat/deployment-allow-null-replicas
    path: traefik
    helm:
      parameters:
        - name: deployment.replicas
          value: "0"
      values: |-
        extraObjects:
          - apiVersion: argoproj.io/v1alpha1
            kind: Rollout
            metadata:
              name: traefik-testing-with-rollouts
            spec:
              workloadRef:
                apiVersion: apps/v1
                kind: Deployment
                name: traefik-testing-with-rollouts
              strategy:
                canary:
                  steps:
                  - setWeight: 10
                  - pause:
                      duration: 5m
          - apiVersion: autoscaling/v2
            kind: HorizontalPodAutoscaler
            metadata:
              name: traefik-rollout-hpa
            spec:
              scaleTargetRef:
                apiVersion: argoproj.io/v1alpha1
                kind: Rollout
                name: traefik-testing-with-rollouts
              minReplicas: 5
              maxReplicas: 50
              metrics:
              - type: Resource
                resource:
                  name: cpu
                  target:
                    type: Utilization
                    averageUtilization: 80
  project: default
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
      - CreateNamespace=true
      - ServerSideApply=true
```
image

mloiseleur:
Thanks. It's missing a test when replicas is nullified.
Unless I missed something, the HPA can be configured with _values_, no need to use _extraObjects_.

Instead of hiding it in this PR, wdyt about adding an EXAMPLE on this use case ?

na2na-p:
@mloiseleur 

Thanks to comments!

---

> It's missing a test when replicas is nullified.

Thanks for pointing that out! I've added the test case.

---

> Unless I missed something, the HPA can be configured with values, no need to use extraObjects.

Good point. Unless I'm mistaken, my understanding is that the `apiVersion` and `kind` are hardcoded here and cannot be changed through values:
https://github.com/traefik/traefik-helm-chart/blob/8873fd68ec36561b8853fb357848f2423e597fc9/traefik/templates/hpa.yaml#L20-L21

Additionally, this Helm chart is designed for the proxy workload to run as either a Deployment or a DaemonSet, so I believe there's little need to modify them.

---

> Instead of hiding it in this PR, wdyt about adding an EXAMPLE on this use case ?

That's a great suggestion, thank you! I'll work on adding an `Install with Argo Rollouts` example in this same PR.

I'll start on the modifications now, so please give me a little time.

EXAMPLES.md
```diff
@@ -76,6 +76,53 @@ autoscaling:
         averageUtilization: 80
 ```
 
+# Install with Argo Rollouts
+
+When using [ArgoCD Rollouts](https://argoproj.github.io/rollouts/), one can delegate replica management to a `Rollout` resource, enabling progressive delivery strategies like canary and blue-green deployments.
+To do so, `replicas` should be set to `0` and the `Rollout` resource defined in `extraObjects`.
```
na2na-p: I'm not a native English speaker, so I would appreciate it if you could point out any errors. 
(Translate assisted with Gemini😇)

EXAMPLES.md
```diff
@@ -76,6 +76,53 @@ autoscaling:
...
+
+```yaml
+deployment:
+  replicas: 0
+
+extraObjects:
+  - |
```
na2na-p:
```yaml
extraObjects:
  - apiVersion: argoproj.io/v1alpha1
    kind: Rollout
    metadata:
      name: {{ .Release.Name }}
```

This causes error ↓
```bash
helm install traefik-helm-install-testing .   
Error: INSTALLATION FAILED: cannot load values.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{".Release.Name":interface {}(nil)}
```

mloiseleur:
> Unless I'm mistaken, my understanding is that the `apiVersion` and `kind` are hardcoded here and cannot be changed through values

\o that's right. I opened #1453 in order to add this feature. 
Feel free to review and test it.

na2na-p:
@mloiseleur 

I'm planning to add conditional logic based on scaleTargetRef.apiVersion and scaleTargetRef.kind, so I think it would be best for me to rebase after #1453 is merged first.

It worked successfully after I applied this fix based on your code.
This works with the configuration below.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: traefik-testing-with-rollouts
  namespace: argocd
spec:
  destination:
    namespace: traefik-testing-with-rollouts
    server: https://kubernetes.default.svc
  source:
    repoURL: https://github.com/na2na-p/traefik-helm-chart.git
    targetRevision: feat/deployment-allow-zero-replicas-with-configuable-hpa
    path: traefik
    helm:
      values: |-
        deployment:
          replicas: 0
        autoscaling:
          enabled: true
          minReplicas: 5
          maxReplicas: 50
          metrics:
            - type: Resource
              resource:
                name: cpu
                target:
                  type: Utilization
                  averageUtilization: 80
          scaleTargetRef:
            apiVersion: argoproj.io/v1alpha1
            kind: Rollout
        extraObjects:
          - |
            apiVersion: argoproj.io/v1alpha1
            kind: Rollout
            metadata:
            spec:
              workloadRef:
                apiVersion: apps/v1
                kind: Deployment
              strategy:
                canary:
                  steps:
                  - setWeight: 10
                  - pause:
                      duration: 5m
  project: default
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
      - CreateNamespace=true
      - ServerSideApply=true

```

image

EXAMPLES.md
```diff
@@ -76,6 +76,46 @@ autoscaling:
...
+      workloadRef:
+        apiVersion: apps/v1
+        kind: Deployment
+        name: "{{ template \"traefik.fullname\" . }}"
+      strategy:
+        canary:
+          steps:
+          - setWeight: 10
+          - pause:
+              duration: 5m
```
na2na-p: I had tested that

image
image
image
image

EXAMPLES.md
```diff
@@ -76,6 +76,46 @@ autoscaling:
         averageUtilization: 80
 ```
 
+# Install with Argo Rollouts
+
+When using [ArgoCD Rollouts](https://argoproj.github.io/rollouts/), one can delegate replica management to a `Rollout` resource, enabling progressive delivery strategies like canary and blue-green deployments.
+To do so, `replicas` should be set to `0` and the `Rollout` resource defined in `extraObjects`.
```
```suggestion
When delegating replica management, `replicas` should be set to `0` and the `Rollout` resource can be defined in a separate YAML or in `extraObjects`.
```

EXAMPLES.md
mloiseleur: See my suggestion

EXAMPLES.md
mloiseleur: BTW, are you sure it's recommended to set it to 0 ? Shouldn't it be set to null ?

na2na-p: Thank you for your comment. That's a valid question.

My understanding is that setting `replicas` to `0` is the correct approach, and I have two main reasons for this.

First, the official Argo Rollouts documentation on migration specifically mentions this practice:
> ...manually scale down an existing Deployment by changing replicas field of an existing Deployment to zero.

(Source: https://argo-rollouts.readthedocs.io/en/stable/migrating/)

Second, if `replicas` is set to `null` or the field is omitted from the Deployment manifest (and no HPA is active), Kubernetes will automatically default the replica count to `1`. Setting it explicitly to `0` ensures that the Rollout resource takes full control of the replicas, preventing Kubernetes from creating an unintended pod.
(Source: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#replicas)

Hope this clarifies my reasoning. Please let me know if you have a different perspective!

mloiseleur: Thanks, it's more clear to me now.

EXAMPLES.md
```diff
@@ -76,6 +76,46 @@ autoscaling:
         averageUtilization: 80
 ```
 
+# Install with Argo Rollouts
+
+When using [ArgoCD Rollouts](https://argoproj.github.io/rollouts/), one can delegate replica management to a `Rollout` resource, enabling progressive delivery strategies like canary and blue-green deployments.
+When delegating replica management, `replicas` should be set to `0` and the `Rollout` resource can be defined in a separate YAML or in `extraObjects`.
```
```suggestion
In order to delegate replica management, `deployment.replicas` should be set to `0` and the `Rollout` resource can be defined in a separate YAML or in `extraObjects`.
```

na2na-p: Thanks!

mloiseleur:
cc @jnoordsij @darkweaver87 for 2nd review

EXAMPLES.md
```diff
@@ -76,6 +76,46 @@ autoscaling:
...
+# Install with Argo Rollouts
+
+When using [ArgoCD Rollouts](https://argoproj.github.io/rollouts/), one can delegate replica management to a `Rollout` resource, enabling progressive delivery strategies like canary and blue-green deployments.
+In order to delegate replica management, `deployment.replicas` should be set to `0` and the `Rollout` resource can be defined in a separate YAML or in `extraObjects`.
+
+```yaml
+deployment:
+  replicas: 0
+autoscaling:
+	enabled: true
```
jnoordsij: nitpick: indentation differs here (tab vs. space I assume); maybe make it consistent to improve readability?

traefik/tests/deployment-config_test.yaml
```diff
@@ -30,6 +30,52 @@ tests:
...
+      autoscaling:
+        enabled: true
+        scaleTargetRef:
+          apiVersion: apps/v1
+          kind: Deployment
+          name: '{{ template "traefik.fullname" . }}'
+    asserts:
+      - notExists:
+          path: spec.replicas
+  - it: should have replicas when autoscaling is enabled for another resource
```
jnoordsij:
```suggestion
  - it: should have zero replicas by default when autoscaling is enabled for another resource
```

EXAMPLES.md
na2na-p: Thanks!

darkweaver87 (approved):
LGTM

jnoordsij (approved):

Issue #1457, opened 2025-06-18
Support "Selector" for allowed routes in the Gateway API entity.


ajago:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What do you want to achieve with this Chart ?

I want to be able to configure allowed routes using a selector with this chart. For example like this page in the docs - https://gateway-api.sigs.k8s.io/guides/multiple-ns/#shared-gateway

The current functionality only lets the user configure the `from` field in the namespaces section. This works for `Same` and `All` values. However it is not flexible enough to use the selector functionality, where a `selector` field is also needed.

My initial suggestion is to repurpose the `namespaces` value to be an object to pass in the values needed. This is simple but is a breaking change. If this is not acceptable then I would be happy to take guidance on the best way forward with this.

Example of what I mean - https://github.com/ajago/traefik-helm-chart/pull/1

mloiseleur:
At first glance, it makes sense to me. 
Feel free to open this PR on this repository.
Do you think it should support templating, too ?

ajago:
I didn't think it needed to support further templating. The Selector can be either matchExpressions or matchLabels, so it didn't seem worth putting this behind a layer of abstraction.

Will open up a PR now

mloiseleur:
Part of v37.0.0

PR #1465, opened 2025-06-25
feat!: Support Selector for namespace policy in Gateway API

ajago:
### What does this PR do?

Can now enable the Gateway API object to specify allowed namespaces based on user defined selector.

### Motivation

The current helm chart only supports enabling the Gateway for the same namespace as the gateway API or all namespaces.
This change gives more flexibility to specify allowed namespaces based on a selector. 

### More


ajago:
Fixes #1457

mloiseleur:
@ajago Would you please run `make docs` and commit the result ? It should put the ci to green and then we can merge it.

@jnoordsij The new doc check is working :) !

mloiseleur:
@ajago We will probably make a new release next week, so in order to this PR to be a part of it, I've run the `make docs` and pushed the commit.

ajago:
@mloiseleur sorry just seen this. Thanks for sorting out the doc updates!

PR #1468, opened 2025-07-01
chore(release): 🚀 publish 36.2.1

mloiseleur:
### More


CONTRIBUTING.md
```diff
@@ -118,6 +118,27 @@ Several implementations have been experimented. Here are pros and cons of each:
 
 Consequently, we chose the last option, until the situation evolve on Helm side.
 
+# Release process
```
darkweaver87, line 121: WDYT about a implementing a `make release` command instead ?
At least let's track this idea.

PR #1470, opened 2025-07-01
fix(Deployment): revision history should be disableable

### LINKED ISSUE #1283: revisionHistoryLimit cannot be set to 0
### Welcome!

33.2.1

3.2.2

### What did you do?

In [values.yaml#L26](https://github.com/traefik/traefik-helm-chart/blob/v33.2.1/traefik/values.yaml#L26) the `revisionHistoryLimit` is defined with `type:[integer, null];minimum:0`.

In [deployment.yaml#L41](https://github.com/traefik/traefik-helm-chart/blob/v33.2.1/traefik/templates/deployment.yaml#L41) the conditional `if .Values.deployment.revisionHistoryLimit` is false, when setting the `revisionHistoryLimit` to 0.
A workaround would be to use "0" (as string) but the schema does not allow a string in that case. Therefore `revisionHistoryLimit` cannot be set to 0.
Another solution could be to remove the if-clause and to define a default instead (e.g. the Kubernetes default of 10). 

Relates to helm issue [#3164](https://github.com/helm/helm/issues/3164#issuecomment-437531151).

### What did you see instead?

When setting `revisionHistoryLimit` to 0, the revisionHistoryLimit is not part of the rendered manifest.

``` 
  ...
  deployment:
    enabled: true
    kind: Deployment
    revisionHistoryLimit: 0
    ...
```

### Additional Information

```shell

```

bpsoraggi:
### What does this PR do?

Allows ``revisionHistoryLimit`` to be set to 0

### Motivation

Fixes #1283 

### More


mloiseleur (approved):
LGTM

traefik/tests/deployment-config_test.yaml
```diff
@@ -84,6 +84,14 @@ tests:
       - equal:
           path: spec.revisionHistoryLimit
           value: 1
+  - it: should have revisionHistoryLimit set to 0 when specified as 0
```
darkweaver87, line 87: Please add a test showing `revisionHistoryLimit` is not set when `.Values.deployment.revisionHistoryLimit` is not set (which is the purpose of this PR).
It probably needs to be located in `container-config_test.yaml`.

PR #1471, opened 2025-07-02
fix(entryPoint): allow scheme to be unset on redirect

### LINKED ISSUE #1345: helm chart shouldn't be check for scheme on redirect
https://github.com/traefik/traefik-helm-chart/blob/4b54ebaa3dd039edd68e3db6c8decee9d95c34a7/traefik/templates/_podtemplate.tpl#L611

The helm chart shouldn't need to force this value to be set as the default when unset is correct.

bpsoraggi:
### What does this PR do?

Allows ``scheme`` to not be set when redirect on websecure

### Motivation

Fixes #1345 

### More


mloiseleur (approved):
The test are complete. Thanks :+1:

PR #1477, opened 2025-07-10
feat(hub): offline mode

jspdown:
### What does this PR do?

This PR brings support for Traefik Hub API Management offline mode, enabling deployments in air-gapped environments without external connectivity requirements. To achieve this, the following changes have been implemented:

- Added `APIPortalAuth` and `APIAuth` CRDs to enable configuration of authentication methods on Portals and individual APIs respectively. These new resources provide granular control over authentication strategies at both the portal and API levels.
- Disabled admission-related configurations, including webhook creation and secret This allows the system to operate without requiring external validation endpoints.
- Implemented the `--hub.offline` flag activation when the `hub.offline` option is set to `true`, providing explicit declaration of offline mode operation.

The `hub.offline` option enhances the user experience by allowing explicit configuration of the deployment mode. While the `hub.token` JWT contains an `offline` claim indicating the license type, the `--hub.offline={boolean-value}` flag ensures clear intent and proper validation between the license type and deployment configuration. This prevents configuration mismatches and provides better operational clarity for offline deployments.

### Motivation

Allow to run Traefik Hub API Management disconnected from the Traefik Hub platform.

### More

### DIFF
--- a/traefik/tests/rbac-config_test.yaml
+++ b/traefik/tests/rbac-config_test.yaml
@@ -1144,9 +1144,9 @@ tests:
               - managedsubscriptions
               - managedapplications
             verbs:
+              - get
               - list
               - watch
-              - get
               - create
               - update
               - patch


traefik/values.yaml
```diff
@@ -974,7 +974,7 @@ hub:
   # It enables API Gateway.
   token: ""
   # -- Disables all external network connections.
-  offline: false
+  offline:  # @schema type:[boolean, null]
```
mloiseleur, line 977: @jspdown The CI is failing because of this change. Would you please detail why you set it from false to nil by default ?

traefik/templates/rbac/clusterrole.yaml
```diff
@@ -240,11 +242,13 @@ rules:
     verbs:
       - list
       - watch
+      - get
```
mloiseleur:
```suggestion
      - get
      - list
      - watch
```

PR #1482, opened 2025-07-22
feat: update helm template to provide a spec that users can define fo…

yquansah:
…r a service

### What does this PR do?

This PR creates a way for users to define `spec.ports` under the `service` key for overriding the default `ports`.

### Motivation

The motivation of this change is to fully support automated workflows for `NodePort` services. My use case is that I provision infrastructure for a `kubeadm` cluster on cloud providers and I need for a way to know what `NodePort` traefik is listening on up front so that I can run automated configuration (Terraform in this case) for cloud load balancers to forward traffic onto traefik.

Currently, since the `NodePort` is selected randomly, once the `Service` is created in a Kubernetes cluster, I have to opt for provisioning my cloud load balancer after all my other infrastructure is provisioned once the `NodePort` is chosen, and that can be quite cumbersome.

My values look like this:

```
gateway:
  listeners:
    websecure:
      port: 8443
      protocol: HTTPS
      namespacePolicy: All
      mode: Terminate
      certificateRefs:
        - name: yoofionline-tls
          namespace: default
providers:
  kubernetesIngress:
    enabled: false
  kubernetesGateway:
    enabled: true
service:
  type: NodePort
  spec:
    ports:
    - nodePort: 31000
      name: web
      targetPort: web
      protocol: TCP
    - nodePort: 31500
      name: websecure
      targetPort: websecure
      protocol: TCP
```

And with that I can know that `web` and `websecure` are listening on known ports which is useful for the automation.

### More

### DIFF
--- a/traefik/templates/_service.tpl
+++ b/traefik/templates/_service.tpl
@@ -26,7 +26,10 @@
   loadBalancerClass: {{ . }}
   {{- end}}
   {{- with .service.spec }}
-  {{- toYaml . | nindent 2 }}
+  {{- $specWithoutPorts := omit . "ports" }}
+  {{- if $specWithoutPorts }}
+  {{- toYaml $specWithoutPorts | nindent 2 }}
+  {{- end }}
   {{- end }}
   selector:
     {{- include "traefik.labelselector" .root | nindent 4 }}
--- a/traefik/templates/service.yaml
+++ b/traefik/templates/service.yaml
@@ -54,7 +54,7 @@ metadata:
   {{- end }}
 spec:
   {{- template "traefik.service-spec" (dict "root" $ "service" $service) }}
-  {{- if $service.spec.ports }}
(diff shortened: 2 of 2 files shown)


mloiseleur:
@yquansah With current code and values like this:

```yaml
ports:
  web:
    nodePort: 31000
  websecure:
    nodePort: 31500

service:
  type: nodePort
```

It generates the expected `Service` resource (helm template...) :

```yaml
spec:
  type: nodePort
  selector:
    app.kubernetes.io/instance: test-flux-system
  ports:
  - port: 80
    name: web
    targetPort: web
    protocol: TCP
    nodePort: 31000
  - port: 443
    name: websecure
    targetPort: websecure
    protocol: TCP
    nodePort: 31500
```

Without the need of adding this new code, does it solve your issue or did I missed something ?

mloiseleur:
No answers since last month, so I close it.
Feel free to re-open or open a new one if needed.

Issue #1308, opened 2025-01-14
[CRD] Support Gateway API Experimental CRDs


smoench:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Since there is a opportunity of installing CRDs separately via a helm chart, I was wondering if it would make sense to support the Gateway API experimental channel as well. Traefik has an opt-in configuration and even needs the experimental channel CRDs to be applied -> https://doc.traefik.io/traefik/providers/kubernetes-gateway/#experimentalchannel

I would suggest to introduce a new option `gatewayAPIExperimentalChannel` (or similar) to the CRD helm chart. This option would win over the existing `gatewayAPI`-option when both are activated. It wouldn't make sense to install both at the same time even it won't break anything as far as the experimental channels crds are installed last

mloiseleur:
Hello @smoench ,

It's true that CRDs of experimental channel could be installed with the CRD Chart. 

Feel free to open a PR if you need it.

discostur:
https://github.com/traefik/traefik-helm-chart/issues/1253

PR #1485, opened 2025-07-24
add gatewayAPIExperimental

discostur:
### What does this PR do?

- adds the ability to install `gatewayAPIExperimentalChannel`
- you can switch weather you want to install standard or experimental channel
- i added the gateway api release v1.2.1 (like standard channel); might be updated soon because 1.3.0 should be supported with traefik 3.5.0

### Motivation

- fix https://github.com/traefik/traefik-helm-chart/issues/1308

### DIFF
--- a/traefik-crds/templates/crds.yaml
+++ b/traefik-crds/templates/crds.yaml
@@ -7,5 +7,9 @@
 {{- end }}
 
 {{- if eq (.Values.gatewayAPI | toString) "true" -}}
-{{ include "traefik-crds.render-crds" (dict "scope" . "path" "crds-files/gatewayAPI/*.yaml") }}
+{{ include "traefik-crds.render-crds" (dict "scope" . "path" "crds-files/gatewayAPI/gateway-standard-install.yaml") }}
+{{- end }}
+
+{{- if eq (.Values.gatewayAPIExperimental | toString) "true" -}}
+{{ include "traefik-crds.render-crds" (dict "scope" . "path" "crds-files/gatewayAPI/gateway-experimental-install.yaml") }}
 {{- end }}
--- a/traefik-crds/Changelog.md
+++ b/traefik-crds/Changelog.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## 1.10.0  ![Kubernetes: >=1.22.0-0](https://img.shields.io/static/v1?label=Kubernetes&message=%3E%3D1.22.0-0&color=informational&logo=kubernetes) ![Helm: v3](https://img.shields.io/static/v1?label=Helm&message=v3&color=informational&logo=helm)
+
+**Release date:** TBD
+
+* feat(CRDs): add support for experimental Gateway API CRDs installation
+
+
 ## 1.9.0  ![Kubernetes: >=1.22.0-0](https://img.shields.io/static/v1?label=Kubernetes&message=%3E%3D1.22.0-0&color=informational&logo=kubernetes) ![Helm: v3](https://img.shields.io/static/v1?label=Helm&message=v3&color=informational&logo=helm)
 
 **Release date:** 2025-06-23
--- a/traefik-crds/VALUES.md
+++ b/traefik-crds/VALUES.md
@@ -30,6 +30,7 @@ Kubernetes: `>=1.22.0-0`
 | deleteOnUninstall | bool | `false` | Set it to true if you want to uninstall CRDs when uninstalling this chart. By default, CRDs will be kept so your custom resources will not be deleted accidentally. |
 | enabled | bool | `true` | Field that can be used as a condition when this chart is a dependency. This definition is only here as a placeholder such that it is included in the json schema. See https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags for more info. |
 | gatewayAPI | bool | `false` | Set it to true to install GatewayAPI CRDs. Needed if you set providers.kubernetesGateway.enabled to true in main chart |
+| gatewayAPIExperimental | bool | `false` | Set it to true to install experimental GatewayAPI CRDs. This includes additional experimental features beyond the standard Gateway API |
 | global | object | `{}` | Global values This definition is only here as a placeholder such that it is included in the json schema. |
 | hub | bool | `false` | Set it to true to install Traefik Hub CRDs. Needed if you set hub.enabled to true in main chart |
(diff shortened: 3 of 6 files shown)


mloiseleur:
Thanks @discostur for this PR.
We have a consistency check between CRDs of regular Chart and CRDs of this additional Chart. 

You'll need to update this CI check accordingly, and also ensure version are consistent (ie: versions are the same between experimental and standard) 

Would you please also update to Gateway API v1.3.0 ?

discostur:
@mloiseleur version is updated to v1.3.0; i'm not sure if i understand it right but i updates the consistency script to support the check between standard and experimental CRDs (in both charts, traefik and traefik-crd).

traefik-crds/kustomization.yaml
```diff
@@ -2,7 +2,7 @@
 kind: Kustomization
 apiVersion: kustomize.config.k8s.io/v1beta1
 resources:
-  # curl -o crds-files/gatewayAPI/gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml
+  # curl -o crds-files/gatewayAPI/gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml
```
bpsoraggi, line 5:
```suggestion
  # curl -o crds-files/gatewayAPI/gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml
  # curl -o crds-files/gatewayAPI/gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/experimental-install.yaml
```

traefik-crds/templates/crds.yaml
```diff
@@ -7,5 +7,9 @@
 {{- end }}
 
 {{- if eq (.Values.gatewayAPI | toString) "true" -}}
-{{ include "traefik-crds.render-crds" (dict "scope" . "path" "crds-files/gatewayAPI/*.yaml") }}
+{{ include "traefik-crds.render-crds" (dict "scope" . "path" "crds-files/gatewayAPI/gateway-standard-install.yaml") }}
+{{- end }}
+
+{{- if eq (.Values.gatewayAPIExperimental | toString) "true" -}}
+{{ include "traefik-crds.render-crds" (dict "scope" . "path" "crds-files/gatewayAPI/gateway-experimental-install.yaml") }}
```
bpsoraggi: This throws an error when both `gatewayAPI` and `gatewayAPIExperimental` are set to true. You can swap the order if you want to follow the approach suggested in the original issue (#1308)
```suggestion
{{ include "traefik-crds.render-crds" (dict "scope" . "path" "crds-files/gatewayAPI/gateway-standard-install.yaml") }}

{{- else if eq (.Values.gatewayAPIExperimental | toString) "true" -}}
{{ include "traefik-crds.render-crds" (dict "scope" . "path" "crds-files/gatewayAPI/gateway-experimental-install.yaml") }}
```

traefik/crds/kustomization.yaml
```diff
@@ -2,7 +2,7 @@
 kind: Kustomization
 apiVersion: kustomize.config.k8s.io/v1beta1
 resources:
-  # curl -o gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml
+  # curl -o gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml
```
bpsoraggi, line 5:
```suggestion
  # curl -o gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml
  # curl -o crds-files/gatewayAPI/gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/experimental-install.yaml
```

bpsoraggi (changes requested):
Could you please add some tests? Otherwise looks mostly good to me

discostur:
@bpsoraggi 
* thanks for your suggestions, all accepted
* tests added
* fixed the if condition that experimental wins over standard if both are active

traefik-crds/kustomization.yaml
```diff
@@ -2,7 +2,8 @@
...
+  # curl -o crds-files/gatewayAPI/gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/experimental-install.yaml
```
bpsoraggi:
```suggestion
  # curl -o crds-files/gatewayAPI/gateway-experimental-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/experimental-install.yaml
```

traefik/crds/kustomization.yaml
```diff
@@ -2,7 +2,8 @@
...
+  # curl -o crds-files/gatewayAPI/gateway-standard-install.yaml -L https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/experimental-install.yaml
```
bpsoraggi: My bad, I realize we don't need this one here
```suggestion
```

discostur:
@bpsoraggi fixed

mloiseleur:
Overall, the PR LGTM. I have concerns about the _values_. 

Unless I missed something, I understand that with this implementation, user can enable both standard and experimental channel. When there is a breaking change in experimental channel, it may lead to severe consequences.

@discostur Would you please update the PR to avoid that ? I see two options: 

1. Turn gatewayAPI setttings into a list of two options (standard and experimental)
2. Add requirements (like those on [traefik config](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/requirements.yaml)) on values to fail gracefully before it's applied when user enable both channels

discostur:
@mloiseleur ok i added a new commit with the changes you requested: 

- gatewayAPI settings is now a list of two options (standard and experimental)
- it can either be standard or experimental active at the same time 
- if both are active you get an error printed and the chart does not render
- unit tests adjusted

mloiseleur:
@discostur You just need to run `make docs`, add updated files and the CO should go green.

mloiseleur (approved):
LGTM.
Thanks :+1:

PR #1488, opened 2025-07-29
chore(release): 🚀 Publish 37.0.0 and 1.10.0

bpsoraggi:
### What does this PR do?

Release a new version

### Motivation

Release of Traefik Proxy v3.5.0

### More


traefik-crds/Chart.yaml-e
```diff
@@ -0,0 +1,27 @@
+apiVersion: v2
```
mloiseleur: @bpsoraggi I guess this file is unintended and should be removed from this PR ?

traefik/Chart.yaml-e
```diff
@@ -0,0 +1,37 @@
+apiVersion: v2
```
mloiseleur: Same here

mloiseleur (approved):

PR #1489, opened 2025-07-29
fix(deployment): allow to disable checkNewVersion via values.yaml

### LINKED ISSUE #1449: Unable to disable version check with (new) `global.checkNewVersion` value
### Welcome!

36.1.0

3.4.1

Setting value `global.checkNewVersion=false` should disable actual check.

Traefik still performs new version check.

### What are your values ?

```yaml
global:
  checkNewVersion: false
```

### Additional Information

Introduced at https://github.com/traefik/traefik-helm-chart/pull/1436, default `global.checkNewVersion=true`. When flipping this value to `false`, the chart removes `--global.checkNewVersion` from list of container args. But according to https://doc.traefik.io/traefik/reference/static-configuration/cli/ `--global.checknewversion` is `true` by default, so Traefik still attempts to perform version check.

KyriosGN0:
Signed-off-by: AvivGuiser <avivguiser@gmail.com>

### What does this PR do?

This PR adds the option to disable CheckNewVersion flag via the chart's values.yaml file 

### Motivation

fixes #1449

### More

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -182,8 +182,10 @@
           {{- end }}
         args:
           {{- with .Values.global }}
-           {{- if .checkNewVersion }}
-          - "--global.checkNewVersion"
+           {{- if ne .checkNewVersion nil }}
+          - "--global.checkNewVersion={{ .checkNewVersion | toString }}"
+           {{- end }}
+           {{- if ne .checkNewVersionInterval nil }}
            {{- end }}
            {{- if .sendAnonymousUsage }}
           - "--global.sendAnonymousUsage"
--- a/traefik/tests/deployment-config_test.yaml
+++ b/traefik/tests/deployment-config_test.yaml
@@ -232,7 +232,7 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].args
           value:
-            - --global.checkNewVersion
+            - --global.checkNewVersion=true
             - --entryPoints.metrics.address=:9100/tcp
             - --entryPoints.traefik.address=:8080/tcp
             - --entryPoints.web.address=:8000/tcp


traefik/templates/_podtemplate.tpl
```diff
@@ -182,8 +182,10 @@
           {{- end }}
         args:
           {{- with .Values.global }}
-           {{- if .checkNewVersion }}
-          - "--global.checkNewVersion"
+           {{- if ne .checkNewVersion nil }}
+          - "--global.checkNewVersion={{ .checkNewVersion | toString }}"
+           {{- end }}
+           {{- if ne .checkNewVersionInterval nil }}
```
jnoordsij:
```suggestion
```

KyriosGN0: @jnoordsij done, plus added sendAnonymousUsage as well

mloiseleur (changes requested):
PR LGTM for checkNewVersion.
Would you please remove the changes on _sendAnonymousUsage_ ?
It's already disabled by default, and can be enabled if the user set it, so the current code covers already all case.

KyriosGN0:
@mloiseleur, Thanks for the review, updated the PR!

mloiseleur (approved):

KyriosGN0:
hey @bpsoraggi Thanks for the review, tests added!

ChristianCiach:
> Would you please remove the changes on sendAnonymousUsage ?
> It's already disabled by default, and can be enabled if the user set it, so the current code covers already all case.

I disagree with that. By the same logic the chart should only render the `--global.checknewversion=false` parameter if `global.checkNewVersion=false` has been set (and should continue to render nothing for `true` or `null`), but it will now render `--global.checknewversion=true` or `--global.checknewversion=false` depending on the configuration. This is now inconsistent with `global.sendAnonymousUsage`. I think this PR was better and more future-proof before https://github.com/traefik/traefik-helm-chart/pull/1489/commits/d59bb39e772a01c337c01ef7433b3e24287b94fc. Previously this PR would continue to work if Traefik decides to flip the defaults again.

mloiseleur:
@ChristianCiach We're a small team here and the code of this chart is already quite big.

=> That's why we try our best to reduce or avoid unnecessary code when possible
=> Feel free to help us maintain it

On this specific field, as you can see in [the PR](https://github.com/traefik/traefik/pull/2447) linked to [the proposal](https://github.com/traefik/traefik/issues/2369), it's disabled by default since the _beginning_. 

If there is no more GDPR **and** a decision to switch yes, of course, this chart would change accordingly. 

About your point on consistency, _all_ Traefik Proxy parameters have default value. To be consistent, we would need to display all those parameters in the generated YAML (!) That would be very verbose and also unnecessary: users have the _values_, the _examples_ and the upstream documentation to answer their questions about this topic. 

@ChristianCiach Does that answer or address your concern ?

ChristianCiach:
@mloiseleur Thank you for your quick response. I hope my initial comment didn't come across as rude, because I didn't intent it to.

I agree with everything you wrote in your comment. 

> To be consistent, we would need to display all those parameters in the generated YAML (!) That would be very verbose and also unnecessary

I understand that, but this is even more reason to only render `--global.checknewversion=false` if `global.checkNewVersion=false` has been configured and to render *nothing* otherwise, to fall back to Traefik's default configuration.

In other words: Just as `--global.sendanonymoususage=true` should only be rendered if the chart value `global.checknewversion=true` is set, the same should be done with `checkNewVersion`, but adjusted to Traefik's default value. I think this should be consistent with what you just said, no? I would be ready to submit a PR for that.

mloiseleur:
Yes, it would be (more) consistent. Feel free to open a PR about this.

PR #1490, opened 2025-07-29
fix(CI): fix helm signing

bpsoraggi:
### What does this PR do?

Fixes [error](https://github.com/traefik/traefik-helm-chart/actions/runs/16599938448/job/46956469627) when signing helm chart 

### Motivation

Fix helm chart signing

### More

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -116,7 +116,7 @@ jobs:
         if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
 
       - name: Publish Helm chart
-        uses: bpsoraggi/helm-gh-pages@ae39446ad80fcb6ec81b125f3ce3da8890751297
+        uses: bpsoraggi/helm-gh-pages@144dd827bf5cd326785b15fa1385ffeb5dfd7eb7
         with:
           token: ${{ secrets.CHARTS_TOKEN }}
           charts_dir: .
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -128,10 +128,10 @@ jobs:
           index_dir: .
           commit_username: traefiker
           commit_email: 30906710+traefiker@users.noreply.github.com
-          key: ${{ steps.gpg.outputs.name }}
+          key: ${{ steps.gpg.outputs.email }}
           private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           passphrase: ${{ secrets.GPG_PASSPHRASE }}
-        if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
+        if: steps.tag_exists.outputs.TAG_EXISTS == 'false' && steps.gpg.outputs.email != ''
```
mloiseleur:
```suggestion
        if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
```

.github/workflows/release.yml
```diff
@@ -264,10 +264,10 @@ jobs:
...
```
mloiseleur:
```suggestion
        if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
```

mloiseleur (commented):
To avoid creating release and fail on upload, wdyt of moving "import GPG key" upper in the steps and fail if there is no email ?

PR #1491, opened 2025-07-30
chore(docs): update readme verification section

bpsoraggi:
### What does this PR do?

Updates the `README` verification section

### Motivation

Update the instructions for verifying the helm chart

### More

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,19 +88,48 @@ helm list # should display two charts installed
 
 ## Verification
 
-Each release of the chart is signed using *provenance files*.
-You can verify the chart by following the instructions below, depending on the registry.
+Each release of the chart is signed using *provenance files*. You can verify the chart by following the instructions below:
+
+### 1. Download the Public Signing Key
+
+To download the official Traefik Helm chart signing key, run:
+
+```shell
+gpg --keyring $HOME/.gnupg/traefik.pubring.kbx --keyserver hkps://keys.openpgp.org --no-default-keyring --receive-keys 'B0FBA7678F685E9B7024B79FFD92BB57C5A71A99'
+```
+
+Example output:
+
+```shell
+gpg: key FD92BB57C5A71A99: public key "TraefikLabs Chart Signing Key <noreply@traefik.io>" imported
+gpg: Total number processed: 1
+gpg:               imported: 1
+```
+
+This command downloads the key and stores it in a separate keyring. To add the key to your default keyring, omit the ``--keyring <keyring> --no-default-keyring`` gpg options when running the command.
+
+### 2. Export the Signing Key
+
+By default, GnuPG v2 stores keyrings in a format that is not compatible with Helm chart provenance verification. Before you can verify a Helm chart, you must convert your keyrings to the legacy format:
+
+```shell
+gpg --export --output $HOME/.gnupg/traefik.pubring.gpg  --keyring $HOME/.gnupg/traefik.pubring.kbx  --no-default-keyring 'B0FBA7678F685E9B7024B79FFD92BB57C5A71A99'
+```
+
+### 3. Verify the Chart
+
+To verify the chart, run the appropriate command for your registry:
 
 - OCI Registry
 
   ```shell
-  helm pull --verify verify oci://ghcr.io/traefik/helm/traefik:<VERSION>
+  helm fetch --verify --keyring $HOME/.gnupg/traefik.pubring.gpg oci://ghcr.io/traefik/helm/traefik:<VERSION>
   ```
 
 - Helm Registry (GH Pages)
 
   ```shell
-  helm fetch --verify <REPO>/traefik --version <VERSION>
+  helm fetch --verify --keyring $HOME/.gnupg/traefik.pubring.gpg traefik/traefik --version <VERSION>
   ```
 
 ## Upgrading


README.md
```diff
@@ -88,19 +88,48 @@ helm list # should display two charts installed
 
 ## Verification
 
-Each release of the chart is signed using *provenance files*.
-You can verify the chart by following the instructions below, depending on the registry.
+Each release of the chart is signed using *provenance files*. You can verify the chart by following the instructions below:
```
mloiseleur:
```suggestion
Starting with 37.0.0, chart releases are signed using *provenance files*. One can verify the chart by following the instructions below:
```

README.md
```diff
@@ -88,19 +88,48 @@ helm list # should display two charts installed
...
+
+Example output:
+
+```shell
+gpg: key FD92BB57C5A71A99: public key "TraefikLabs Chart Signing Key <noreply@traefik.io>" imported
+gpg: Total number processed: 1
+gpg:               imported: 1
+```
+
+This command downloads the key and stores it in a separate keyring. To add the key to your default keyring, omit the ``--keyring <keyring> --no-default-keyring`` gpg options when running the command.
```
mloiseleur:
```suggestion
This command downloads the key and stores it in a separate keyring. To add the key to your default keyring, omit the `--no-default-keyring` gpg options when running the command.
```

bpsoraggi:
> The provided commands worked on my side. Out of curiosity, why do you recommend to use `--no-default-keyring` on each command ?

@mloiseleur It can be done either way, I just generally think separation of concerns is usually best. I searched a bit about what the best practices are and couldn't find a definitive answer, so honestly we could have it either way 🤷‍♀️

With GnuPG, if the default home directory doesn't exist, the `common.conf` file is created with `use-keyboxd` (which was the case for me), but the commands with the separate keyring work for either case

README.md
bpsoraggi: Why keep the ``--keyring <keyring>`` op?

README.md
bpsoraggi: Just saw we have a few inconsistencies between the use of 'you' or 'one':

> One can check what has changed in the [Changelog](./traefik/Changelog.md).

> You can customize the install with a `values` file. There are some [EXAMPLES](./EXAMPLES.md) provided.

> ... etc.

Should I change it all to 'one'?

darkweaver87: Yes, we use to "impersonate" messages. In this case you can also say:
```suggestion
Starting with 37.0.0, chart releases are signed using *provenance files*. Those can be verified using the following the instructions below:
```

README.md
darkweaver87: Do we really care ? I see it like apt repo keys. Once you use a repo you generally add the key in trust DB. Then, even if you don't use the repo anymore, you usually don't take the time to remove it (which is not a good practice). Here I think people who will want to bother with this will have enough knowledge to do it by themselves.

````suggestion
```shell
gpg --keyserver hkps://keys.openpgp.org --receive-keys 'B0FBA7678F685E9B7024B79FFD92BB57C5A71A99'
```

Example output:

```shell
gpg: key FD92BB57C5A71A99: public key "TraefikLabs Chart Signing Key <noreply@traefik.io>" imported
gpg: Total number processed: 1
gpg:               imported: 1
```

This command downloads TraefikLabs public key and adds it to `~/.gnupg/pubring.kbx`.
````

File: README.md [hunk 3]
```diff
@@ -88,19 +88,48 @@ helm list # should display two charts installed
...
+
+### 2. Export the Signing Key
+
+By default, GnuPG v2 stores keyrings in a format that is not compatible with Helm chart provenance verification. Before you can verify a Helm chart, you must convert your keyrings to the legacy format:
+
+```shell
+gpg --export --output $HOME/.gnupg/traefik.pubring.gpg  --keyring $HOME/.gnupg/traefik.pubring.kbx  --no-default-keyring 'B0FBA7678F685E9B7024B79FFD92BB57C5A71A99'
```
darkweaver87: Same here:
````suggestion
```shell
gpg --export --output $HOME/.gnupg/traefik.pubring.gpg 'B0FBA7678F685E9B7024B79FFD92BB57C5A71A99'
````

File: README.md (hunk 1 above)
Comment (mloiseleur) [reply README.md]: That's a good question, I do not have a strong opinion on this so I digged a bit. 
Following [diataxis](https://diataxis.fr/how-to-guides/), we should use _conditional imperatives_. So something like: _To verify the chart, follow the instructions below_
Wdyt ?

File: README.md [hunk 4]
```diff
@@ -88,19 +88,48 @@ helm list # should display two charts installed
...
-Each release of the chart is signed using *provenance files*.
-You can verify the chart by following the instructions below, depending on the registry.
+Each release of the chart is signed using *provenance files*. You can verify the chart by following the instructions below:
+
+### 1. Download the Public Signing Key
+
+To download the official Traefik Helm chart signing key, run:
+
+```shell
+gpg --keyring $HOME/.gnupg/traefik.pubring.kbx --keyserver hkps://keys.openpgp.org --no-default-keyring --receive-keys 'B0FBA7678F685E9B7024B79FFD92BB57C5A71A99'
```
mloiseleur:
```suggestion
gpg --keyring $HOME/.gnupg/traefik.pubring.kbx  --no-default-keyring --keyserver hkps://keys.openpgp.org --receive-keys 'B0FBA7678F685E9B7024B79FFD92BB57C5A71A99'
```

README.md
mloiseleur: Ah :sweat_smile: I did not see the `--keyring` option in the command before, because of different order.

README.md
bpsoraggi: I also don't have a strong opinion about the language itself, as long as it's consistent. Out of the 3, I think the last option (what Michel suggested) might sound the best

README.md
bpsoraggi: > Here I think people who will want to bother with this will have enough knowledge to do it by themselves.

Fair enough, I agree

PR #1492, opened 2025-08-04
feat(plugins): support ConfigMap inline local plugins alongside hostPath

cilasbeltrame:
### What does this PR do?

This PR adds ConfigMap support for local plugins in the `experimental.localPlugins` configuration, providing an alternative to the existing `hostPath` approach. The enhancement allows users to embed plugin source code directly in their Helm values, which the chart automatically converts into Kubernetes ConfigMaps and mounts into Traefik containers.

Key Changes:

- New inlinePlugin field for embedding plugin source code in `values.yaml` with automatic ConfigMap generation from inline plugin content (up to 1MiB - etcd limit)
- Backward compatibility - existing hostPath configurations continue to work unchanged
- Support for both Deployment and DaemonSet deployment modes

Example configuration:
```
experimental:
  localPlugins:
    prodplugin:
      moduleName: github.com/example/prodplugin
      mountPath: /plugins-local/src/github.com/example/prodplugin
      inlinePlugin:
        go.mod: |
          module github.com/example/prodplugin
          go 1.23
        .traefik.yml: |
          displayName: Your Traefik Plugin
          type: middleware
          import: github.com/example/prodplugin
          // Your plugin settings
        main.go: |
          package main
          // Your plugin implementation
```
### Motivation

The current hostPath approach for local plugins has several limitations in containerized environments:
Problems with hostPath:
- Node dependency - Plugin source must be available on every node where Traefik pods might be scheduled, also difficult to maintain consistency across nodes.
- Poor GitOps integration - Plugin source code stored outside of Helm values
- In case, plugins are huge, they can be still used with hostPath backward compatibility.

### More

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -1573,3 +1573,70 @@ hub:
     enabled: true
     maxRequestBodySize: 10485760 # optional, default to 1MiB
 ```
+
+## Deploy multiple Gateways with a single Traefik Deployment/DaemonSet
+
+This example exposes two Gateways (e.g., `internal` and `external`) from a single Traefik installation.
+
+```yaml
+# Ports for external gateway
+ports:
+  web-ext:
+    port: 9080
+    exposedPort: 80
+    expose:
+      external: true
+  websecure-ext:
+    port: 9443
+    exposedPort: 443
+    expose:
+      external: true
+
+gateway:
+  enabled: true
+  name: traefik-internal
+
+gatewayClass:
+  enabled: true
+
+service:
+  additionalServices:
+    external:
+      type: LoadBalancer
+
+providers:
+  kubernetesGateway:
+    enabled: true
+    statusAddress:
+      service:
+        enabled: false
+```
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: traefik-external
+spec:
+  gatewayClassName: traefik
+  listeners:
+    - name: web-ext
+      protocol: HTTP
+      port: 9080
+      allowedRoutes:
+        namespaces:
+          from: All
+    # Comment out if you have a valid TLS certificate
+    # - name: websecure-ext
+    #   protocol: HTTPS
+    #   port: 9443
+    #   allowedRoutes:
+    #     namespaces:
+    #       from: All
+    #   tls:
+    #     mode: Terminate
+    #     certificateRefs:
+    #       - group: ""
+    #         kind: Secret
+    #         name: some-tls-cert
+```
--- a/traefik/Chart.yaml
+++ b/traefik/Chart.yaml
@@ -4,7 +4,7 @@ description: A Traefik based Kubernetes ingress controller
 type: application
 version: 37.3.0
 # renovate: image=traefik
-appVersion: v3.6.0
+appVersion: v3.6.2
 kubeVersion: ">=1.22.0-0"
 keywords:
   - traefik
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -1,6 +1,6 @@
 # traefik
 
-![Version: 37.3.0](https://img.shields.io/badge/Version-37.3.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v3.6.0](https://img.shields.io/badge/AppVersion-v3.6.0-informational?style=flat-square)
+![Version: 37.3.0](https://img.shields.io/badge/Version-37.3.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v3.6.2](https://img.shields.io/badge/AppVersion-v3.6.2-informational?style=flat-square)
 
 A Traefik based Kubernetes ingress controller
 


mloiseleur (approved):
PR looks technically correct. 
I'll need review of other chart maintainers, just to be sure I'm not missing anything here.

EXAMPLES.md
```diff
@@ -586,6 +586,32 @@ experimental:
...
+          module github.com/example/prodplugin
+          go 1.23
+        .traefik.yml: |
+          displayName: Your Traefik Plugin
+          type: middleware
+          // Your plugin settings
+          import: github.com/example/prodplugin
+        main.go: |
+          package main
+          // Your plugin implementation
```
mloiseleur: Would you please change this example to a working example ? Something like an hello world plugin ?

cilasbeltrame: @mloiseleur Just did, thank you for looking into it.

cilasbeltrame: Also pasting the hello world working example here for convenience:

```yaml
experimental:
  localPlugins:
    helloworld-plugin:
      moduleName: github.com/example/helloworldplugin
      mountPath: /plugins-local/src/github.com/example/helloworldplugin
      inlinePlugin:
        go.mod: |
          module github.com/example/helloworldplugin

          go 1.23
        .traefik.yml: |
          displayName: Hello World Plugin
          type: middleware

          import: github.com/example/helloworldplugin

          summary: |
            This is a simple plugin that prints "Hello, World!" to the response.

          testData:
            message: "Hello, World!"
        main.go: |
          package helloworldplugin

          import (
            "context"
            "net/http"
          )

          type Config struct{}

          func CreateConfig() *Config {
            return &Config{}
          }

          type HelloWorld struct {
            next http.Handler
          }

          func New(ctx context.Context, next http.Handler, config *Config, name string) (http.Handler, error) {
            return &HelloWorld{next: next}, nil
          }

          func (h *HelloWorld) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
            rw.Write([]byte("Hello, World!"))
            h.next.ServeHTTP(rw, req)
          }
```

traefik/templates/_podtemplate.tpl
```diff
@@ -934,10 +934,18 @@
         {{- if .Values.deployment.additionalVolumes }}
           {{- toYaml .Values.deployment.additionalVolumes | nindent 8 }}
         {{- end }}
+        {{- $root := . }}
```
jnoordsij: This hack seems a little weird to me, given you can use `$` here (see also https://helm.sh/docs/chart_template_guide/variables/#:~:text=However,%20there%20is%20one%20variable%20that%20will%20always%20point%20to%20the%20root%20context). However I just realized this is a pattern used elsewhere in this chart, so probably we should update it there. I'll follow up with a PR to correct it elsewhere. Here, you should be able to remove this and just use `$`.

traefik/templates/_podtemplate.tpl
```diff
@@ -934,10 +934,18 @@
...
         {{- range $localPluginName, $localPlugin := .Values.experimental.localPlugins }}
```
jnoordsij, line 944: I find the ordering of volumes a bit weird looking at it now; probably the plugin-related volumes, i.e. this range and the `/plugins-storage` mount, should be grouped. However that is probably more of a follow-up thing to clean.

traefik/templates/_podtemplate.tpl
```diff
@@ -934,10 +934,18 @@
...
         - name: {{ $localPluginName | replace "." "-" }}
+          {{- if $localPlugin.hostPath }}
           hostPath:
             path: {{ $localPlugin.hostPath | quote }}
+          {{- else if $localPlugin.inlinePlugin }}
+          configMap:
+            name: {{ template "traefik.fullname" $root }}-local-plugin-{{ $localPluginName | replace "." "-" }}
```
jnoordsij: Given this is re-used naming (used here and in config maps), maybe make it a helper? E.g. `traefik.localpluginmapname`?

traefik/templates/_podtemplate.tpl
```diff
@@ -934,10 +934,18 @@
         {{- if .Values.deployment.additionalVolumes }}
           {{- toYaml .Values.deployment.additionalVolumes | nindent 8 }}
         {{- end }}
+        {{- $root := . }}
         {{- range $localPluginName, $localPlugin := .Values.experimental.localPlugins }}
         - name: {{ $localPluginName | replace "." "-" }}
+          {{- if $localPlugin.hostPath }}
           hostPath:
             path: {{ $localPlugin.hostPath | quote }}
+          {{- else if $localPlugin.inlinePlugin }}
```
jnoordsij: In terms of values, I find it to be slightly weird that now basically it would be allowed to set both `hostPath` and `inlinePlugin`, which will (silently) result in just using the former.

However, I think it's probably still fine, as the alternatives I can come up with are not much better:
- add a required `type` value to localPlugin = bad because it's a breaking change and introduces an overhead field
- add an optional `type` value which defaults to `host` = bad because overhead field
- create an entire new `inlinePlugins` fields = bad because of a lot of duplicated code or merging is required
- add an explicit check in the first case that errs if both values are set = bad because a bit ugly

@mloiseleur WDYT?

jnoordsij (changes requested):
Thanks for this, I think this is a neat thing to add, especially for people employing small plugins!

From what I can see this is all technically sound, just some coding style questions.

traefik/templates/_podtemplate.tpl
jnoordsij: See #1495

traefik/templates/_podtemplate.tpl
cilasbeltrame: Thanks, @jnoordsij — agreed. My goal here was to avoid any breaking changes. Since localPlugin only landed in v37.0.0, it may not be widely used yet, but changing its shape now would still be a breaking change. One alternative would be keeping the current key path and introduce the “more organized” structure alongside it, then deprecate the old one gradually. Not sure how this kind of situation is handled for this helm chart.

traefik/templates/_podtemplate.tpl
cilasbeltrame: Later on, I'm happy to work on this in a separate PR.

traefik/templates/_podtemplate.tpl
cilasbeltrame: thank you, changed in bc2af2f778e5609b7b0e9de7fe693a92f28c928b

traefik/templates/_podtemplate.tpl
cilasbeltrame: Added helper in bc2af2f778e5609b7b0e9de7fe693a92f28c928b, as it was used the another helper in the naming, I had to provide a full context to the helper, pls review it.

traefik/templates/_podtemplate.tpl
mloiseleur: @cilasbeltrame What kind of "more organized" structure you have in mind ? 
On helm chart, it's quite rare to maintain both solution and deprecate. 

Usually, we:
1. Advertise the breaking changes: major version + upgrade instructions
2. Add a check in `requirements.yaml` that will fail gracefully in case user did not notice the change despite 1. 

See here for an [example PR](https://github.com/traefik/traefik-helm-chart/pull/961/) of breaking change about plugins.

cilasbeltrame: I was thinking to implement what @jnoordsij said:
*  having a `typePlugin` with the options of: `inline`, `host` and possibly a `localPath`, cause having hostPath is not that well recommended, especially for security reasons and it requires additional work to pull the plugins from a repo or blob storage for example and save it into k8s nodes filesystem. `localPath` could be another future feature using Blob storage with PV/PVC, taking advantage of safer solutions such as s3-csi-driver or FUSE CSI driver for Google.
* keep backwards compatibility with both and alert it to be depricated soon
* next major version, we remove the backward compatibility.

mloiseleur:
@jnoordsij Anything left on your side ?

traefik/templates/_podtemplate.tpl
jnoordsij: Sounds perfectly fine to me!

EXAMPLES.md
```diff
@@ -586,6 +586,62 @@ experimental:
 >[!IMPORTANT]
 > When using ``hostPath`` volumes, the plugin source code must be available on every node where Traefik pods might be scheduled.
 
+### Using Inline Plugin
+
+For testing or production deployments, embed plugin source directly in values.yaml:
```
darkweaver87: I would not mention production deployments. Please keep in mind plugins usage remains experimental (from proxy perspective) even if it's widely used.

cilasbeltrame: @darkweaver87 thanks for looking into it, just did in https://github.com/traefik/traefik-helm-chart/pull/1492/commits/eb1cfaf2ffd52755cbde3f90e10b0793c2ec281d

traefik/templates/_podtemplate.tpl
cilasbeltrame: @jnoordsij could you pls check the commit https://github.com/traefik/traefik-helm-chart/pull/1492/commits/b33f4f566350b3e51d9199aa57b914de97c07bf1

where I have enabled backwards compatibility with previous localPlugin declaration + added two new options: `inlinePlugin` (config map based) and `localPathPlugin` where we can mount plugins using pod mountpoints, it allows us to extend storage support such as PVC, CSI drivers (s3-csi-driver, FUSE), and other volume types.

traefik/tests/traefik-config_test.yaml
```diff
@@ -431,6 +431,100 @@ tests:
...
+            type:
+              hostPathPlugin:
+                path: /plugins-local/src/github.com/traefik/hostdemo
+          pvc-demo:
+            moduleName: github.com/traefik/pvcdemo
+            mountPath: /plugins-local/src/github.com/traefik/pvcdemo
+            type:
+              localPathPlugin:
+                volumeName: plugin-storage
+                subPath: plugins/pvcdemo
```
jnoordsij: I think there may have been some misunderstanding about my intended working of the new `type` key/value pair. In my mind this would be a new 'required' key for plugins, that for legacy compatibility would warn when empty and default to `hostPath`. Then based on the actual value of `type`, we'll require some further type-specific key/value pairs to be present in the rest of the plugin values. 

To illustrate here is how I would expect the values to look like:

```suggestion
          # Legacy hostPath (backward compatibility)
          legacy-host:
            moduleName: github.com/traefik/legacyhost
            mountPath: /plugins-local/src/github.com/traefik/legacyhost
            # type: hostPath # implicty assumed for backwards compatibility
            hostPath: /plugins-legacy
          # New structured plugins
          inline-demo:
            moduleName: github.com/traefik/inlinedemo
            mountPath: /plugins-local/src/github.com/traefik/inlinedemo
            type: inlinePlugin
            main.go: "package main"
          host-demo:
            moduleName: github.com/traefik/hostdemo
            mountPath: /plugins-local/src/github.com/traefik/hostdemo
            type: hostPath
            hostPath: /plugins-local/src/github.com/traefik/hostdemo
          pvc-demo:
            moduleName: github.com/traefik/pvcdemo
            mountPath: /plugins-local/src/github.com/traefik/pvcdemo
            type: localPath
            volumeName: plugin-storage
            subPath: plugins/pvcdemo
```

traefik/templates/_podtemplate.tpl
```diff
@@ -942,9 +957,24 @@
           {{- toYaml .Values.deployment.additionalVolumes | nindent 8 }}
         {{- end }}
         {{- range $localPluginName, $localPlugin := .Values.experimental.localPlugins }}
+        {{- $pluginType := include "traefik.getLocalPluginType" (dict "plugin" $localPlugin "pluginName" $localPluginName) }}
```
jnoordsij, line 945: Then here we would have something like `$pluginType := default "hostPath" plugin.type`

traefik/templates/_podtemplate.tpl
```diff
@@ -942,9 +957,24 @@
...
+        {{- if ne $pluginType "localPath" }}
```
jnoordsij, line 946: And here we would probably be able to just do if type === x, then (print body), if type === y (print body), ... else "error unknown plugin type.

jnoordsij (changes requested):
Thanks for your efforts in this @cilasbeltrame. I fear however that there's been some misunderstanding in my ideas about the plugin types. I've left some comments to clarify what my intentions with this structure were.

traefik/tests/traefik-config_test.yaml
cilasbeltrame: sure, I have pushed the changed as you suggested, my intention with my previous code  was that we would have the key values  for cm under `inlinePlugin`

```
type:
  inlinePlugin:
    main.go: |
      my code..
    .traefik.yml: |
      my code..
    go.mod: |
      my code ..

```
instead of your suggestion:
```
  inline-demo:
    moduleName: github.com/traefik/inlinedemo
    mountPath: /plugins-local/src/github.com/traefik/inlinedemo
    type: inlinePlugin
    main.go: "package main"
```

darkweaver87:
Now that you implemented @jnoordsij idea, I'm wondering if the code should be at the same level of the other keys ?
I'm thinking about:

```yaml
inline-demo:
  moduleName: github.com/traefik/inlinedemo
  mountPath: /plugins-local/src/github.com/traefik/inlinedemo
  type: inlinePlugin
  source:
    main.go: "package main"
```

This will help to distinguish the expected files from the helmchart perspective versus all the files people might want to add into their plugin. This also simplifies the code that create the ConfigMap.

Any thoughts about this ?

cilasbeltrame:
> Now that you implemented @jnoordsij idea, I'm wondering if the code should be at the same level of the other keys ? I'm thinking about:
> 
> ```yaml
> inline-demo:
>   moduleName: github.com/traefik/inlinedemo
>   mountPath: /plugins-local/src/github.com/traefik/inlinedemo
>   type: inlinePlugin
>   source:
>     main.go: "package main"
> ```
> 
> This will help to distinguish the expected files from the helmchart perspective versus all the files people might want to add into their plugin. This also simplifies the code that create the ConfigMap.
> 
> Any thoughts about this ?

I'm reverting me previous back, where I think should be better organized the cm structure, so that we can discuss based on that.

darkweaver87:
@mloiseleur , @jnoordsij, to not make @cilasbeltrame loose his time reverting and implementing again, we might agree on specifications on that topic.

On my side, I thought about a different approach in order to keep the code clean and consistent.
I wouldn't put source code directly in the plugin definition.
Instead I would introduce a - K8S way - `from`  parameter and directly reference a configMap:

configMap:
```yaml
experimental:
  localPlugins:
    inline-demo:
      moduleName: github.com/traefik/inlinedemo
      mountPath: /plugins-local/src/github.com/traefik/inlinedemo
      from:
        configMap:
          name: mysrc
          namespace: default
```

hostPath:
```yaml
experimental:
  localPlugins:
    inline-demo:
      moduleName: github.com/traefik/inlinedemo
      mountPath: /plugins-local/src/github.com/traefik/inlinedemo
      from:
        hostPath: /plugins-local/src/github.com/traefik/bar
```

From here, user will have to:
- for hostPath, put the code in the given directory as he used to do it
- for configMap, use tools like kustomize ConfigMap generator or this chart `extraObjects` to create the ConfigMap we are refering to

I also don't think we necessarily need backward compatibility and we can break here. People we need to update their values while updating to new major release. We can just raise an error if they are using the old config.

WDTY ?

mloiseleur:
@darkweaver87 I agree with you that we can introduce breaking change, if needed.

One of the need/motivation outlined by @cilasbeltrame in the head of this PR is:
> Poor GitOps integration - Plugin source code stored outside of Helm values

So I'm not sure that this proposal with `ConfigMap` handled outside of values answers his need.

On my side, I'm ok with your [previous proposal](https://github.com/traefik/traefik-helm-chart/pull/1492#issuecomment-3264930178). 

@jnoordsij Any comment ?

jnoordsij:
I think both proposals @darkweaver87 are perfectly fine; you could extend the second even with allowing `pluginName.from.source.*` as to fulfill the initial goals of the PR.

In terms of preference, I still like the idea of the explicit `type` key, so that there always is a "canonical truth" about how to interpret the values. If one adds values not matching with the provided type, they will just be ignored and it is always clear why. Moreover it is the easiest to implement in a backwards compatible manner (although I do agree that should not necessarily be a goal).

For comparison: in the second proposal, if one would add both `configMap` and `hostPath` to `from`, what should happen? Pick just the first? Use both? Add some validation to throw an error? Of course we can just pick any option and it should be fine; it just seems a bit more ambiguous to me.

As for whether or not allowing source code: as @darkweaver87 correctly pointed out, this can be achieved in a similar manner with the `extraObjects` key we have and using `configMap`, but it would be somewhat complicated. I don't have any strong opinion on whether or not to allow it, but it doesn't do any harm as far as I can see.

All in all: I do agree implementing just the suggestion from https://github.com/traefik/traefik-helm-chart/pull/1492#issuecomment-3264930178 should be preferred the way to go here.

darkweaver87:
@cilasbeltrame,thanks for your patience :-) 
We have a consensus on that proposition: https://github.com/traefik/traefik-helm-chart/pull/1492#issuecomment-3264930178 so if it makes sense to you, you can update your PR based on it.

cilasbeltrame:
> @cilasbeltrame,thanks for your patience :-) We have a consensus on that proposition: [#1492 (comment)](https://github.com/traefik/traefik-helm-chart/pull/1492#issuecomment-3264930178) so if it makes sense to you, you can update your PR based on it.

No problem - the PR grew a bit more than I expected!
Updated according to commits 38924a7 and 53c6eb9.
@darkweaver87 @jnoordsij feel free to do a final check. Thanks!

traefik-crds/README.md
```diff
@@ -0,0 +1,39 @@
+# traefik-crds
```
jnoordsij: I'm not exactly sure what happened here, but this `README` (and also the one in the `traefik` chart directory) are autogenerated and thus should not be committed; definitely not as part of this PR.

@mloiseleur should we gitignore these maybe?

traefik/templates/local-plugins-cm.yaml
jnoordsij, line 1: I'd just call it `local-plugins-configmap`, or maybe even `configmap-local-plugins`, given that I don't see the benefit of abbreviating filenames and personally when debugging a chart very much like that I can find templates quickly based on their output kind.

traefik/templates/_helpers.tpl
```diff
@@ -273,6 +277,102 @@ Hash: {{ sha1sum ($cert.Cert | b64enc) }}
...
+    {{- end -}}
+{{- end -}}
+
+{{/* 
+Get inline plugin files (new structure only)
+*/}}
+{{- define "traefik.getLocalPluginInlineFiles" -}}
+    {{- $plugin := .plugin -}}
+    {{- if eq $plugin.type "inlinePlugin" -}}
+        {{- if $plugin.source -}}
```
jnoordsij: (haven't tested) shouldn't `{{ required "localPlugin with type inlinePlugin must have a source field!" $plugin.source | toYaml }}` work here (and similar elsewhere)?

traefik/templates/local-plugins-cm.yaml
cilasbeltrame: I followed the same pattern as the existent one `provider-file-cm.yaml`, let me know if still want to change this

traefik/templates/_helpers.tpl
cilasbeltrame: awesome, good catch.
added in 1d958112b1baa9fcd22ce10e6f7457f15dd01f5a

traefik-crds/README.md
darkweaver87: @jnoordsij this README should be removed from the PR.
There is only one `README.md` at the root of this project and it's not auto-generated.

traefik/templates/local-plugins-cm.yaml
darkweaver87: Let's rename `provider-file-cm.yaml` to `configmap-providerr-file.yaml` :-)

traefik/templates/_helpers.tpl
```diff
@@ -263,6 +263,10 @@ Hash: {{ sha1sum ($cert.Cert | b64enc) }}
     {{- end -}}
 {{- end -}}
 
+{{- define "traefik.localPluginCmName" -}}
+{{ include "traefik.fullname" .context }}-local-plugin-{{ .pluginName | replace "." "-" }}
```
darkweaver87:
```suggestion
  {{ include "traefik.fullname" .context }}-local-plugin-{{ .pluginName | replace "." "-" }}
```

traefik/templates/_podtemplate.tpl
```diff
@@ -176,9 +176,23 @@
...
+            {{- if $localPathConfig.subPath }}
+            subPath: {{ $localPathConfig.subPath }}
+            {{- end }}
+          {{- else }}
           - name: {{ $localPluginName | replace "." "-" }}
             mountPath: {{ $localPlugin.mountPath | quote }}
+            {{- if eq $pluginType "inlinePlugin" }}
+            readOnly: true
+            {{- end }}
+          {{- end }}
```
darkweaver87:
```suggestion
            {{- $pluginType := include "traefik.getLocalPluginType" (dict "plugin" $localPlugin "pluginName" $localPluginName) }}
            {{- if eq $pluginType "localPath" }}
              {{- $localPathConfig := include "traefik.getLocalPluginLocalPath" (dict "plugin" $localPlugin) | fromYaml }}
          - name: {{ $localPathConfig.volumeName }}
            mountPath: {{ $localPlugin.mountPath | quote }}
              {{- if $localPathConfig.subPath }}
            subPath: {{ $localPathConfig.subPath }}
              {{- end }}
            {{- else }}
- name: {{ $localPluginName | replace "." "-" }}
mountPath: {{ $localPlugin.mountPath | quote }}
              {{- if eq $pluginType "inlinePlugin" }}
            readOnly: true
              {{- end }}
            {{- end }}
```

traefik/templates/local-plugins-cm.yaml
```diff
@@ -0,0 +1,21 @@
...
+    {{- include "traefik.labels" $ | nindent 4 }}
+data:
+  {{- $inlineFiles := include "traefik.getLocalPluginInlineFiles" (dict "plugin" $localPlugin "pluginName" $localPluginName) | fromYaml }}
+  {{- range $fileName, $fileContent := $inlineFiles }}
+  {{ $fileName }}: |
+    {{- $fileContent | nindent 4 }}
+  {{- end }}
+{{- end }}
+{{- end }}
+{{- end }}
```
darkweaver87:
```suggestion
{{- if .Values.experimental.localPlugins }}
  {{- range $localPluginName, $localPlugin := .Values.experimental.localPlugins }}
    {{- $pluginType := include "traefik.getLocalPluginType" (dict "plugin" $localPlugin "pluginName" $localPluginName) }}
    {{- if eq $pluginType "inlinePlugin" }}
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ include "traefik.localPluginCmName" (dict "context" $ "pluginName" $localPluginName) }}
  namespace: {{ template "traefik.namespace" $ }}
  labels:
    {{- include "traefik.labels" $ | nindent 4 }}
data:
  {{- $inlineFiles := include "traefik.getLocalPluginInlineFiles" (dict "plugin" $localPlugin "pluginName" $localPluginName) | fromYaml }}
    {{- range $fileName, $fileContent := $inlineFiles }}
  {{ $fileName }}: |
    {{- $fileContent | nindent 4 }}
      {{- end }}
    {{- end }}
  {{- end }}
{{- end }}
```

traefik/templates/NOTES.txt
```diff
@@ -82,6 +82,30 @@ for more info. 🚨
...
+        # Choose one of the following types:
+        type: inlinePlugin        # Recommended for small/medium plugins: secure ConfigMap-based
+        source:                  # Required for inlinePlugin
+          # your plugin files here
+        # type: hostPath         # Use with caution for security reasons
+        #   hostPath: /path/to/plugin
+        # type: localPath       # Advanced: Uses additionalVolumes, can be used with PVC, CSI drivers (s3-csi-driver, FUSE), etc.
+        #   volumeName: plugin-storage
+{{- printf "\n" -}}
+{{- end -}}
```
darkweaver87:
```suggestion
{{/* Warn about deprecated localPlugins */}}
  {{- if include "traefik.hasDeprecatedLocalPlugins" . }}
    {{- printf "\n" -}}
⚠️ DEPRECATION WARNING: You are using the deprecated legacy 'hostPath' configuration.
Please migrate to the new structured 'type.hostPathPlugin' configuration within localPlugins.
The legacy root-level hostPath configuration will be removed in the next major version.

Migration example:
  experimental:
    localPlugins:
      your-plugin:
        moduleName: github.com/example/yourplugin
        mountPath: /plugins-local/src/github.com/example/yourplugin
        # Choose one of the following types:
        type: inlinePlugin        # Recommended for small/medium plugins: secure ConfigMap-based
        source:                  # Required for inlinePlugin
          # your plugin files here
        # type: hostPath         # Use with caution for security reasons
        #   hostPath: /path/to/plugin
        # type: localPath       # Advanced: Uses additionalVolumes, can be used with PVC, CSI drivers (s3-csi-driver, FUSE), etc.
        #   volumeName: plugin-storage
    {{- printf "\n" -}}
  {{- end -}}
```

traefik-crds/README.md
mloiseleur: Yes, that makes sense 👍

EXAMPLES.md
```diff
@@ -623,20 +623,135 @@ extraObjects:
 
 To develop or test plugins without pushing them to a public registry, you can load plugin source code directly from your local filesystem.
 
->[!NOTE]
-> The ``hostPath`` must point to a directory containing the plugin source code and a valid ``go.mod`` file. The ``moduleName`` must match the module name specified in the ``go.mod`` file.
+>[!WARNING]
+> The legacy `hostPath` configuration at the `experimental.localPlugins` level is deprecated. Please use the new structured `experimental.localPlugins.<yourplugin>.type` configuration for better organization and future features.
+
+### Legacy Configuration (Backward Compatibility)
```
```suggestion
### Legacy Configuration
```
There is already a warning just below. We should keep the title as straightforward and simple as we can.

EXAMPLES.md
```diff
@@ -623,20 +623,135 @@ extraObjects:
...
+```yaml
+experimental:
+  localPlugins:
+    legacy-demo:
+      moduleName: github.com/traefik/legacydemo
+      mountPath: /plugins-local/src/github.com/traefik/legacydemo
+      hostPath: /path/to/plugin-source  # ⚠️ Deprecated - use type: hostPath instead
+```
+
+## Structured Local Plugins (Current Approach)
```
```suggestion
## Structured Local Plugins
```

EXAMPLES.md
```diff
@@ -623,20 +623,135 @@ extraObjects:
...
+
+The `localPlugins` configuration supports a structured `experimental.localPlugins.<yourplugin>.type` approach that provides better organization, security, and flexibility:
+
+### Using Inline Plugin (Recommended for small/medium plugins)
```
```suggestion
### Using Inline Plugin

> [!NOTE]  
> Can be used with small or medium plugins
```

EXAMPLES.md
```diff
@@ -623,20 +623,135 @@ extraObjects:
...
+
+          func (h *HelloWorld) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+            rw.Write([]byte("Hello, World!"))
+            h.next.ServeHTTP(rw, req)
+          }
+```
+
+> **Advantages**: Secure (no host filesystem access), portable, version controlled with Helm values, supports up to 1MB of plugin code.
+
+### Using Host Path Plugin (Use with Caution)
```
```suggestion
### Using Host Path Plugin

File: traefik/templates/_podtemplate.tpl [hunk 20]
```diff
@@ -913,9 +928,24 @@
...
+          configMap:
+            name: {{ include "traefik.localPluginCmName" (dict "context" $ "pluginName" $localPluginName) }}
+          {{- end }}
+        {{- else }}
+        {{- $localPathConfig := include "traefik.getLocalPluginLocalPath" (dict "plugin" $localPlugin) | fromYaml }}
+        {{- $volumeExists := include "traefik.volumeExistsInAdditionalVolumes" (dict "volumeName" $localPathConfig.volumeName "additionalVolumes" $.Values.deployment.additionalVolumes) }}
+        {{- if ne $volumeExists "true" }}
+          {{- fail (printf "ERROR: localPlugin %s references volume '%s' which is not found in deployment.additionalVolumes!" $localPluginName $localPathConfig.volumeName) }}
+        {{- end }}
+        {{- end }}
```
darkweaver87, line 962:
```suggestion
          {{- range $localPluginName, $localPlugin := .Values.experimental.localPlugins }}
            {{- $pluginType := include "traefik.getLocalPluginType" (dict "plugin" $localPlugin "pluginName" $localPluginName) }}
            {{- if ne $pluginType "localPath" }}
          - name: {{ $localPluginName | replace "." "-" }}
            {{- if eq $pluginType "hostPath" }}
              {{- $hostPath := include "traefik.getLocalPluginHostPath" (dict "plugin" $localPlugin) }}
            hostPath:
              path: {{ $hostPath | quote }}
            {{- else if eq $pluginType "inlinePlugin" }}
            configMap:
              name: {{ include "traefik.localPluginCmName" (dict "context" $ "pluginName" $localPluginName) }}
            {{- end }}
            {{- else }}
            {{- $localPathConfig := include "traefik.getLocalPluginLocalPath" (dict "plugin" $localPlugin) | fromYaml }}
            {{- $volumeExists := include "traefik.volumeExistsInAdditionalVolumes" (dict "volumeName" $localPathConfig.volumeName "additionalVolumes" $.Values.deployment.additionalVolumes) }}
            {{- if ne $volumeExists "true" }}
              {{- fail (printf "ERROR: localPlugin %s references volume '%s' which is not found in deployment.additionalVolumes!" $localPluginName $localPathConfig.volumeName) }}
            {{- end }}
            {{- end }}
          {{- end }}
  ```

File: traefik/templates/local-plugins-cm.yaml
Comment (jnoordsij) [reply traefik/templates/local-plugins-cm.yaml:1]: Ah fair, then leave it as is and maybe I'll follow up with changing this afterwards if I prefer the other variant ;)

File: traefik-crds/README.md (hunk 10 above)
Comment (jnoordsij) [reply traefik-crds/README.md]: I just figured out these are actually the output from `make docs`, however outputted as `README.md` rather than `VALUES.md`.

@cilasbeltrame can you remove this file (and the one in `traefik` chart) from this PR, then run `make docs` and ensure it updates `VALUES.md` rather than adding readmes?

Comment (cilasbeltrame) [reply traefik-crds/README.md]: @jnoordsij That's correct, let me review this.

Comment (cilasbeltrame) [reply traefik-crds/README.md]: @jnoordsij just did it, I've removed the files. When running `make docs` it didn't update the files `values.md`, my gues is because I haven't changed the existent schema:
`experimental.localPlugins | object | {}`

It should be good. Pls confirm.
```
cilasbeltrame in traefik-helm-chart on  feat/support-cm-local-plugins [⇣$✘?] 
❯ make docs
docker run --rm -v "/Users/cilasbeltrame/Documents/projects/opensource/traefik-helm-chart:/helm-docs" jnorwood/helm-docs:v1.14.2 -o VALUES.md
time="2025-11-06T09:41:54Z" level=info msg="Found Chart directories [traefik, traefik-crds]"
time="2025-11-06T09:41:54Z" level=info msg="Generating README Documentation for chart traefik-crds"
time="2025-11-06T09:41:54Z" level=info msg="Generating README Documentation for chart traefik"
```

**Comment by cilasbeltrame** [comment]
> I tested it with the following values:
> 
> ```yaml
> experimental:
>   localPlugins:
>     demo-inline:
>       moduleName: github.com/traefik/plugindemo
>       mountPath: /plugins-local/src/github.com/traefik/plugindemo
>       type: inlinePlugin
>       source:
>         go.mod: |
>           module github.com/traefik/plugindemo
>           
>           go 1.19
> 
>         .traefik.yml: |
>           displayName: Demo Plugin
>           type: middleware
>           iconPath: .assets/icon.png
>           
>           import: github.com/traefik/plugindemo
>           
>           summary: '[Demo] Add Request Header'
>           
>           testData:
>             Headers:
>               X-Demo: test
>               X-URL: '{{URL}}'
> 
>         demo.go: |
>           // Package plugindemo a demo plugin.
>           package plugindemo
>           
>           import (
>           "bytes"
>           "context"
>           "fmt"
>           "net/http"
>           "text/template"
>           )
>           
>           // Config the plugin configuration.
>           type Config struct {
>           Headers map[string]string `json:"headers,omitempty"`
>           }
> 
>           // CreateConfig creates the default plugin configuration.
>           func CreateConfig() *Config {
>             return &Config{
>           Headers: make(map[string]string),
>           }
>           }
> 
>           // Demo a Demo plugin.
>           type Demo struct {
>           next     http.Handler
>           headers  map[string]string
>           name     string
>           template *template.Template
>           }
> 
>           // New created a new Demo plugin.
>           func New(ctx context.Context, next http.Handler, config *Config, name string) (http.Handler, error) {
>           if len(config.Headers) == 0 {
>           return nil, fmt.Errorf("headers cannot be empty")
>           }
> 
>           return &Demo{
>           headers:  config.Headers,
>           next:     next,
>           name:     name,
>           template: template.New("demo").Delims("[[", "]]"),
>           }, nil
>           }
>   
>             func (a *Demo) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
>             for key, value := range a.headers {
>             tmpl, err := a.template.Parse(value)
>             if err != nil {
>             http.Error(rw, err.Error(), http.StatusInternalServerError)
>             return
>           }
>   
>             writer := &bytes.Buffer{}
>             
>             err = tmpl.Execute(writer, req)
>             if err != nil {
>             http.Error(rw, err.Error(), http.StatusInternalServerError)
>             return
>           }
>   
>             req.Header.Set(key, writer.String())
>           }
>   
>             a.next.ServeHTTP(rw, req)
>           }
> 
>     # Test the warning; BTW traefik should ignore the error with abortOnPluginFailure but that's not the case
>     #demo-legacy-hostpath:
>     #  moduleName: github.com/traefik/plugindemolegacyhostpath
>     #  mountPath: /plugins-local/src/github.com/traefik/plugindemolegacyhostpath
>     #  hostPath: /tmp/plugin-source  # This will trigger deprecation warning; just to test, path doesn't exist
> 
>     demo-localpath:
>       moduleName: github.com/traefik/plugindemolocalpath
>       mountPath: /plugins-local/src/github.com/traefik/plugindemolocalpath
>       type: localPath
>       volumeName: plugindemolocalpath
> 
> extraObjects:
>   - apiVersion: v1
>     kind: ConfigMap
>     metadata:
>       name: "plugindemo-localpath"
>     data:
>       go.mod: |
>         module github.com/traefik/plugindemolocalpath
>         
>         go 1.19
> 
>       .traefik.yml: |
>         displayName: Demo Plugin
>         type: middleware
>         iconPath: .assets/icon.png
>         
>         import: github.com/traefik/plugindemolocalpath
>         
>         summary: '[Demo] Add Request Header'
>         
>         testData:
>           Headers:
>             X-Demo: test
>             X-URL: '{{"{{"}}URL{{"}}"}}'
> 
>       demo.go: |
>         // Package plugindemolocalpath a demo plugin.
>         package plugindemolocalpath
>         
>         import (
>         "bytes"
>         "context"
>         "fmt"
>         "net/http"
>         "text/template"
>         )
>         
>         // Config the plugin configuration.
>         type Config struct {
>         Headers map[string]string `json:"headers,omitempty"`
>         }
>         
>         // CreateConfig creates the default plugin configuration.
>         func CreateConfig() *Config {
>           return &Config{
>         Headers: make(map[string]string),
>         }
>         }
>         
>         // Demo a Demo plugin.
>         type Demo struct {
>         next     http.Handler
>         headers  map[string]string
>         name     string
>         template *template.Template
>         }
>         
>         // New created a new Demo plugin.
>         func New(ctx context.Context, next http.Handler, config *Config, name string) (http.Handler, error) {
>         if len(config.Headers) == 0 {
>         return nil, fmt.Errorf("headers cannot be empty")
>         }
>         
>         return &Demo{
>         headers:  config.Headers,
>         next:     next,
>         name:     name,
>         template: template.New("demo").Delims("[[", "]]"),
>         }, nil
>         }
>         
>           func (a *Demo) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
>           for key, value := range a.headers {
>           tmpl, err := a.template.Parse(value)
>           if err != nil {
>           http.Error(rw, err.Error(), http.StatusInternalServerError)
>           return
>         }
>         
>           writer := &bytes.Buffer{}
>         
>           err = tmpl.Execute(writer, req)
>           if err != nil {
>           http.Error(rw, err.Error(), http.StatusInternalServerError)
>           return
>         }
>         
>           req.Header.Set(key, writer.String())
>         }
>         
>           a.next.ServeHTTP(rw, req)
>         }
>   - apiVersion: apps/v1
>     kind: Deployment
>     metadata:
>       name: whoami
>     spec:
>       replicas: 1
>       selector:
>         matchLabels:
>           app: whoami
>       template:
>         metadata:
>           labels:
>             app: whoami
>         spec:
>           containers:
>             - name: whoami
>               image: traefik/whoami:latest
>               ports:
>                 - containerPort: 80
> 
>   - apiVersion: v1
>     kind: Service
>     metadata:
>       name: whoami
>     spec:
>       ports:
>         - port: 80
>           targetPort: 80
>       selector:
>         app: whoami
> 
>   - apiVersion: traefik.io/v1alpha1
>     kind: Middleware
>     metadata:
>       name: demo-inline
>     spec:
>       plugin:
>         demo-inline:
>           headers:
>             X-Demo: "demo"
>   - apiVersion: traefik.io/v1alpha1
>     kind: Middleware
>     metadata:
>       name: demo-localpath
>     spec:
>       plugin:
>         demo-localpath:
>           headers:
>             X-Demo: "demo-localpath"
>   - apiVersion: traefik.io/v1alpha1
>     kind: IngressRoute
>     metadata:
>       name: whoami-test
>     spec:
>       entryPoints:
>         - web
>       routes:
>         - match: Host(`whoami.localhost`) || PathPrefix(`/whoami`)
>           kind: Rule
>           services:
>             - name: whoami
>               port: 80
>           middlewares:
>             - name: demo-inline
>         - match: Host(`whoami.localhost`) || PathPrefix(`/whoami2`)
>           kind: Rule
>           services:
>             - name: whoami
>               port: 80
>           middlewares:
>             - name: demo-localpath
> ingressRoute:
>   dashboard:
>     enabled: true
> 
> logs:
>   general:
>     level: INFO
>   access:
>     enabled: true
> 
> deployment:
>   additionalVolumes:
>   - name: plugindemolocalpath
>     configMap:
>       name: plugindemo-localpath
> ```
> 
> It misses a control on `mountPath` which should be required and produces:
> 
> ```
> $ helm upgrade --install traefik ./traefik -f test-pr1492-values.yaml -n traefik --create-namespace
> Error: UPGRADE FAILED: cannot patch "traefik" with kind Deployment: Deployment.apps "traefik" is invalid: spec.template.spec.containers[0].volumeMounts[3].mountPath: Required value
> ```
> 
> otherwize LGTM

@darkweaver87 I think you are having old ref on your local, using your values file I'm able to test it sucessfully from my end. Pls review and confirm.

```
❯ helm install test . -f ../dark-file.yaml  -n traefik2 --create-namespace  
NAME: test
LAST DEPLOYED: Thu Nov  6 07:55:24 2025
NAMESPACE: traefik2
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
test with docker.io/traefik:v3.5.4 has been deployed successfully on traefik2 namespace !
cilasbeltrame in traefik-helm-chart/traefik took 21.1s 
cilasbeltrame in traefik-helm-chart/traefik on  feat/support-cm-local-plugins [$?] took 21.1s 
❯ k get pods -n traefik2                                                                     
NAME                            READY   STATUS    RESTARTS   AGE
test-traefik-85c7596c99-hz5z7   1/1     Running   0          3m2s
whoami-55dd44b4d9-9qzhh         1/1     Running   0          3m2s
cilasbeltrame in traefik-helm-chart/traefik took 2.2s
❯ k logs test-traefik-85c7596c99-hz5z7 -n traefik2 | grep -i plugin
2025-11-06T10:55:36Z INF Loading plugins... plugins=["demo-inline","demo-localpath"]
2025-11-06T10:55:36Z INF Plugins loaded. plugins=["demo-inline","demo-localpath"] 
```

cilasbeltrame:
@darkweaver87 were you able to review this? 
Thanks

darkweaver87:
@cilasbeltrame sorry I missed your message. Everything is fine now. 
Did you see my indentation suggestions ?

@jnoordsij any last comment before we merge ?

cbeltra:
> @cilasbeltrame sorry I missed your message. Everything is fine now. Did you see my indentation suggestions ?
> 
> @jnoordsij any last comment before we merge ?

Feel free to apply the indentation, would be nice to have similar check on PRs eventually

darkweaver87:
oops :sweat_smile: fixing

darkweaver87:
Many thanks @cilasbeltrame for your work and your patience :-)

PR #1493, opened 2025-08-06
Only render --global.checkNewVersion if it differs from default

ChristianCiach:
### What does this PR do?

Follow up of https://github.com/traefik/traefik-helm-chart/pull/1489#issuecomment-3160112717

Parameters should only be rendered if they differ from the default configuration of Traefik. 

### Motivation

Conclusion of discussion in https://github.com/traefik/traefik-helm-chart/pull/1489#issuecomment-3160212846

### More


mloiseleur (approved):
LGTM, thanks :+1:

PR #1495, opened 2025-08-11
Remove `$root` hacks in favor of using `$`

jnoordsij:
### What does this PR do?

Replace the manual declaration of `$root` with accessing the root values through Helm's global `$` variable; see also https://helm.sh/docs/chart_template_guide/variables/#:~:text=However,%20there%20is%20one%20variable%20that%20will%20always%20point%20to%20the%20root%20context.

### Motivation

Reviewing #1492 and noticing this pattern.

### More


mloiseleur (approved):
That makes sense to me. Thanks :+1:

PR #1498, opened 2025-08-15
Add AKS monitor package support

zalbiraw:
### What does this PR do?

AKS is really annoying and they for their users to set a custom name for the moniroring.coreos.com/v1 package. This PR will allow Traefik users to set the apiVersion of the `PrometheusRule` and `ServiceMonitor` objects.

`metrics.prometheus.disableAPICheck` must be set, I decided not change the checks to not pollute the charts. 

### Motivation

http://learn.microsoft.com/en-us/azure/azure-monitor/containers/prometheus-metrics-scrape-crd

### More

### DIFF
--- a/traefik/templates/prometheusrules.yaml
+++ b/traefik/templates/prometheusrules.yaml
@@ -5,7 +5,7 @@
       {{- fail "ERROR: You have to deploy monitoring.coreos.com/v1 first" }}
     {{- end }}
   {{- end }}
-apiVersion: monitoring.coreos.com/v1
+apiVersion: {{ .Values.metrics.prometheus.prometheusRule.apiVersion | default "monitoring.coreos.com/v1" }}
 kind: PrometheusRule
 metadata:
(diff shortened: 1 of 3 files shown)


immanuelfodor:
Could @mloiseleur maybe review it?

mloiseleur:
thanks for this PR @zalbiraw 
For fixing the CI, would you please update the schema ? It requires an helm plugin ([source](https://github.com/traefik/traefik-helm-chart/blob/master/Makefile#L31)). 

Wdyt about adding a real life example of AKS in [EXAMPLES.md](https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#install-on-azure) with your test values ?

bpsoraggi (commented):
Thank you for your contribution!

Could you please add some tests? Otherwise looks good to me

mloiseleur:
@zalbiraw I added some tests and fix the examples accordingly.

PR #1505, opened 2025-08-28
Prestop hook improvements

jnoordsij:
### What does this PR do?

- Illustrate the usage of the new stable "Sleep Action for PreStop Hook"; see also https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/3960-pod-lifecycle-sleep-action
- Update example values to prefer this new easier and slightly more robust syntax
- Replace suggested and test value of 40 with 20 given default termination period is 30s and therefore 40 is probably not a great default choice

This new syntax has two main benefits, namely:
- being (slightly) easier to grasp
- being run by `kubelet` rather than inside the container, therefore not requiring `sleep` to be present as binary in the container

Therefore I think we should encourage users to use this if anything.

### Motivation

The release of Kubernetes 1.34 yesterday; see also https://kubernetes.io/blog/2025/08/27/kubernetes-v1-34-release/.

### More

- [] Yes, I updated the schema accordingly

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -91,8 +91,8 @@ deployment:
   # -- Pod lifecycle actions
   lifecycle: {}
   # preStop:
-  #   exec:
-  #     command: ["/bin/sh", "-c", "sleep 40"]
+  #   sleep:
+  #     seconds: 20
   # postStart:
   #   httpGet:
   #     path: /ping
--- a/traefik/tests/daemonset-config_test.yaml
+++ b/traefik/tests/daemonset-config_test.yaml
@@ -67,18 +67,30 @@ tests:
       - equal:
           path: spec.revisionHistoryLimit
           value: 1
-  - it: should have a preStop hook with specified values
+  - it: should have a preStop command hook with specified values
     set:
       deployment:
         kind: DaemonSet
         lifecycle:
           preStop:
             exec:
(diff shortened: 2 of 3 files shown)


traefik/values.yaml
```diff
@@ -91,8 +91,8 @@ deployment:
   # -- Pod lifecycle actions
   lifecycle: {}
   # preStop:
-  #   exec:
-  #     command: ["/bin/sh", "-c", "sleep 40"]
+  #   sleep:
+  #     seconds: 20
   # postStart:
```
mloiseleur, line 96: Wdyt of removing this and transform it into an example in EXAMPLES.md ?

mloiseleur:
PR LGTM.
@jnoordsij Do you know since which k8s version this syntax is working ?

traefik/values.yaml
jnoordsij: To be fair, though there may be a viable purpose for this in practice, I think generally using the `requestAcceptGraceTimeout` (configured for all ports) is a better way to ensure non-disrupted traffic handling with different kind of load balancers. So I think if anything we should probably document that and just leave this as a reference for specific needs (which I can't currently come up with).

jnoordsij:
> PR LGTM. @jnoordsij Do you know since which k8s version this syntax is working ?

Beta (without feature flag) 1.30+. Just added this to the PR description for reference.

PR #1517, opened 2025-09-17
Fix tracer creation warning with default security context

weijiany:
### Problem

When starting Traefik with the default [securityContext(65532:65532)](https://github.com/traefik/traefik-helm-chart/blob/9b0943844f3d6ec7104542a7072e99d467a8817a/traefik/values.yaml#L962-L965), a warning occurs preventing tracer initialization:

```
{"level":"warn","error":"building resource: error detecting resource: user: Current requires cgo or $USER set in environment","time":"2025-09-17T08:10:43Z","message":"Unable to create tracer"}
```

**Root Cause**: The OTEL `WithProcess` method in `pkg/types/tracing.go:78` calls Linux C function `getpwuid_r`, which attempts to look up UID 65532 in `/etc/passwd`. Since this UID doesn't exist, tracer creation fails.

### Solution

Set the `USER` environment variable when running with non-root UID to bypass the system user lookup.

### Steps to Reproduce

1. Create `traefik.yaml`:
```yaml
log:
  level: INFO

tracing:
  otlp:
    grpc:
      insecure: true
```

2. **Observe warning** (without USER env):
```bash
docker run -v ./traefik.yaml:/app/traefik.yaml --rm -u 65532:65532 \
  traefik:v3.5.2 --configfile /app/traefik.yaml
```

3. **Successful start** (with USER env):
```bash
docker run -v ./traefik.yaml:/app/traefik.yaml --rm -u 65532:65532 \
  -e USER=traefik traefik:v3.5.2 --configfile /app/traefik.yaml
```

### Impact

- Fixes tracer initialization for non-root containers
- Enables complete distributed tracing functionality
- Resolves warning logs that may confuse operators

### Testing

### Environment

- **Traefik Chart**: 37.1.1
- **Traefik Image**: v3.5.2

### Additional Context

Setting the non-root user in the Dockerfile rather than at pod runtime provides a more robust and maintainable solution.

1. **Consistency**: Ensures all container instances run with the same user configuration
2. **Simplicity**: Eliminates the need for runtime security context configuration
3. **Compatibility**: Avoids UID lookup issues by properly configuring `/etc/passwd` during image build
4. **Best Practice**: Follows container security recommendations to define non-root users at build time


traefik/tests/pod-config_test.yaml
```diff
@@ -897,3 +897,10 @@ tests:
       - contains:
           path: spec.template.spec.containers[0].args
           content: "--accesslog.fields.names.StartUTC=drop"
+  - it: "should have set USER: traefik in env by default"
```
bpsoraggi: LGTM, just some minor wording adjustments:

```suggestion
  - it: "should set traefik as user in env by default"
```

bpsoraggi (approved):
LGTM

Issue #1426, opened 2025-05-25
additionnalServices doesn't create new ReplicaSet/DaemonSet


TsengSR:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

traefik-35.3.0

### What version of Traefik are you using?

3.4.0

### What did you expect to happen ?

I wanted setup traefik with two gateways, for internal and external traefik. After trying for a few days, I came accross `additionalServices` and tried this attempt. 

Everything gets up, i can reach 80, 443, 8080 and 8443 ports, the correct certificates for each are passed. But the gateways kept getting the wrong ip assigned (from the first defined gateway)

I expected to see two different pods within eacah of the service, but instead they shared the same pod. And the IPs of the both gateways were of the main, external gateway. 



### What did you notice instead ?


The services get the correct IP from their metallb pools too, 192.168.0.80 (internal) and 192.168.0.90 (external). 

```bash
NAME               TYPE           CLUSTER-IP     EXTERNAL-IP    PORT(S)                                        AGE   SELECTOR
traefik            LoadBalancer   10.110.3.211   192.168.0.90   80:30128/TCP,443:32560/TCP                     20h   app.kubernetes.io/instance=traefik-traefik,app.kubernetes.io/name=traefik
traefik-internal   LoadBalancer   10.105.53.28   192.168.0.80   8081:32057/TCP,8080:30635/TCP,8443:31865/TCP   48m   app.kubernetes.io/instance=traefik-traefik,app.kubernetes.io/name=traefik
```

But the gateways, get wrong ips assigned

```bash
NAME                       CLASS              ADDRESS        PROGRAMMED   AGE
traefik-gateway            traefik            192.168.0.90   True         20h
traefik-gateway-internal   traefik-internal   192.168.0.90   True         20h
```
while fiddling around, I found out that both services reference to the same pod. So there is only one pod running for BOTH services. 

Looking closer at it, there is also only one replica set. Looking further, both services have 

```yaml                                                                                                                                                                                                                                    selector:
   app.kubernetes.io/instance: traefik-traefik
   app.kubernetes.io/name: traefik
```

as selector and only one replica which manages the one pod. Hence both services show this pod as their own. 

[...]

PR #1539, opened 2025-10-16
docs(example): multiple gateways example

moonlight8978:
### What does this PR do?

Add multiple gateways with single traefik deployment example

### Motivation

I've encountered the same problem. I searched the on the internet and the issues as well, and most of the answers were deploying multiple traefik charts into different namespace. 

- https://community.traefik.io/t/multiple-gateways-with-traefik-helm-chart/25862
- #1426

I didn't like this solution, and then I tried to investigate the chart a little bit, and end up with this solution

I'd like to share my worked solution, since there is no official example yet.
https://github.com/traefik/traefik-helm-chart/issues/1426#issuecomment-2935302634

### More

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -1456,7 +1456,7 @@ hub:
     maxRequestBodySize: 10485760 # optional, default to 1MiB
 ```
 
-## Deploy multiple Gateways with a single traefik deployment/daemonset
+## Deploy multiple Gateways with a single Traefik Deployment/DaemonSet
 
 This example exposes two Gateways (e.g., `internal` and `external`) from a single Traefik installation.
 


zikphil:
@moonlight8978 interesting! I got this working without the need to create a second gateway-class. I just put 2 gateways on the same gateway-class, can you expand on how your method is different?

moonlight8978:
Hi @zikphil, the gateway class provide parameters for the gateway, so you can make it work without multiple gateway classes, both gateways use the same configuration

zikphil:
@moonlight8978 I was having an issue where Traefik would constantly update it's own configuration and detect changes when none was done. I wonder if that could be the cause, I will try with 2 gateway class. Thank you for posting this!

EXAMPLES.md
```diff
@@ -1304,3 +1304,108 @@ hub:
...
+# New ports for external gateway
+ports:
+  web-ext:
+    port: 9080
+    expose:
+      enabled: true
+  websecure-ext:
+    port: 9443
+    expose:
+      enabled: true
```
bpsoraggi: Thank you for your contribution, examples like these are super helpful! Everything looks mostly good, I just have a few tweaks/suggestions:

```suggestion
## Deploy multiple Gateways with a single traefik deployment/daemonset

This example exposes two Gateways (e.g., `internal` and `external`) from a single Traefik installation.

```yaml
# Ports for external gateway
ports:
  web-ext:
    port: 9080
    expose:
      default: true
  websecure-ext:
    port: 9443
    expose:
      default: true
```
```

File: EXAMPLES.md [hunk 2]
```diff
@@ -1304,3 +1304,108 @@ hub:
...
+
+# This will make traefik stop updating the gateway IP, it's not important though
+# https://github.com/traefik/traefik-helm-chart/issues/1291
+service:
+  enabled: false
```
bpsoraggi: Is disabling the service really necessary? I'd keep it enabled and define secondary services with `additionalServices`, WDYT?


```suggestion

```

EXAMPLES.md
```diff
@@ -1304,3 +1304,108 @@ hub:
...
+    kind: GatewayClass
+    metadata:
+      name: traefik-external
+    spec:
+      controllerName: traefik.io/gateway-controller
+      parametersRef:
+        group: gateway.networking.k8s.io
+        kind: Gateway
+        name: traefik-gateway
+```
```
bpsoraggi:
```suggestion
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: traefik-external
  namespace: traefik
spec:
  gatewayClassName: traefik-external
  listeners:
    - name: http
      protocol: HTTP
      port: 9080
      allowedRoutes:
        namespaces:
          from: All
``` # separate code blocks (GitHub suggestions breaks, so I'll comment this here)

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: traefik-external
  namespace: traefik
spec:
  controllerName: traefik.io/gateway-controller
``` # separate code blocks

```

EXAMPLES.md
moonlight8978: thanks for your review, I disabled the service because I didn't want traefik to update  gateways' IP addresses, both the gateway received the same external IP addresses, which is the built-in service IP, it made me so confused
https://github.com/traefik/traefik-helm-chart/blob/3093fc0e09824ba606bb7ec61495d77f40aa021a/traefik/templates/_podtemplate.tpl#L531-L534
but I realized that I can disable it by setting `providers.kubernetesGateway.statusAddress.service.enabled=false`, I'll try it out

EXAMPLES.md
moonlight8978: sure, I'll separate the extra manifests into new files

moonlight8978:
hi @bpsoraggi, I've refactored the example to leverage the chart values instead of creating extra manifests. I also deleted `traefik-external` gateway class for simplicity.

I tested it on minikube and it worked well, here is the snippets which I used for testing

```bash
sudo kubectl port-forward -n default <services/traefik|services/traefik-external> 80:80
```

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whoami
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: whoami
  template:
    metadata:
      labels:
        app: whoami
    spec:
      containers:
        - name: whoami
          image: traefik/whoami:latest
          ports:
            - name: http
              containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: whoami
  namespace: default
spec:
  selector:
    app: whoami
  ports:
    - name: http
      port: 80
      targetPort: 80
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: whoami-internal
  namespace: default
spec:
  hostnames:
    - whoami-internal-127-0-0-1.sslip.io
  parentRefs:
    - name: traefik-internal
      namespace: default
      sectionName: web
  rules:
    - backendRefs:
        - name: whoami
          port: 80
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: whoami-external
  namespace: default
spec:
  hostnames:
    - whoami-external-127-0-0-1.sslip.io
  parentRefs:
    - name: traefik-external
      namespace: default
      sectionName: web-ext
  rules:
    - backendRefs:
        - name: whoami
          port: 80
```
Screenshot 2025-11-07 at 10 09 43

Screenshot 2025-11-07 at 10 09 45

PR #1540, opened 2025-10-20
feat: make dashboard toggleable

### LINKED ISSUE #1529: Dashboard is always enable (even if no ingressRoute is enabled) ?
### Welcome!

v37.1.2

v3.5.3

To disable the dashboard

It's enabled by default even if not exposed by any ingressRoute

### What are your values ?

```bash
additionalArguments:
  - "--api.dashboard=false"

```

### Additional Information

```shell
When using the chart, even if no ingressRoute is enable for the dashboard, it is activated in `traefik/templates/deployment.yaml` by the args `"--api.dashboard=true"`
The example documentation explains that the dashboard is not exposed by default and how to expose it but not how to completly disable it (or maybe it's just plain dumb from me to want to disable it when it's not exposed ?) ...
https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#access-traefik-dashboard-without-exposing-it

I tried using  using the `additionalArguments` option by  providing a custom `values.yaml` file to pass the command `--api.dashboard=false"` but that only adds the additional argument at the end of the current Args section... so I see both true and false in the describe output ...

kubectl describe deploy traefik |grep -e Args -e dashboard
    Args:
      --api.dashboard=true
      --api.dashboard=false


It "kind of works" only because Traefik apply only the last arg it receives but is there another method to properly disable the dashboard ?

Thank you
```

bpsoraggi:
### What does this PR do?

Makes it possible to disable the dashboard

### Motivation

Fixes #1529 

### More

### DIFF
--- a/traefik/templates/NOTES.txt
+++ b/traefik/templates/NOTES.txt
@@ -90,8 +90,3 @@ for more info. 🚨
     {{- printf "\nWARNING: webhook secret %s for Traefik hub is self managed and was not found in %s namespace.\n" $.Values.hub.apimanagement.admission.secretName (include "traefik.namespace" .) -}}
   {{- end -}}        
 {{- end -}}
-
-{{/* Warn about ingressRoute pointing to missing dashboard service */}}
-{{- if and (not .Values.api.dashboard) .Values.ingressRoute.dashboard.enabled }}
-  {{- printf "\nWARNING: the dashboard API is disabled, so the ingressRoute points to nowhere.\n" -}}  
-{{- end -}}
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -212,6 +212,8 @@
           {{- end }}
           {{- if .Values.api.dashboard }}
           - "--api.dashboard=true"
+          {{- else if .Values.ingressRoute.dashboard.enabled }}
+            {{- fail "ERROR: Cannot create an IngressRoute for the dashboard without enabling api.dashboard" -}}
           {{- end }}
           - "--ping=true"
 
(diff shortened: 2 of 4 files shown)


traefik/templates/NOTES.txt
```diff
@@ -90,3 +90,8 @@ for more info. 🚨
     {{- printf "\nWARNING: webhook secret %s for Traefik hub is self managed and was not found in %s namespace.\n" $.Values.hub.apimanagement.admission.secretName (include "traefik.namespace" .) -}}
   {{- end -}}        
 {{- end -}}
+
+{{/* Warn about ingressRoute pointing to missing dashboard service */}}
+{{- if and (not .Values.api.dashboard) .Values.ingressRoute.dashboard.enabled }}
+  {{- printf "\nWARNING: the dashboard API is disabled, so the ingressRoute points to nowhere.\n" -}}
```
mloiseleur: 🤔 Since it does not make sense, shouldn't we fail in this case ?

bpsoraggi: I left it as a warning because nothing is actually broken, the resources would still be valid and it'd just return a 404 for the dashboard. There are some similar warnings in `NOTES.txt` (where nothing is broken _per se_, just sort of useless), so I assumed errors were reserved for truly invalid states. But I can also see the appeal to failing it :thinking: WDYT?

mloiseleur: The other cases in NOTES.txt are misconfiguration that can still work or maybe even can make sense in some cases, where some configurations or resources are done outside of this Chart.

In this case, user enable the dashboard and cannot access it so, to me, it's a "truly invalid state".
When one enables the dashboard, one should be able to access it.

bpsoraggi: That's fair enough. I've updated the PR to fail instead

PR #1531, opened 2025-10-13
feat(logs): :memo: add missing support of OTLP logs

darkweaver87:
### What does this PR do?

This PR adds the support of OTLP parameters. It also refactors those parameters handling as most of them are common.

### Motivation

Support traefik v3.3 OTLP missing features.

### More

- [x] Yes, I updated the tests accordingly
- [x] Yes, I updated the schema accordingly
- [x] Yes, I ran `make test` and all the tests passed

PR #1545, opened 2025-10-30
fix: add missing flag arg

msaah-cleric:
### What does this PR do?

https://github.com/traefik/traefik-helm-chart/pull/1531 added support for the experimental OTLP logs exporter, but it forgot to add the flag arg that allows the binary to use this feature. this PR adds that flag.

see config in first section [here](https://doc.traefik.io/traefik/reference/install-configuration/observability/logs-and-accesslogs/#opentelemetry_1)

### Motivation

I want to use the exporter

### More

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -416,6 +416,9 @@
           - "--experimental.fastProxy.debug"
             {{- end }}
           {{- end }}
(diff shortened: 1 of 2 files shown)


darkweaver87 (approved):
LGTM.

Many thanks and good catch @msaah-cleric.

msaah-cleric:
thanks @darkweaver87, any idea when this will make it into a tagged release?

PR #1548, opened 2025-11-02
fix: add missing api flag for dashboard

remyj38:
### What does this PR do?

The `api.dashboard` command line argument require api to be enabled.

Adding option to enable the API, enabled by default.

### Motivation

Dashboard is enabled by default but needs `api=true` to be added to additional arguments.

### More


mloiseleur:
:thinking: AFAIR, enabling the dashboard enable also the API.
And I find no such requirements explained in current [upstream documentation on API](https://doc.traefik.io/traefik/reference/install-configuration/api-dashboard/#configuration-options).

@remyj38 Where does come from this requirement ? Did you observe it on a specific version ?

remyj38:
I must have misread the doc, I remembered found it somewere in the doc, but can't found it...
And it works without the api flag, so definitely, I don't know why I did that 😄

Issue #1550, opened 2025-11-03
Initial support for Traefik Proxy v3.6.0


mloiseleur:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What do you want to achieve with this Chart ?

Run Traefik Proxy v3.6.X

bpsoraggi:
After sifting through the listed changes in [Traefik v3.6.0-rc1](https://github.com/traefik/traefik/releases/tag/v3.6.0-rc1), here are the changes/features we need to provide support for in the helm chart:

- [x] Add new certificatesresolvers options - [#11977](https://github.com/traefik/traefik/pull/11977)
    #1553
- [ ] Add Knative provider - [#11448](https://github.com/traefik/traefik/pull/11448)
    #1552

And here are the changes which do not require new support (or require only CRD updates):
- [x] Allow discovering non-running Docker containers - [#10645](https://github.com/traefik/traefik/pull/10645)
- [x] Add least time load balancing strategy - [#12167](https://github.com/traefik/traefik/pull/12167)
- [x] Add TCP Healthcheck - [#11238](https://github.com/traefik/traefik/pull/11238)
- [x] Add passive health checks - [#11351](https://github.com/traefik/traefik/pull/11351)
- [x] Add highest random weight in Kubernetes CRD - [#12061](https://github.com/traefik/traefik/pull/12061)
- [x] Bump sigs.k8s.io/gateway-api to v1.4.0 - [#12140](https://github.com/traefik/traefik/pull/12140)
- [x] Support syscall - [#11939](https://github.com/traefik/traefik/pull/11939)
- [x] Allow publishing services with type ExternalName - [#12065](https://github.com/traefik/traefik/pull/12065)
- [x] Add warning when maxBodySize is not set - [#12085](https://github.com/traefik/traefik/pull/12085)
- [x] Multi-layer routing - [#12130](https://github.com/traefik/traefik/pull/12130)
- [x] Implement HTTP2 HPACK table size options - [#12050](https://github.com/traefik/traefik/pull/12050)

PR #1551, opened 2025-11-04
feat(CRDs): update for Traefik Proxy v3.6 and Gateway API v1.4.0

bpsoraggi:
### What does this PR do?

Updates CRDs for Traefik Proxy v3.6 and Gateway API v1.4.0

### Motivation

Initial support for Traefik v3.6

Part of #1550 

### More

- [ ] <s>Yes, I updated the schema accordingly</s>

### DIFF
--- a/traefik-crds/tests/crds_test.yaml
+++ b/traefik-crds/tests/crds_test.yaml
@@ -45,7 +45,7 @@ tests:
       gatewayAPI: true
     asserts:
       - hasDocuments:
-          count: 5
+          count: 6
       - isKind:
           of: CustomResourceDefinition
       - equal:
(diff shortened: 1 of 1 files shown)


darkweaver87 (approved):
LGTM

traefik-crds/crds-files/traefik/traefik.io_ingressroutes.yaml
```diff
@@ -418,4 +459,4 @@ spec:
         - spec
         type: object
     served: true
-    storage: true
```
jnoordsij, line 421: Minor nitpick: I have no idea what caused this, but it seems the empty line at the end of the file here and in a lot of other CRDs were removed, causing "incorrect" diffs. Probably we should aim to keep those around and also ensure any tooling that updates these files should always behave that way.

PR #1552, opened 2025-11-05
feat: knative provider

bpsoraggi:
### What does this PR do?

Adds support for the Knative provider experimental feature (cf. https://github.com/traefik/traefik/pull/11448)

### Motivation

Initial support for Traefik v3.6

Part of #1550 

### More

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -6,6 +6,7 @@
     "MD033": false,
     "MD036": false,
     "MD024": false,
+    "MD028": false,
     "MD041": false,
     "MD029": false,
     "MD034": false,


traefik/templates/_podtemplate.tpl
```diff
@@ -560,6 +563,17 @@
           {{- end }}
           {{- end }}
           {{- end }}
+          {{- with .Values.providers.knative }}
+           {{- if .enabled }}
+          - "--providers.knative"
+            {{- if or .namespaces (and $.Values.rbac.enabled $.Values.rbac.namespaced) }}
```
darkweaver87, line 569: You don't have associated modifications in `role.yaml` to support namespaced RBACs.

traefik/templates/rbac/clusterrole.yaml
```diff
@@ -188,6 +188,22 @@ rules:
     verbs:
       - update
   {{- end }}
+  {{- if (.Values.providers.knative).enabled }}
+  - apiGroups:
+      - networking.internal.knative.dev
```
mloiseleur, line 193: 🤔 Shouldn't we add KNative CRD, like we do for Gateway API ?

We may also choose to wait for users requests and/or GA on this feature.

@darkweaver87 @jnoordsij @bpsoraggi: Wdyt ?

traefik/values.yaml
```diff
@@ -346,6 +348,14 @@ providers:  # @schema additionalProperties: false
     # -- File content (YAML format, go template supported) (see https://doc.traefik.io/traefik/providers/file/)
     content: ""
 
+  knative:
+    # -- Enable Knative provider
+    enabled: false
+    # -- Array of namespaces to watch. If left empty, Traefik watches all namespaces. . When using `rbac.namespaced`, it will watch helm release namespace and namespaces listed in this array.
+    namespaces: []
+    # -- Allow filtering Knative Ingress objects
+    labelselector: ""
```
mloiseleur, line 357: I agree that we should not support all knative options in initial support.
Nonetheless, wdyt about adding [providers.knative.throttleduration](https://doc.traefik.io/traefik/master/reference/install-configuration/providers/kubernetes/knative/#opt-providers-knative-throttleduration) ? It looks like a parameter that will be used very frequently.

traefik/templates/rbac/clusterrole.yaml
darkweaver87: We support experimental features so why not that one (which will be "non experimental") :sweat_smile: 
BTW, I would say yes for KNative CRD.

bpsoraggi: I think it'd be good to include the CRDs, yes. Should we also include the core components? WDYT? @mloiseleur @jnoordsij

traefik/values.yaml
bpsoraggi: I didn't initially include it because we haven't included it for any of the other providers (and I do believe it can be set anyways with `additionalArguments`), but I'm not against adding it

traefik/templates/rbac/clusterrole.yaml
jnoordsij: Maintaining and updating another set of CRDs is some additional burden, but probably not too excessive, so on that end it's fine to me.

The only real risk is this chart installing quite a lot of CRDs then by default, with potentially a lot of them irrelevant for most users. Generally this does not hurt, but it feels a little unclean to me. We could opt for adding them to the CRD chart only and adding people to manually add them with the regular chart (given regular chart does not allow for updating anyways), but that also feels somewhat inconsistent.

All in all not sure what's the best way to go here, but I'm fine with either pick.

mloiseleur: >  We could opt for adding them to the CRD chart only and adding people to manually add them with the regular chart (given regular chart does not allow for updating anyways), but that also feels somewhat inconsistent.

Reading this, I think that in fact it would be consistent. That's what we do for Gateway API: the CRDs of experimental channel is an option in the CRD chart. KNative is an experimental feature, so to me it's like Gateway API.

traefik/values.yaml
mloiseleur: Let's skip it for this PR, then. It makes sense.

traefik/templates/rbac/clusterrole.yaml
bpsoraggi: I agree that it would be more consistent, so I have added the CRDs to `traefik-crds` only

traefik/templates/rbac/role.yaml
```diff
@@ -1,6 +1,7 @@
 {{- $version := include "traefik.proxyVersion" $ }}
 {{- $ingressNamespaces := concat (include "traefik.namespace" . | list) .Values.providers.kubernetesIngress.namespaces -}}
 {{- $CRDNamespaces := concat (include "traefik.namespace" . | list) .Values.providers.kubernetesCRD.namespaces -}}
+{{- $knativeNamespaces := concat (include "traefik.namespace" . | list) .Values.providers.knative.namespaces -}}
 {{- $hubNamespaces := concat (include "traefik.namespace" . | list) .Values.hub.namespaces -}}
 {{- $allNamespaces := sortAlpha (uniq (concat $ingressNamespaces $CRDNamespaces $hubNamespaces)) -}}
```
mloiseleur:
```suggestion
{{- $allNamespaces := sortAlpha (uniq (concat $ingressNamespaces $CRDNamespaces $hubNamespaces knativeNamespaces)) -}}
```

mloiseleur: @bpsoraggi You probably also needs a test here.

EXAMPLES.md
```diff
@@ -1234,6 +1234,45 @@ Once it's applied, whoami should be accessible on https://whoami.docker.localhos
 
 </details>
 
+## Use Knative Provider
+
+One can use the Knative provider (_experimental_) by setting the following _values_:
```
```suggestion
Starting with Traefik Proxy v3.6, one can use the Knative provider (_experimental_) by setting the following _values_:
```

mloiseleur (approved):

PR #1554, opened 2025-11-05
fix(doc): :books: comment grammar in values.yaml

ehpc:
### What does this PR do?

Fixes a small grammar mistake in the comment.


darkweaver87:
Hello @ehpc ,

Thanks for paying attention to this kind of detail :-)
I'm not an english native speaker so I'll let other maintainers comment this. You're right:
- than is used for comparisons involving differences (e.g., bigger than, faster than)
- as is used for equivalence or sameness (e.g., the same as)

So in this context I think "as" is correct.

darkweaver87 (approved):
LGTM

PR #1565, opened 2025-11-14
feat: allow `publishedService` to be set without service being enabled

Wieneo:
### What does this PR do?

This PR allows users to disable the `Service` provided by this Helm chart and still set the `--providers.kubernetesingress.ingressendpoint.publishedservice` arg.
This is useful if users want to bring  their own service, that is managed outside of this Helm chart.

The new logic retains the old behavior but also sets the argument if `.providers.kubernetesIngress.publishedService.enabled` is set to true and `.providers.kubernetesIngress.publishedService.pathOverride` is not empty. 

### Motivation

We have a custom ACL solution in our Kubernetes clusters, which provisions secured LoadBalancers via Kubernetes `Service` objects.
The solution expects to always managed "its own" `Service` and cannot adopt "external" `Service` objects.

### More


mloiseleur:
@Wieneo It's a very specific use case. Nonetheless, this implementation is understandable and consistent.

Code looks good to me. Would you please update the comment in values.yaml accordingly ? 

(You'll need to run `make docs` afterwards, too).

Wieneo:
@mloiseleur adapted the docs and `values.yaml` 😊

traefik/VALUES.md
```diff
@@ -406,7 +406,7 @@ Kubernetes: `>=1.22.0-0`
 | providers.kubernetesIngress.namespaces | list | `[]` | Array of namespaces to watch. If left empty, Traefik watches all namespaces. . When using `rbac.namespaced`, it will watch helm release namespace and namespaces listed in this array. |
 | providers.kubernetesIngress.nativeLBByDefault | bool | `false` | Defines whether to use Native Kubernetes load-balancing mode by default. |
 | providers.kubernetesIngress.publishedService.enabled | bool | `true` | Enable [publishedService](https://doc.traefik.io/traefik/providers/kubernetes-ingress/#publishedservice) |
-| providers.kubernetesIngress.publishedService.pathOverride | string | `""` | Override path of Kubernetes Service used to copy status from. Format: namespace/servicename. Default to Service deployed with this Chart. |
+| providers.kubernetesIngress.publishedService.pathOverride | string | `""` | Override path of Kubernetes Service used to copy status from. Format: namespace/servicename. If the service of this Chart is enabled, it defaults to that. If the service of this Chart is disabled, must be set to enable publishedService. |
```
mloiseleur:
```suggestion
| providers.kubernetesIngress.publishedService.pathOverride | string | `""` | Override path of Kubernetes Service used to copy status from. Format: namespace/servicename. Default to Service deployed with this Chart. |
```

traefik/VALUES.md
```diff
@@ -406,7 +406,7 @@ Kubernetes: `>=1.22.0-0`
 | providers.kubernetesIngress.namespaces | list | `[]` | Array of namespaces to watch. If left empty, Traefik watches all namespaces. . When using `rbac.namespaced`, it will watch helm release namespace and namespaces listed in this array. |
 | providers.kubernetesIngress.nativeLBByDefault | bool | `false` | Defines whether to use Native Kubernetes load-balancing mode by default. |
 | providers.kubernetesIngress.publishedService.enabled | bool | `true` | Enable [publishedService](https://doc.traefik.io/traefik/providers/kubernetes-ingress/#publishedservice) |
```
mloiseleur:
```suggestion
| providers.kubernetesIngress.publishedService.enabled | bool | `true` | Enable [publishedService](https://doc.traefik.io/traefik/providers/kubernetes-ingress/#publishedservice), usually with the Service provided by this Chart. It's possible to use it with an external Service using `pathOverride`. |
```

Wieneo:
@mloiseleur I adapted your changes, I think the purpose of the value is clearer now 👍

PR #1574, opened 2025-11-25
docs: Update EXAMPLES.md with REGIONAL load balancer note for Digital Ocean

kamikazechaser:
Added note about REGIONAL type load balancers. This is communication from an official ticket. It is undocumented on DO docs.

### What does this PR do?

Other types of DO LoadBalancers don't support proxy protocol. This clarifies it.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -562,6 +562,8 @@ service:
   type: LoadBalancer
   annotations:
     # This will tell DigitalOcean to enable the proxy protocol.
+    # Note that only REGIONAL type loadbalancers are supported.
+    # service.beta.kubernetes.io/do-loadbalancer-type: "REGIONAL"
     service.beta.kubernetes.io/do-loadbalancer-enable-proxy-protocol: "true"
   spec:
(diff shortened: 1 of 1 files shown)


EXAMPLES.md
```diff
@@ -562,6 +562,8 @@ service:
   type: LoadBalancer
   annotations:
     # This will tell DigitalOcean to enable the proxy protocol.
+    # Note that only REGIONAL type loadbalancers are supported.
+    # service.beta.kubernetes.io/do-loadbalancer-type: "REGIONAL"
```
mloiseleur, line 566:
```suggestion
    service.beta.kubernetes.io/do-loadbalancer-type: "REGIONAL"
```
If this is the only type supported, shouldn't this be uncommented?

kamikazechaser: There are multiple types. There is REGIONAL_NETWORK, REGIONAL and GLOBAL.

* [docs](https://docs.digitalocean.com/reference/api/digitalocean/#tag/Load-Balancers/operation/loadBalancers_create)

I left it commented to guide the user if they set the other 2.

PR #1577, opened 2025-11-26
add loadBalancerIP support for traefik service

qiaodeli111:
### What does this PR do?

add loadBalancerIP support for traefik service

### Motivation

Currently traefik service is not able to stablize the the LoadBalancer external IP as the template does not render `loadBalancerIP` defination.

It should be treated as a bug. Test cases seems not working correctly here, the `make test` always succeed.

#### Process to make stablelize the LB IP (Tested on CKS over CloudStack)

1. Create a public IP on the destination VPC (just create IP, no other configuration required)
2. Configure `service.loadBalancerIP` in values file
3. Apply the helm chart. The IP will be picked and LB configuration will be applied. 

### More

### DIFF
--- a/traefik/templates/_service.tpl
+++ b/traefik/templates/_service.tpl
@@ -31,6 +31,9 @@
   selector:
     {{- include "traefik.labelselector" .root | nindent 4 }}
   {{- if eq $type "LoadBalancer" }}
+  {{- if .service.loadBalancerIP }}
+  loadBalancerIP: {{ .service.loadBalancerIP }}
+  {{- end }}
   {{- with .service.loadBalancerSourceRanges }}
   loadBalancerSourceRanges:
   {{- toYaml . | nindent 2 }}


fopina:
I think "cannot" is not correct, I use

```yaml
service:
  spec:
    loadBalancerIP: "192.168.111.111"
    externalTrafficPolicy: Local
```

You can add anything to service spec with current values, so it's possible (and quite clean).

IMO, it'd be nice to have the value clearly in the schema, but I'm not sure it's easy to cover all possible options (as in my example using `externalTrafficPolicy: Local` too) without bloating values.yaml 🤷

mloiseleur:
That's right. 
There is even an example for this use case in _values.yaml_ ([here](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml#L989)). 

@qiaodeli111 Is there an other place where we can / should add this example ?

qiaodeli111:
> That's right. There is even an example for this use case in _values.yaml_ ([here](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml#L989)).
> 
> @qiaodeli111 Is there an other place where we can / should add this example ?

My bad... first time using traefik and didn't understand the example, thought those options are outside of `spec: {}`

Issue #1537, opened 2025-10-16
Ability to choose "" storageClassName


fopina:
https://github.com/traefik/traefik-helm-chart/blob/956a98c5c3f50693c9367166df821f2d1068b7c4/traefik/templates/pvc.yaml#L20

This check prevents me to explicitly set `storageClassName: ""`. Ommiting class name will use default (local-path) while explicitly setting to empty will allow using `nfs` for instance.

I think the check should be done on whether the key is defined rather than "truthiness".

Currently working around it by using `existingClaim` and then `extraObjects` to define the whole claim from scratch.

mloiseleur:
Thanks for your suggestion, we think it makes sense.  Unfortunately, this would not make it to our roadmap for a while as we are focused elsewhere.  If you or another community member would like to build it, let us know, and we will work with you to make sure you have all the information needed so that it can be merged.  

Don’t forget to check out the [contributor docs](https://github.com/traefik/traefik-helm-chart/blob/master/CONTRIBUTING.md) and to link the PR to this issue.

fopina:
Hey @mloiseleur found a little time to push this through and open https://github.com/traefik/traefik-helm-chart/pull/1579
Thanks!

PR #1579, opened 2025-11-26
fix(pvc): allow empty storageClassName

fopina:
### What does this PR do?

This PR aims to fix https://github.com/traefik/traefik-helm-chart/issues/1537

In PVC schema, not defining `storageClassName` means it will use _cluster default_.
However setting it to `""` (empty string) means _do not use a storage class_, disabling dynamic provisioning.
Behavior is not the same and the second case is not currently possible with that template condition.

I suggest changing `persistence.storageClass` default to `null` and handling `null` as "cluster default" (ommiting `storageClassName`) - schema updated to reflect that.
This should keep consistent behavior for everyone using the chart and _not_ re-defining `storageClass`.
Those that happen to have re-defined it as `""` in their local values will need to update though.

We can do it differently (`null` or `false` being "no storage class") but I think it would be confusing for those familiar with the PVC object schema: empty would mean ommitting and null mean empty... "why not empty mean empty?" would they ask 😄 

### Motivation

In my setup, I need to NOT usage storage classes 😄 

### More

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -314,7 +314,7 @@ Kubernetes: `>=1.22.0-0`
 | persistence.name | string | `"data"` |  |
 | persistence.path | string | `"/data"` |  |
 | persistence.size | string | `"128Mi"` |  |
-| persistence.storageClass | string | `""` |  |
+| persistence.storageClass | string | `nil` |  |
 | persistence.subPath | string | `""` | Only mount a subpath of the Volume into the pod |
 | persistence.volumeName | string | `""` |  |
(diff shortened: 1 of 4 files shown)


mloiseleur (approved):

mloiseleur:
LGTM, Thanks for this PR :+1:

fopina:
Should I rebase now or any extra review needed?

mloiseleur:
@fopina both.

Yes, you can rebase. 
Another maintainer will review it before merging.

fopina:
Sorry for the push force mess, I messed up one of the resets 😄 But rebase done now 👍

mloiseleur:
@fopina You need to rebase again (or allow us to rebase it) before we can merge it.

fopina:
@mloiseleur thanks for the review and the heads up, rebased now 👍 

@bpsoraggi obrigado pela review também! 🙇

PR #1586, opened 2025-12-04
feat(ports): enforce schema

remyj38:
### What does this PR do?

Merging all ports schema to be able to validate custom ports with schema

### Motivation

Custom ports are not validated by schema and we issued an incident caused by a typo error in custom ports.

### More

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -338,6 +338,7 @@ Kubernetes: `>=1.22.0-0`
 | ports.traefik.observability.tracing | string | `nil` | Defines whether a router attached to this EntryPoint produces traces by default. |
 | ports.traefik.port | int | `8080` |  |
 | ports.traefik.protocol | string | `"TCP"` | The port protocol (TCP/UDP) |
+| ports.web.asDefault | string | `nil` |  |
 | ports.web.expose.default | bool | `true` |  |
 | ports.web.exposedPort | int | `80` |  |
 | ports.web.forwardedHeaders.insecure | bool | `false` |  |
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -785,7 +785,9 @@ env: []
 # -- Environment variables to be passed to Traefik's binary from configMaps or secrets
 envFrom: []
 
+# @schema mergeProperties: true
 ports:
+  # @schema additionalProperties: false
   traefik:
     port: 8080
     # -- Use hostPort if set.
@@ -819,7 +821,7 @@ ports:
       traceVerbosity:  # @schema enum:[minimal, detailed, null]; type:[string, null]; default: minimal
   web:
     ## -- Enable this entrypoint as a default entrypoint. When a service doesn't explicitly set an entrypoint it will only use this entrypoint.
-    # asDefault: true
+    asDefault:  # @schema type: [boolean, null]; default: null
     port: 8000
     # hostPort: 8000
     # containerPort: 8000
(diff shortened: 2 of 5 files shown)


traefik/values.yaml
```diff
@@ -866,7 +868,7 @@ ports:
       # -- Enables tracing for this entryPoint.
       tracing:  # @schema type:[boolean, null]; default: true
       # -- Defines the tracing verbosity level for this entryPoint.
-      traceVerbosity:  # @schema enum:[minimal, detailed, null]; type:[string, null]; default: minimal
```
mloiseleur, line 869: Thanks for this PR.
Would you please detail why you are removing this enum check ?

remyj38: Because when merging properties, the schema plugin also merge enum possible values without removing duplicates.
But json schema enum list can't have duplicated values.

And only the last occurrence schema of a property is kept when merging. So schema only applied to last occurrence.

I opened an issue about it : https://github.com/losisin/helm-values-schema-json/issues/298

mloiseleur: The maintainer has published a new release with a fix for this issue:
https://github.com/losisin/helm-values-schema-json/releases/tag/v2.3.1

That was fast :). Does it mean we can keep it ?

remyj38: Too fast to be notified by GH :smile: 

With the fix, everything is good with duplicated schema for this property.

So, I updated the commit to keep it

applejag: As a fan of traefik, anything for my dear maintainers <3

traefik/values.yaml
```diff
@@ -927,6 +929,7 @@ ports:
     # It follows the provider naming convention: https://doc.traefik.io/traefik/providers/overview/#provider-namespace
     #   - namespace-name1@kubernetescrd
     #   - namespace-name2@kubernetescrd
+    # @schema type: [array, null]
     middlewares: []
```
```suggestion
    middlewares: []  # @schema type: [array, null]
```
With this change, the schema comment should not appear in VALUES.md.

traefik/VALUES.md
```diff
@@ -364,7 +365,7 @@ Kubernetes: `>=1.22.0-0`
 | ports.websecure.hostPort | string | `nil` |  |
 | ports.websecure.http3.advertisedPort | string | `nil` |  |
 | ports.websecure.http3.enabled | bool | `false` |  |
-| ports.websecure.middlewares | list | `[]` | /!\ It introduces here a link between your static configuration and your dynamic configuration /!\ It follows the provider naming convention: https://doc.traefik.io/traefik/providers/overview/#provider-namespace   - namespace-name1@kubernetescrd   - namespace-name2@kubernetescrd |
+| ports.websecure.middlewares | list | `[]` | /!\ It introduces here a link between your static configuration and your dynamic configuration /!\ It follows the provider naming convention: https://doc.traefik.io/traefik/providers/overview/#provider-namespace   - namespace-name1@kubernetescrd   - namespace-name2@kubernetescrd @schema type: [array, null] |
```
mloiseleur:
```suggestion
| ports.websecure.middlewares | list | `[]` | /!\ It introduces here a link between your static configuration and your dynamic configuration /!\ It follows the provider naming convention: https://doc.traefik.io/traefik/providers/overview/#provider-namespace   - namespace-name1@kubernetescrd   - namespace-name2@kubernetescrd |
```
See my other suggestion

traefik/values.yaml
remyj38: I missed it, it fixed !

mloiseleur:
Many thanks for this PR @remyj38 :+1: 
It will clearly help plenty of users!

PR #1589, opened 2025-12-05
feat(gateway-api): add support for defaultScope experimental feature

### LINKED ISSUE #1578: Add support for Default Gateway experimental feature
### Welcome!

Hey,
I’m in the process of migrating from Ingress to Gateway API. To make it easier for our users I want to make use of the new experimental feature regarding [default gateways](https://kubernetes.io/blog/2025/11/06/gateway-api-v1-4/#introducing-default-gateways).

However looking at the traefik helm chart it seems currently impossible to do so. I can’t add `defaultScope: All` via a variable to the helm created `Gateway`. It also seems like the crd installed by the traefik helm chart do allow me to use the experimental CRDs (but I can always just use the `traefik-crds` charts to manage them.

Ideally I would like a way to configure this `defaultScope` field in the `Gateway` object.

shubhamch71:
### What does this PR do?

Adds support for the experimental **Default Gateway** feature introduced in Gateway API v1.4 (GEP-3793).

- Add `gateway.defaultScope` to `values.yaml`
- Render `spec.defaultScope` in `templates/gateway.yaml` when set
- Update `values.schema.json`
- Documentation automatically generated via `make docs`
- Add a helm-unittest verifying the rendered defaultScope

Closes #1578

### Motivation

- Many users are migrating from Ingress to Gateway API.  The **Default Gateway** feature (new in Gateway API v1.4) allows an HTTP Route to automatically attach to a Gateway without explicitly referencing it via `parentRefs`.  This dramatically simplifies migrations and day-to-day operations.

- Before this change, the Traefik Helm chart had no way to set `spec.defaultScope`, making the feature unusable with the chart.  
- This PR enables users to fully leverage Default Gateways while continuing to deploy Traefik via Helm.

### Manual validation

```bash
helm template test ./traefik \
  --kube-version 1.29.0 \
  --set providers.kubernetesGateway.enabled=true \
  --set gateway.enabled=true \
  --set gateway.defaultScope=All \
  | yq 'select(.kind == "Gateway") | .spec.defaultScope'
  ```
  Output 
  ```
  ALL
  ```

### More

### CODE REVIEW

File: traefik/values.yaml [hunk 1]
```diff
@@ -155,6 +155,9 @@ gateway:
   annotations: {}
   # -- [Infrastructure](https://kubernetes.io/blog/2023/11/28/gateway-api-ga/#gateway-infrastructure-labels)
   infrastructure: {}
+  # -- Configure this Gateway as a [Default Gateway](https://kubernetes.io/blog/2025/11/06/gateway-api-v1-4/#introducing-default-gateways)
+  # by setting the `defaultScope` field (e.g. `All` or `Namespace`).
+  defaultScope: ""
```
mloiseleur: According to [upstream doc](https://gateway-api.sigs.k8s.io/reference/spec/#gatewaydefaultscope), it's an enum.
Wdyt of adding schema check here, in the values, to fail fast and improve UserXP ?

traefik/tests/gateway-config_test.yaml
```diff
@@ -328,3 +328,14 @@ tests:
               group: traefik.io
               kind: TraefikProxy
               name: custom-config
+  - it: should be possible to configure defaultScope on the Gateway
```
mloiseleur, line 331: 🤔 I'm not sure if this test is valid. 

Since it's using the experimental channel, isn't it required to [enable it](https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-gateway/#opt-providers-kubernetesGateway-experimentalChannel) ? 

Wdyt about adding a requirement similar to [this one](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/requirements.yaml#L16) for this new setting ?

shubhamch71:
Thanks for the review! @mloiseleur I've applied the requested changes:

- Updated `values.schema.json` to validate `defaultScope` as an enum and to allow the empty string as the default value.
- Added the experimental requirement check for `defaultScope` in `requirements.yaml`.
- Updated the unit test to enable `experimental.kubernetesGateway.enabled=true` when using `defaultScope`.

mloiseleur:
@shubhamch71 The CI is failing. It seems you need to update the schema (see Makefile for how to install the helm plugin and `make schema`)

shubhamch71:
> @shubhamch71 The CI is failing. It seems you need to update the schema (see Makefile for how to install the helm plugin and `make schema`)

Thanks a lot! @mloiseleur CI is passing now 🙂

traefik/values.yaml
darkweaver87: Per https://gateway-api.sigs.k8s.io/reference/spec/#gatewaydefaultscope:
```suggestion
  defaultScope: null # @schema enum:["All", "None", null]; type:[string, null]; default: null
```

traefik/templates/requirements.yaml
```diff
@@ -21,6 +21,9 @@
   {{- fail "ERROR: Knative is experimental. Enable it by setting experimental.knative to true" -}}
 {{- end }}
 
+{{- if and (.Values.providers.kubernetesGateway.enabled) (.Values.gateway.defaultScope) (not .Values.experimental.kubernetesGateway.enabled) }}
```
darkweaver87: Even if I understand what you want to achieve here,`.Values.experimental.kubernetesGateway.enabled` is a boolean which is used to enable **traefik** experimental features, not kubernetes ones. If we don't want to add an additional boolean flag it would be better to use `providers.kubernetesGateway.experimentalChannel`. WDYT ?

shubhamch71: yes ! I agree 

@darkweaver87 Sorry for the late reply, and happy new year! 🎉  
Thanks for the clarification 👍 I’ve switched `defaultScope` to `null`, updated the schema accordingly, and gated the feature on `providers.kubernetesGateway.experimentalChannel`. The test has also been updated to reflect this.

PR #1598, opened 2025-12-16
feat(CRDs)!: support Traefik Hub v3.19.0

traefiker:
### What does this PR do?

Support only Traefik Hub v3.19.0 or superior with current CRDs and code of this Chart.
This PR is breaking change. It requires increasing Chart version to the next major.

### Motivation

There is a breaking change on CRDs between Traefik Hub v3.18.0 and inferior and the CRDs of Traefik Hub v3.19.0+ preview versions (ea & rc).
With this PR, we introduce the support of Traefik Hub v3.19.0+ versions.

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -190,6 +190,11 @@ It requires a dict with "Version" and "Hub".
  {{- $version := .Version -}}
  {{- if .Hub -}}
    {{- $hubProxyVersion := "v3.6.7" }}
+   {{- if regexMatch "v[0-9]+.[0-9]+.[0-9]+" (default "" $version) }}
+     {{- if semverCompare "<v3.19.0-0" $version }}
+        {{- $hubProxyVersion = "v3.6.3" }}
+     {{- end -}}
+   {{- end -}}
    {{- $hubProxyVersion }}
  {{- else -}}
    {{- $version }}


mloiseleur:
This PR can be reviewed.
`no-merge` until v3.19.0 is GA.

traefik/templates/_helpers.tpl
```diff
@@ -189,20 +189,10 @@ It requires a dict with "Version" and "Hub".
 {{- define "traefik.proxyVersionFromHub" -}}
  {{- $version := .Version -}}
  {{- if .Hub -}}
-   {{- $hubProxyVersion := "v3.5" }}
+   {{- $hubProxyVersion := "v3.6.6" }}
```
darkweaver87: If we only support >= v3.19.0 then just set it to v3.6.3 and done.

traefik/tests/requirements-config_test.yaml
```diff
@@ -145,39 +155,12 @@ tests:
           enabled: true
     asserts:
       - notFailedTemplate: {}
-  - it: should fail when using offline mode on hub < 3.17.0
```
darkweaver87, line 148: We should add a test for offline mode then.

darkweaver87: That's fine we have those in hub-admission-controler_test.yaml and hub-apigateway-config_test.yaml

traefik/templates/_helpers.tpl
mloiseleur: fixed with b0e107f

traefik/templates/_helpers.tpl
```diff
@@ -189,25 +189,10 @@ It requires a dict with "Version" and "Hub".
...
-     {{- else if semverCompare "<v3.15.3-0" $version -}}
-        {{ $hubProxyVersion = "v3.3" }}
-     {{- else if or (semverCompare "=v3.16.0" $version) (semverCompare "=v3.16.1" $version) -}}
-        {{ $hubProxyVersion = "v3.3" }}
-     {{- else if semverCompare "<v3.18.0" $version -}}
-        {{ $hubProxyVersion = "v3.4" }}
-     {{- end -}}
-   {{- end -}}
-   {{ $hubProxyVersion }}
+   {{- $hubProxyVersion := "v3.6.3" }}
```
darkweaver87: It has been bumped to v3.6.7 yesterday 😅

PR #1604, opened 2025-12-18
fix(ports): 🐛 add missing `http.maxHeaderBytes` option

mloiseleur:
### What does this PR do?

Add `http.maxHeaderBytes` on entrypoints

### Motivation

It's available in Traefik Proxy since [this PR](https://github.com/traefik/traefik/pull/10995).

### More

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -909,7 +909,7 @@ ports:
         allowEncodedQuestionMark: false
         allowEncodedHash: false
       # -- Maximum size of request headers in bytes. Default: 1048576 (1 MB)
-      maxHeaderBytes:  # @schema type:[string, null]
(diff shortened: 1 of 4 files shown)


traefik/values.yaml
```diff
@@ -908,6 +908,8 @@ ports:
         allowEncodedPercent: false
         allowEncodedQuestionMark: false
         allowEncodedHash: false
+      # -- Maximum size of request headers in bytes. Default: 1048576 (1 MB)
+      maxHeaderBytes:  # @schema type:[string, null]
```
darkweaver87: Either use type `integer` or use  `traefik.convertMemToBytes` to make it easier for users.

mloiseleur: `integer` or `number` convert _2097152_ to _2.097152e+06_. 
I'm unsure this is what we want :thinking:

PR #1616, opened 2026-01-07
fix(CRDs): enforce the fact that this Chart does not support Traefik Hub v3.19.0

mloiseleur:
### What does this PR do?

Fail install when trying to deploy Traefik Hub v3.19.0 with current CRDs of this Chart.

### Motivation

There is a breaking change on CRDs between Traefik Hub v3.18.0 and inferior and the CRDs of Traefik Hub v3.19.0+ preview versions (ea & rc).
With this PR and the [crd revert](https://github.com/traefik/traefik-helm-chart/pull/1615), we fix the issue for all users with stable version.

When Traefik Hub v3.19.0 is GA, we will release a new major version of this Chart, that will only accept Traefik Hub v3.19.0+ versions.

### More

### DIFF
--- a/traefik/templates/requirements.yaml
+++ b/traefik/templates/requirements.yaml
@@ -118,7 +118,11 @@
     {{ $hubVersion = "v3.99" }}
   {{- end }}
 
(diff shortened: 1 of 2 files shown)


darkweaver87 (approved):
LGTM

PR #1618, opened 2026-01-07
fix(security): set the seccomp profile to RuntimeDefault

kkrypt0nn:
### What does this PR do?

It sets the seccomp profile to be `RuntimeDefault` rather than not being set.

### Motivation

A tool I've used to check for the security standards "Restricted" (https://kubernetes.io/docs/concepts/security/pod-security-standards/) triggered this, it was the only security change that needed attention.

### More

- [ ] ~~Yes, I updated the tests accordingly~~ I did not see any kind of tests for other security context settings other than the merging one.

### DIFF
--- a/traefik/values.schema.json
+++ b/traefik/values.schema.json
@@ -1508,6 +1508,14 @@
                 },
                 "runAsUser": {
                     "type": "integer"
+                },
+                "seccompProfile": {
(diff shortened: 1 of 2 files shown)


mloiseleur:
Thanks :+1: 
In fact, I thought it was already there :).

PR #1630, opened 2026-01-22
docs(values): avoid unbreakable lines in table output of VALUES.md

mloiseleur:
### What does this PR do?

Use `# @default --` trick (or similar) to improve table output.

### Motivation

UserXP when reading doc.


darkweaver87 (approved):
LGTM

PR #1603, opened 2025-12-18
fix(ports)!: 🐛 entrypoints `http` options

mloiseleur:
### What does this PR do?

This PR is a breaking change. 
It replaces current syntax (without `http`) with the same syntax as upstream, with `http`.

**Before**

```yaml
ports:
  web:
    redirections:
      entryPoint:
        to: websecure
        scheme: https
  websecure:
    middlewares:
    - strip@kubernetes-crd
    tls:
      enabled:
        options: "foo"
        certResolver: "certResolverFoo"
        domains:
        - main: "sub.domain.tld"
          sans:
          - "alt1.domain.tld"
          - "alt2.domain.tld"
```

**After**

```yaml
ports:
  web:
    http:
      redirections:
        entryPoint:
          to: websecure
          scheme: https
  websecure:
    http:
      middlewares:
      - strip@kubernetes-crd
      tls:
        enabled:
          options: "foo"
          certResolver: "certResolverFoo"
          domains:
          - main: "sub.domain.tld"
            sans:
            - "alt1.domain.tld"
            - "alt2.domain.tld"
```

### Motivation

1. Fix a part of https://github.com/traefik/traefik-helm-chart/issues/1602
2. Fix a part of https://github.com/traefik/traefik-helm-chart/issues/1476
3. Avoid confusion and align as much as possible chart values with Proxy config

### More

- [x] Yes, I updated the tests accordingly
- [x] Yes, I updated the schema accordingly
- [x] Yes, I ran `make test` and all the tests passed

PR #1633, opened 2026-01-23
docs(examples): use v39.0.0 syntax and improve wording

mloiseleur:
### What does this PR do?

Apply v39.0.0 syntax on EXAMPLES.md

### Motivation

https://github.com/traefik/traefik-helm-chart/pull/1603#issuecomment-3790202902

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -27,7 +27,7 @@ deployment:
 
 ## Extending DNS config
 
-In order to configure additional DNS servers for your traefik pod, you can use `dnsConfig` option:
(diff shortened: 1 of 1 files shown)


EXAMPLES.md
bpsoraggi: LGTM 🔎

While you're at it, if you could fix the "an other" to "another" on the line above it, that'd be nice:
```suggestion
It's possible to redirect all incoming requests on an entrypoint to another entrypoint.
```

mloiseleur: @bpsoraggi Wdyt of 737ec3b ?

PR #1638, opened 2026-01-27
fix(chart): allow to be used as optional subchart

schildwaechter:
### What does this PR do?

This allows the Helm chart to be used as an optional subchart.

### Motivation

Using optional subcharts is a common way to make use of different settings accross environments.
Helm [best practices recommend](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags) the use of a top level `enabled` flag for this purpose.
This is not currently allowed by the schema.

This PR introduces the `enabled` flag as optional parameter, defaulting to true.

There is no application of the value for rendering of the chart and only the value `true` is acceptable.
This is refelected in the schema.

### More

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -2,9 +2,6 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into templates
 
-# -- Allow the Helm chart to be used as optional subchart.
-enabled: true  # @schema type:boolean; const:true
-
 image:  # @schema additionalProperties: false
   # -- Traefik image host registry
   registry: docker.io
@@ -1386,3 +1383,6 @@ oci_meta:
     hub:
       image: traefik-hub
       tag: latest
+
+# -- Allow the Helm chart to be used as optional subchart.
+enabled: true  # @schema type:boolean; const:true


traefik/values.schema.json
```diff
@@ -187,6 +187,12 @@
                 }
             }
         },
+        "enabled": {
+            "type": "boolean",
+            "const": true,
+            "default": true,
+            "description": "Allow the Helm chart to be used as optional subchart."
```
mloiseleur:
```suggestion
            "const": true
```
We are using a Helm plugin to generate the schema; see
https://github.com/traefik/traefik-helm-chart/blob/master/Makefile#L31

This suggestion may pass the test green, but not tested.

schildwaechter: It tested fine locally for me.
However, this diff suggests that the helm plugin does some schema manipulation that is missing in the released schema files.
Alternatively, it may be enough to add the plugin in the process of creating the OCI artifact, though that would be less transparent.

traefik/values.yaml
```diff
@@ -2,6 +2,9 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into templates
 
+# -- Allow the Helm chart to be used as optional subchart.
+enabled: true  # @schema type:boolean; const:true
+
```
mloiseleur: Since it's only a technical parameter, wdyt of putting it at the end of the _values_ file ?

darkweaver87 (approved):
LGTM

traefik/values.yaml
schildwaechter: Done

PR #1650, opened 2026-02-05
feat(hub): add multi-cluster support

bpsoraggi:
### What does this PR do?

Adds support for the new multi-cluster feature of Traefik Hub

### Motivation

Keep the chart updated with all new features

### More

- [x] Yes, I updated the tests accordingly
- [x] Yes, I updated the schema accordingly
- [x] Yes, I ran `make test` and all the tests passed

PR #1657, opened 2026-02-06
fix(grpc): enforce mutually exclusive TLS and insecure options

### LINKED ISSUE #1634: tracing.otlp.grpc.insecure: true conflicts with default TLS values in Traefik v3.6+
### Welcome!

39.0.0

v3.6.7

When setting `tracing.otlp.grpc.insecure: true`, I expected Traefik to start successfully and send traces to my OpenTelemetry collector without TLS encryption.

Since `insecure: true` explicitly disables TLS, the chart should either:
1. Not render any `tls.*` arguments when `insecure: true` is set
2. Or set `tls` default values to `null` instead of having explicit values like `insecureSkipVerify: false`

Traefik fails to start with the following error:
```
{"level":"error","error":"command traefik error: tracing OTLP GRPC: TLS and Insecure options are mutually exclusive","time":"2026-01-24T14:05:16Z","caller":"github.com/traefik/traefik/v3/cmd/traefik/traefik.go:85","message":"Command error"}
```
The chart generates both `--tracing.otlp.grpc.insecure=true` AND TLS arguments (e.g., `--tracing.otlp.grpc.tls.insecureSkipVerify=false`) from the default values, which Traefik v3.6.7 rejects as mutually exclusive.

**Workaround:** Explicitly set `tls: null` to suppress TLS argument generation:
```yaml
tracing:
  otlp:
    grpc:
      insecure: true
      tls: null  # Required workaround
```

### What are your values ?

```yaml
tracing:
  serviceName: "traefik"
  sampleRate: 1
  otlp:
    enabled: true
    grpc:
      enabled: true
      endpoint: signoz-otel-collector.signoz-system:4317
      insecure: true
```

### Additional Information

```shell
The issue is in the `traefik.oltpCommonParams` helper in `_helpers.tpl`. It uses `traefik.yaml2CommandLineArgs` which converts all fields to CLI arguments without considering that `insecure` and `tls` are mutually exclusive:


{{- include "traefik.yaml2CommandLineArgs" (dict "path" (printf "%s.grpc" $path) "content" (omit . "enabled")) | nindent 2 }}

Since the default values define tls.insecureSkipVerify: false, this gets merged with user values and rendered as a CLI argument even when insecure: true.

Suggested fix: Modify the template to skip tls rendering when insecure: true:


{{- $content := omit . "enabled" }}
{{- if .insecure }}
  {{- $content = omit $content "tls" }}
{{- end }}
{{- include "traefik.yaml2CommandLineArgs" (dict "path" (printf "%s.grpc" $path) "content" $content) | nindent 2 }}

This likely also affects metrics.otlp.grpc and logs.otlp.grpc as they use the same helper.
```

Davee02:
### What does this PR do?

This aims to fix #1634 by adjusting the `traefik.yaml2CommandLineArgs` helper to omit the TLS options when `insecure: true` for the gRPC options.
This avoids the workaround of having to explicitly set `tls: null`.

### Motivation

I want to send traces to my OTLP collector without having to configure TLS.

### More

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -219,7 +219,7 @@ Kubernetes: `>=1.22.0-0`
 | logs.access.otlp.grpc.insecure | bool | `false` | Allows reporter to send access logs to the OpenTelemetry Collector without using a secured protocol. |
 | logs.access.otlp.grpc.tls.ca | string | `""` | The path to the certificate authority, it defaults to the system bundle. |
 | logs.access.otlp.grpc.tls.cert | string | `""` | The path to the public certificate. When using this option, setting the key option is required. |
-| logs.access.otlp.grpc.tls.insecureSkipVerify | bool | `false` | When set to true, the TLS connection accepts any certificate presented by the server regardless of the hostnames it covers. |
+| logs.access.otlp.grpc.tls.insecureSkipVerify | string | `nil` | When set to true, the TLS connection accepts any certificate presented by the server regardless of the hostnames it covers. |
 | logs.access.otlp.grpc.tls.key | string | `""` | The path to the private key. When using this option, setting the cert option is required. |
 | logs.access.otlp.http.enabled | bool | `false` | Set to true in order to send access logs to the OpenTelemetry Collector using HTTP. |
 | logs.access.otlp.http.endpoint | string | `""` | Format: <scheme>://<host>:<port><path>. Default: https://localhost:4318/v1/logs |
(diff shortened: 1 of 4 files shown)


mloiseleur:
@Davee02 This issue was more wide. We've tried a fix [here](https://github.com/traefik/traefik-helm-chart/pull/1650/changes#diff-77f1d51afca5e853adc538429e0e19cc5536c957d2796d2d2166b48e82b8b1dcR253) for an other part of the config. Basically, it skips unset boolean value. 

I imported this fix into your PR with commit d0d4f9ca0823947c91819df75008f0613e7086bc 

Wdyt ?

Davee02:
@mloiseleur LGTM! This should work for our use-case

mloiseleur (approved):

PR #1663, opened 2026-02-09
feat(deployment): support templating for podLabels

oozan:
## What does this PR do?

Adds templating support for `deployment.podLabels` in the Traefik Helm chart, matching the existing behavior of `deployment.podAnnotations`.

## Why is this needed?

`deployment.podLabels` values are currently rendered with `toYaml` only, so template expressions like `{{ template "traefik.name" . }}` are emitted literally instead of being evaluated. This PR resolves that by evaluating pod label values through `tpl`.

## Changes

- Render `deployment.podLabels` via `tpl (toYaml ...)` in `traefik/templates/_podtemplate.tpl`.
- Add unit tests for templated pod labels in both:
  - Deployment path (`traefik/tests/deployment-config_test.yaml`)
  - DaemonSet path (`traefik/tests/daemonset-config_test.yaml`)
- Update documentation strings:
  - `traefik/values.yaml`
  - `traefik/VALUES.md`

## Testing

I could not run the chart unit tests in this environment because Docker daemon access is unavailable (`docker.sock` not reachable).

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -15,7 +15,7 @@
       labels:
       {{- include "traefik.labels" . | nindent 8 -}}
       {{- with .Values.deployment.podLabels }}
-      {{- tpl (toYaml .) $ | nindent 8 }}
+        {{- tpl (toYaml .) $ | nindent 8 }}
       {{- end }}
       {{- if .Values.global.azure.enabled }}
         azure-extensions-usage-release-identifier: {{ .Release.Name }}


traefik/templates/_podtemplate.tpl
```diff
@@ -15,7 +15,7 @@
       labels:
       {{- include "traefik.labels" . | nindent 8 -}}
       {{- with .Values.deployment.podLabels }}
-      {{- toYaml . | nindent 8 }}
+      {{- tpl (toYaml .) $ | nindent 8 }}
```
mloiseleur:
```suggestion
        {{- tpl (toYaml .) $ | nindent 8 }}
```
To ease code readability, we are now indenting code in a `with` block.

mloiseleur (approved):

PR #1673, opened 2026-02-15
fix(gateway): setting port value for `websecure` listener can fail

falzm:
### What does this PR do?

This change fixes an error when trying to use the [pre-set `websecure` listener](https://artifacthub.io/packages/helm/traefik/traefik/39.0.1?modal=values&path=gateway.listeners) in the `traefik` chart values. Apparently, there is a [known issue](https://github.com/helm/helm/issues/11130) issue with `go-yaml` deserializing numbers into `float64` instead of `int64`, and [the usual workaround is to cast it manually to an `int` type](https://github.com/helm/helm/issues/13023).

I've also updated the JSON Schema to actually include support for the `gateway.listeners.websecure` object in addition to `gateway.listeners.web`, as it was previously missing – probably related to the fact that this value is optional and disabled by default, however the schema should include it anyway.

### Motivation

Currently, trying to use the pre-set `websecure` Gateway listener results in this error:

```console
$ helm template --set api.dashboard=false \
  --set providers.kubernetesGateway.enabled=true \
  --set gateway.listeners.web.namespacePolicy.from=All \
  --set gateway.listeners.websecure.namespacePolicy.from=All \
  --set gateway.listeners.websecure.port=8443 \
  --set gateway.listeners.websecure.protocol=TLS \
  --set gateway.listeners.websecure.mode=Passthrough \
  traefik ./traefik-39.0.1 # <- local template obtained via `helm pull traefik/traefik --version 39.0.1`
Error: template: traefik/templates/gateway.yaml:34:15: executing "traefik/templates/gateway.yaml" at <eq $portConfig.port $config.port>: error calling eq: incompatible types for comparison: float64 and int64
```

With this fix, the `Gateway` resource is correctly rendered:

```console
$ helm template --set api.dashboard=false \
  --set providers.kubernetesGateway.enabled=true \
  --set gateway.listeners.web.namespacePolicy.from=All \
  --set gateway.listeners.websecure.namespacePolicy.from=All \
  --set gateway.listeners.websecure.port=8443 \
  --set gateway.listeners.websecure.protocol=TLS \
  --set gateway.listeners.websecure.mode=Passthrough \
  --show-only templates/gateway.yaml \
  traefik ./traefik-helm-chart/traefik # <-- local Git repository fork including fix

---
# Source: traefik/templates/gateway.yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: traefik-gateway
  namespace: default
  labels:
    app.kubernetes.io/instance: traefik-default
    helm.sh/chart: traefik-39.0.1
spec:
  gatewayClassName: traefik
  listeners:
    - name: web
      port: 8000
      protocol: HTTP
      allowedRoutes:
        namespaces:
          from: All


    - name: websecure
      port: 8443
      protocol: TLS
      allowedRoutes:
        namespaces:
          from: All


      tls:

        mode: Passthrough
```

### More

### DIFF
--- a/traefik/values.schema.json
+++ b/traefik/values.schema.json
@@ -354,38 +354,6 @@
                                     "type": "integer"
                                 },
                                 "protocol": {
-                                    "description": "Specify expected protocol on this listener. See [ProtocolType](https://gateway-api.sigs.k8s.io/reference/spec/#protocoltype)",
-                                    "type": "string"
-                                }
-                            }
-                        },
-                        "websecure": {
-                            "type": "object",
-                            "properties": {
-                                "certificateRefs": {
-                                    "description": "Reference to a Secret containing a certificate for `TLS` or `HTTPS` protocols (ignored for `mode:Passthrough`). See [SecretObjectReference](https://gateway-api.sigs.k8s.io/reference/spec/#secretobjectreference)"
-                                },
-                                "hostname": {
-                                    "description": "Optional hostname. See [Hostname](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.Hostname)",
-                                    "type": "string"
-                                },
-                                "mode": {
-                                    "description": "TLS behavior for the TLS session initiated by the client. See [TLSModeType](https://gateway-api.sigs.k8s.io/reference/spec/#tlsmodetype).",
-                                    "type": "string"
-                                },
-                                "namespacePolicy": {
-                                    "description": "Routes are restricted to namespace of the gateway [by default](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.FromNamespaces",
-                                    "type": [
-                                        "object",
-                                        "null"
-                                    ]
-                                },
-                                "port": {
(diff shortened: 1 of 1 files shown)


mloiseleur:
@falzm The schema is [generated from the values](https://github.com/traefik/traefik-helm-chart/blob/master/Makefile#L31), it cannot be changed like that. 

Do you think you can remove the schema part of this PR? It should make the CI go green.

falzm:
> Do you think you can remove the schema part of this PR? It should make the CI go green.

@mloiseleur done ✅

mloiseleur:
Thank :+1:  And, in order to be sure it won't happen again, do you think you can update an existing test or add a new test about this ? This test should fail without this change.

falzm:
@mloiseleur looking at the tests to update, I've actually realized that there are several already existing ones that cover this use case (e.g. [this one](https://github.com/traefik/traefik-helm-chart/blob/v39.0.1/traefik/tests/gateway-config_test.yaml#L70-L104)), and they... pass, even without my fix 🙃 I'm not sure to understand how.

PR #1679, opened 2026-02-23
fix(traefik-hub): set token using a file

komalsukhani:
### What does this PR do?

Added following new configuration options in traefik chart to allow setting of token using file:

- tokenMountPath:
Token k8s secret will be mounted as volume on this path. Set to `/etc/secrets` by default.

### Motivation

Security. Allows to avoid exposing the token in an env variable of the container.

### More


traefik/values.yaml
```diff
@@ -1205,6 +1205,11 @@ hub:
   # -- Name of `Secret` with key 'token' set to a valid license token.
   # It enables API Gateway.
   token: ""
+  # -- Mount path for token secret.
+  tokenMountPath: "/etc/secrets"
+  # -- Name of file containing a valid license token.
+  # It enables API Gateway.
+  tokenFilePath: ""
```
mloiseleur: 🤔 Wouldn't it be simpler to just do the switch from env-based to file-based without those parameters?
Maybe I'm missing something here, but I have a hard time seeing when a user would need to set those two values?

komalsukhani: Fixed in e647b54

traefik/templates/_podtemplate.tpl
```diff
@@ -20,7 +20,7 @@
       {{- if .Values.global.azure.enabled }}
         azure-extensions-usage-release-identifier: {{ .Release.Name }}
       {{- end }}
-      {{- if and .Values.hub.token .Values.hub.apimanagement.enabled .Values.hub.apimanagement.admission.restartOnCertificateChange }}
+      {{- if  and .Values.hub.token .Values.hub.apimanagement.enabled .Values.hub.apimanagement.admission.restartOnCertificateChange }}
```
darkweaver87:
```suggestion
      {{- if and .Values.hub.token .Values.hub.apimanagement.enabled .Values.hub.apimanagement.admission.restartOnCertificateChange }}
```

traefik/values.schema.json
```diff
@@ -831,6 +831,14 @@
                     "description": "Name of `Secret` with key 'token' set to a valid license token. It enables API Gateway.",
                     "type": "string"
                 },
+                "tokenFilePath": {
```
darkweaver87: Need to re-run `make schema`

traefik/templates/_helpers.tpl
```diff
@@ -455,3 +455,10 @@ Check if using old localPlugin hostPath structure (for deprecation warning)
    {{- end }}
   {{- end }}
 {{- end }}
+
+{{/*
+Define hub token mount path
+*/}}
+{{- define "traefik.hubTokenFilePath" }}
+{{- printf "%s/%s" (.Values.hub.tokenMountPath | trimSuffix "/") "token" -}}
+{{- end -}}
```
darkweaver87:
```suggestion
{{- end -}}

```

traefik/tests/hub-license-config_test.yaml
```diff
@@ -57,15 +60,50 @@ tests:
...
           path: spec.template.spec.containers[0].args
-          content: "--hub.token=$(HUB_TOKEN)"
+          content: "--hub.tokenFilePath=/custom/token"
         template: deployment.yaml
+      - containsDocument:
+          kind: Secret
+          apiVersion: v1
+          name: traefik-hub-license
+          namespace: NAMESPACE
+        template: hub-license.yaml
```
darkweaver87:
```suggestion
        template: hub-license.yaml

```

darkweaver87 (commented):
Please link the related traefik-hub PR.

mloiseleur:
@komalsukhani Since this option will only be available in v3.19.3+, wdyt of updating the requirements here?
https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/requirements.yaml#L121

komalsukhani:
> https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/requirements.yaml#L121

@mloiseleur Makes sense. Updated.

traefik/templates/_podtemplate.tpl
```diff
@@ -820,7 +825,7 @@
           {{- end }}
           {{- with .Values.hub }}
            {{- if .token }}
-          - "--hub.token=$(HUB_TOKEN)"
+          - "--hub.tokenFilePath={{ include "traefik.hubTokenFilePath" $ }}"
```
darkweaver87, line 803: Here you break the compatibility with v3.19.2. This should be version dependent unfortunately.

komalsukhani: Since I am setting a value in default values.yaml, even if the user doesn't set that value, it will be used.

komalsukhani: I think I should not set default value in values.yaml and set it in the _helper.tpl instead.
And to set it only tag is v3.19.3+. Will update the PR accordingly.

darkweaver87 (approved):
LGTM

PR #1683, opened 2026-02-25
fix: remove `rbac.secretResourceNames`

### LINKED ISSUE #1653: Setting `secretResourceNames` breaks Traefik
### Welcome!

v39.0.0

v3.6.7

I expect that setting the value `rbac.secretResourceNames` limits the secrets that traefik has access to and that traefik still works.

Setting `rbac.secretResourceNames` limits the secrets in the ClusterRole of traefik (so that works as expected), but setting this value **breaks the functionality of traefik**.

Traefik logs something like:
```
E0205 07:37:56.525125       1 reflector.go:205] "Failed to watch" err="failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:traefik:traefik" cannot list resource "secrets" in API group "" at the cluster scope" logger="UnhandledError" reflector="k8s.io/client-go@v0.34.3/tools/cache/reflector.go:290" type="*v1.Secret"
```

This is a bug of the traefik controller itself which is documented in https://github.com/traefik/traefik/issues/7097.

So limiting resourceNames for secrets breaks traefik. So I think allowing the limitation in this chart is wrong because we now have values in the chart that can break traefik.

### What are your values ?

```
...
rbac:
  secretResourceNames:
    - ingress-tls
...
```

### Additional Information

@alestrix already stated this **2024** in https://github.com/traefik/traefik-helm-chart/pull/1009#issuecomment-2180307397: `I understand that traefik is "broken" in a way that it needs that permission, but being able to define a list of resourceNames gives the user a false feeling of security.`

mloiseleur:
### What does this PR do?

It removes `rbac.secretResourceNames`.

### Motivation

It's not supported by Traefik.
Fixes #1653

### More


darkweaver87 (approved):
LGTM

talha5389:
This change breaks my setup.

I've been running namespaced Traefik with rbac.secretResourceNames scoped to only the TLS secret within the namespace. This has worked correctly for at least 6 months — certs were being renewed and Traefik could access them without issue.

The issue this fixes (#1653) specifically mentions cluster-scoped deployment, where restricting secretResourceNames breaks Traefik's informer because Kubernetes RBAC doesn't support resourceNames filtering on a cluster-scope list. 

However, with namespaced: true, no ClusterRole is created at all — only a namespace-scoped Role. In that case, resourceNames restrictions on list work correctly because the scope is already bounded to the namespace.

Could you confirm whether removing secretResourceNames is intentional for namespaced deployments as well, or whether this was only meant to address the cluster-scoped case? If the former, it would be worth documenting that namespaced users need to manage secret RBAC manually outside of Helm or role out fix for that

Issue #1680, opened 2026-02-23
externalTrafficPolicy: Local seems to be ignored on helm install


parkerfath:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

39.0.2

### What version of Traefik are you using?

3.6.8

### What did you expect to happen ?

In values.yaml, set `service.externalTrafficPolicy: Local` - expected Service to be created with `externalTrafficPolicy: Local`

I want to be able to use IPAllowList middleware and see client IP addresses in backend access logs.

### What did you notice instead ?

Service was created with `externalTrafficPolicy: Cluster`

### What are your values ?

1. In values.yaml:
    1. deployment.enabled: true
    2. deployment.kind: DaemonSet
    3. ingressClass.enabled: true
    4. ingressClass.isDefaultClass: false
    5. gateway.enabled: false
    6. gatewayClass.enabled: false
    7. providers.kubernetesCRD.enabled: true
    8. providers.kubernetesIngress.enabled: true
    9. providers.kubernetesGateway.enabled: false
    10. providers.kubernetesIngressNginx.enabled: false
    11. service.externalTrafficPolicy: Local

### Additional Information

Other aspects of the values.yaml seem to work, e.g. I can change the log level and change Deployment to DaemonSet. externalTrafficPolicy is the only one I've seen that doesn't work.

I can manually update the Service using `kubectl` later, but ideally the helm chart should create it correctly at the start.

parkerfath:
OK actually maybe this was a misunderstanding on my part. I had just uncommented the externalTrafficPolicy line and added Local, which was this:
```
  spec: {}
  externalTrafficPolicy: Local
```

But I realized that to enable it, you need to remove the braces and indent the externalTrafficPolicy line, like this:
```
  spec:
    externalTrafficPolicy: Local
```

On one hand, I'd consider maybe a PR to make that clearer, but maybe this is just me being unaware of how helm/YAML typically work? Happy to close this as no issue depending on what others think.

mloiseleur:
If you had been confused, others may face it too.
So yes, feel free to open a PR to show us how you think we can avoid this.

parkerfath:
[...]

PR #1686, opened 2026-02-25
fix(service)!: align syntax with upstream

parkerfath:
### What does this PR do?

Changes the indent and order of a few values.yaml items under service.spec to clarify which items go inside the "spec" field and which items are directly under "service."

**Before**

```yaml
service:
  type: LoadBalancer
  loadBalancerSourceRanges:
    - 192.168.0.1/32
  externalIPs:
    - 1.2.3.4
  ipFamilyPolicy: SingleStack
  ipFamilies:
    - IPv6
  spec:
    externalTrafficPolicy: Cluster
```

**After**

```yaml
service:
  spec:
    type: LoadBalancer
    externalIPs:
      - 1.2.3.4
    externalTrafficPolicy: Cluster
    ipFamilyPolicy: SingleStack
    ipFamilies:
      - IPv6
    loadBalancerSourceRanges:
      - 192.168.0.1/32
```

### Motivation

See https://github.com/traefik/traefik-helm-chart/issues/1680 - I tried uncommenting and setting externalTrafficPolicy but it was silently ignored.

### More

(No test or schema change needed, but I did run the tests.)

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -455,13 +455,14 @@ Kubernetes: `>=1.25.0-0`
 | readinessProbe.timeoutSeconds | int | `2` | The number of seconds to wait for a probe response before considering it as failed. |
 | resources | object | `{}` | [Resources](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) for `traefik` container. |
 | securityContext | object | See _values.yaml_ | [SecurityContext](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context-1) |
+| service.additionalServices | object | `{}` | Can be used to create multiple Service. See EXAMPLES.md for more details. |
 | service.annotations | object | `{}` | Additional annotations applied to both TCP and UDP services (e.g. for cloud provider specific config) |
 | service.annotationsTCP | object | `{}` | Additional annotations for TCP service only |
 | service.annotationsUDP | object | `{}` | Additional annotations for UDP service only |
 | service.enabled | bool | `true` |  |
 | service.labels | object | `{}` | Additional service labels (e.g. for filtering Service by custom labels) |
 | service.single | bool | `true` |  |
-| service.spec | object | `{"type":"LoadBalancer"}` | Cannot contain selector or ports entries. |
+| service.spec | object | `{"type":"LoadBalancer"}` | Additional entries here will be added to the Service [spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#servicespec-v1-core). Cannot contain selector or ports entries. |
 | serviceAccount | object | `{"name":""}` | The service account the pods will use to interact with the Kubernetes API |
 | serviceAccountAnnotations | object | `{}` | Additional serviceAccount annotations (e.g. for oidc authentication) |
 | startupProbe | object | `{}` | Define [Startup Probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes) |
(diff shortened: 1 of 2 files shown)


mloiseleur (commented):
🤔 In fact, we should not have also `type`, `loadBalancerSourceRanges` and `externalIPs`. And there is the same issue for the _additionalServices_ section.

Like all the others, there are all subparameters of `Service`'s spec. For consistency and simplicity, they all should be set in `service.spec` (except selector and ports entries, that should be computed from related values)

This is not only in _values_, we should also remove nearly all the code (and just keep what the user set on `.spec`) here:

https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/_service.tpl#L22-L50

And also, to avoid values.yaml growing too much, we've moved the examples to EXAMPLES.md

I know it's breaking but sometimes it can be better in the long term. 
This brings two questions:
1. Wdyt of this approach? 
2. Do you think you can try to implement it? Or do you prefer that we update your PR?

parkerfath:
> 🤔 In fact, we should not have also `type`, `loadBalancerSourceRanges` and `externalIPs`. And there is the same issue for the _additionalServices_ section.
> 
> Like all the others, there are all subparameters of `Service`'s spec. For consistency and simplicity, they all should be set in `service.spec` (except selector and ports entries, that should be computed from related values)
> 
> This is not only in _values_, we should also remove nearly all the code (and just keep what the user set on `.spec`) here:
> 
> https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/_service.tpl#L22-L50
> 
> And also, to avoid values.yaml growing too much, we've moved the examples to EXAMPLES.md
> 
> I know it's breaking but sometimes it can be better in the long term. This brings two questions:
> 
> 1. Wdyt of this approach?
> 2. Do you think you can try to implement it? Or do you prefer that we update your PR?

Hmm I *think* I see what you're getting at, but not sure I have enough context to implement. Would it be something like this? https://github.com/parkerfath/traefik-helm-chart/commit/3a6136191934d6f737b863c423745425d72901c5

mloiseleur:
Yes.
There is also [this helper](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/_service.tpl#L22-L50) to rework. It should only set _.service.spec_. And there are also tests to update accordingly to this new behavior.

parkerfath:
I am interested, but might not be able to carve out the time to do it quickly (not to mention this is my first contribution to a helm chart). So, if it's high priority you can feel free to take it off my hands, otherwise, I'll see what I can do with my free cycles.

mloiseleur:
@parkerfath I updated the PR with the simplified syntax. It comes with less code to maintain and should avoid any future confusion this. Since it's a breaking change, I updated PR description with a before / after block.

traefik/templates/_service.tpl
```diff
@@ -15,38 +15,16 @@
...
 
 {{- define "traefik.service-spec" -}}
-  {{- $type := default "LoadBalancer" .service.type }}
-  type: {{ $type }}
-  {{- with .service.loadBalancerClass }}
-  loadBalancerClass: {{ . }}
-  {{- end}}
   {{- with .service.spec }}
-  {{- toYaml . | nindent 2 }}
+    {{- toYaml . | nindent 2 }}
```
jnoordsij, line 24: Maybe we could use `(merge . (dict "type" "LoadBalancer"))` here (if the top-of-my-head Helm skills suffice) to at least retain the previous default? Otherwise requiring people to provide this now is also not too much of a breaking change I'd say.

mloiseleur: All the other moves like this, we failed gracefully (with schema) and did not support it, so the user can update it.

:thinking: 

With current schema verification, it will fail anyway before this `merge` is executed.
So I think we can stay like that. It's only a parent change.

PR #1698, opened 2026-03-04
feat(ci): self-hosted renovate

darkweaver87:
### Motivation

Switch Renovate to GitHub Action on this repo.


.github/workflows/add-to-project.yaml
```diff
@@ -0,0 +1,18 @@
+name: Add backend PRs to TraefikLabs project
```
mloiseleur: This file should not be on this repository.

.github/renovate.json5
```diff
@@ -0,0 +1,48 @@
+{
```
mloiseleur, line 1: There is already a renovate.json

.github/workflows/renovate.yaml
```diff
@@ -0,0 +1,73 @@
...
+          tar -xzf cache-download/$CACHE_ARCHIVE -C $CACHE_DIR
+
+          sudo chown -R 12021:0 /tmp/renovate/
+          ls -R $CACHE_DIR
+
+      - name: Self-hosted Renovate
+        uses: renovatebot/github-action@d65ef9e20512193cc070238b49c3873a361cd50c # v46.1.1
+        with:
+          configurationFile: .github/renovate.json5
+          renovate-version: full
```
viceice, line 56: do not use the full version, it's huge, use `latest` instead which is much faster

viceice: better to set a fixed version or at least the major. so you don't get silent major updates

darkweaver87: Thanks for looking at this, we have some debates internally about this. We will take you suggestions into account :+1:

viceice: you can trust me, I'm a renovate maintainer. 🤗 

your enable manager don't need any extra binary tools, so using the full image is waste of time downloading the fat image.

darkweaver87: ok many thanks :-)

mloiseleur: thanks @viceice. We have switched for using the current default major version of pinned GitHub actions with [this PR](https://github.com/traefik/traefik-helm-chart/pull/1705).

It's clearly faster: https://github.com/traefik/traefik-helm-chart/pull/1705

PR #1715, opened 2026-03-11
fix: :art: align chart and traefik case on CLI for yaml2CommandLineArgs

darkweaver87:
### What does this PR do?

This PR updates the helper `yaml2CommandLineArgs` to generate CLI args which matches traefik ones (even if traefik handles case conversion).

### Motivation

Align chart and traefik CLI.

### More


darkweaver87:
The PR was needed to rework nginx provider implementation without having the fix the test. As discussed @mloiseleur, there is no proper convention for traefik CLI arguments (all camelCase, upper case, lower case, ...).

PR #1716, opened 2026-03-12
docs(gateway): improve wording on namespaces

mloiseleur:
### What does this PR do?

Improve wording on namespaces option of kubernetesGateway provider.

### Motivation

Accuracy.


traefik/values.yaml
```diff
@@ -349,7 +349,7 @@ providers:
     # -- Toggles support for the Experimental Channel resources (Gateway API release channels documentation).
     # This option currently enables support for TCPRoute and TLSRoute.
     experimentalChannel: false
-    # -- Array of namespaces to watch. If left empty, Traefik watches all namespaces. . When using `rbac.namespaced`, it will watch helm release namespace and namespaces listed in this array.
+    # -- Array of namespaces to watch. If left empty, Traefik watches all namespaces.
```
darkweaver87:
```suggestion
    # -- Array of namespaces to watch. If left empty, Traefik watches all namespaces. kubernetesGateway provider requires ClusterRole and as a consequence `rbac.namespaced` is not supported.
```

zalbiraw:
That is great. from a UX perspective. I believe that we should add a check as well. The problem is because some of the helm resources will install partially and the install will fail if the gateway provider is turned on and the rbac.namespaced is set to true

PR #1718, opened 2026-03-12
refactor(chart)!: support only Proxy v3.6+ & Kubernetes v25+

mloiseleur:
### What does this PR do?

Drop support of unmaintained versions of Proxy & update minimum requirements  on Kubernetes to v25

### Motivation

1. Align with upstream
2. Ease maintenance
3. Code simplification

### More

### DIFF
--- a/traefik/tests/azure-config_test.yaml
+++ b/traefik/tests/azure-config_test.yaml
@@ -4,17 +4,17 @@ templates:
 tests:
   - it: should not set global.azure arg by default
     chart:
-      appVersion: v3.3.0
+      appVersion: v3.6.10
     asserts:
       - equal:
           path: spec.template.spec.containers[0].image
-          value: docker.io/traefik:v3.3.0
+          value: docker.io/traefik:v3.6.10
       - notExists:
           path: spec.template.metadata.labels.azure-extensions-usage-release-identifier
 
   - it: should use global.azure with proxy when enabled
     set:
-      versionOverride: v3.3.5
+      versionOverride: v3.7.0
       global:
         azure:
           enabled: true
--- a/traefik/tests/deployment-config_test.yaml
+++ b/traefik/tests/deployment-config_test.yaml
@@ -316,7 +316,7 @@ tests:
   - it: should have abortOnPluginFailure, when enabled on traefik >=3.3.0
     set:
       image:
-        tag: v3.3.0
+        tag: v3.6.10
       experimental:
         abortOnPluginFailure: true
     asserts:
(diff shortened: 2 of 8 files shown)


traefik/templates/rbac/clusterrole.yaml
```diff
@@ -11,33 +11,10 @@ metadata:
...
-      - list
-      - watch
-    {{- end }}
-  {{- else }}
-  - apiGroups:
-      - ""
-    resources:
-      {{- if and .Values.providers.kubernetesCRD.enabled (semverCompare ">=v3.4.0-0" $version) }}
       - configmaps
-      {{- end }}
```
darkweaver87, line 40: The `if .Values.providers.kubernetesCRD.enabled` is still needed.

traefik/templates/rbac/clusterrole.yaml
```diff
@@ -105,9 +69,6 @@ rules:
   - apiGroups:
       - ""
     resources:
-    {{- if semverCompare ">=v3.7.0-0" $version }}
-      - configmaps
-    {{- end }}
```
darkweaver87, line 110: Why removing that one ?

traefik/templates/rbac/role.yaml
```diff
@@ -16,23 +16,10 @@ metadata:
...
-      - get
-      - list
-      - watch
-  {{- else }}
-  - apiGroups:
-      - ""
-    resources:
-      {{- if or (and $.Values.providers.kubernetesCRD.enabled (semverCompare ">=v3.4.0-0" $version)) (semverCompare ">=v3.7.0-0" $version) }}
       - configmaps
-      {{- end }}
```
darkweaver87, line 35: Should be updated, not removed:
```
      {{- if or ($.Values.providers.kubernetesCRD.enabled) (semverCompare ">=v3.7.0-0" $version) }}
      - configmaps
      {{- end }}
```

traefik/tests/hpa-config_test.yaml
```diff
@@ -5,21 +5,6 @@ capabilities:
   majorVersion: 1
   minorVersion: 23
```
darkweaver87:
```suggestion
  minorVersion: 25
```

traefik/templates/rbac/clusterrole.yaml
mloiseleur: Moved to the top. It's now required for all providers.

traefik/templates/rbac/role.yaml
mloiseleur: Moved to the top. It's now required for all providers.

darkweaver87: As discussed with you, that"s not true. It's a shortcut you took. People using `knative`, `file`, `kubernetesGateway` or `kubernetesIngressNginx` (on v3.6), but I won't block on this.

PR #1720, opened 2026-03-12
feat(providers): :rocket: update NGINX with v3.7 options

darkweaver87:
### What does this PR do?

This PR adds provider NGINX new options for upcoming v3.7.

### Motivation

Support v3.7.

### More

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -377,7 +377,7 @@ providers:
 
   # @schema additionalProperties: false
   kubernetesIngressNGINX:
-    # -- Enable Kubernetes Ingress NGINX provider (experimental)
+    # -- Enable Kubernetes Ingress NGINX provider
     enabled: false
     # -- Ingress Class Controller value this controller satisfies
     controllerClass: "k8s.io/ingress-nginx"
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -438,7 +438,7 @@ Kubernetes: `>=1.25.0-0`
 | providers.kubernetesIngressNGINX.customHTTPErrors | list | `[]` | Defines which HTTP status codes should result in calling the default backend to return an error page |
 | providers.kubernetesIngressNGINX.defaultBackendService | string | `""` | Service used to serve HTTP requests not matching any known server name (catch-all). Takes the form 'namespace/name' |
 | providers.kubernetesIngressNGINX.disableSvcExternalName | bool | `false` | Disable support for Services of type ExternalName |
-| providers.kubernetesIngressNGINX.enabled | bool | `false` | Enable Kubernetes Ingress NGINX provider (experimental) |
+| providers.kubernetesIngressNGINX.enabled | bool | `false` | Enable Kubernetes Ingress NGINX provider |
 | providers.kubernetesIngressNGINX.endpoint | string | `""` | Kubernetes server endpoint (required for external cluster client) |
 | providers.kubernetesIngressNGINX.globalAllowedResponseHeaders | list | `[]` | List of allowed response headers inside the custom headers annotations |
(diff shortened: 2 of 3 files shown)


traefik/tests/requirements-config_test.yaml
```diff
@@ -103,6 +103,27 @@ tests:
...
+        kubernetesIngressNGINX:
+          enabled: true
+          httpEntryPoint: "web"
+    asserts:
+      - failedTemplate:
+          errorMessage: "ERROR: some Kubernetes Ingress NGINX provider options require traefik >= v3.7.0."
+  - it: should not fail when using kubernetesIngressNGINX v3.7 options on traefik >= 3.7.0
+    set:
+      image:
+        tag: v3.7.0
```
```suggestion
        tag: v3.7.0-ea.1
```
Nit pick. Just to be sure it works with current early access

traefik/tests/provider-kubernetesingressnginx_test.yaml
```diff
@@ -164,3 +164,144 @@ tests:
...
+          proxyBufferSize: 8192
+          proxyBuffersNumber: 4
+          proxyConnectTimeout: 60
+          proxyReadTimeout: 60
+          proxySendTimeout: 60
+    asserts:
+      - template: deployment.yaml
+        contains:
+          path: spec.template.spec.containers[0].args
+          content: "--providers.kubernetesingressnginx.proxyRequestBuffering=true"
```
mloiseleur, line 185: 🤔 Is this really expected ? The provider is not enabled in this test.

traefik/tests/provider-kubernetesingressnginx_test.yaml
```diff
@@ -164,3 +164,144 @@ tests:
...
+      - template: deployment.yaml
+        contains:
+          path: spec.template.spec.containers[0].args
+          content: "--providers.kubernetesingressnginx.clientBodyBufferSize=16384"
+      - template: deployment.yaml
+        contains:
+          path: spec.template.spec.containers[0].args
+          content: "--providers.kubernetesingressnginx.proxyBodySize=1.048576e+06"
```
mloiseleur: This option fails on traefik v3.7.0-ea.1 with this log:

```json
{"level":"error","error":"command traefik error: failed to decode configuration from flags: strconv.ParseInt: parsing \"1.048576e+06\": invalid syntax","time":"2026-03-13T08:07:40Z","message":"Command error"}
```

I guess we need to set it as a string, not an int ?

traefik/tests/provider-kubernetesingressnginx_test.yaml
darkweaver87: It's enabled for all tests of this file https://github.com/traefik/traefik-helm-chart/blob/master/traefik/tests/provider-kubernetesingressnginx_test.yaml#L7.

traefik/tests/provider-kubernetesingressnginx_test.yaml
darkweaver87: Fixed, thanks !

traefik/values.yaml
```diff
@@ -409,6 +409,45 @@ providers:
     endpoint: ""
     # -- Kubernetes bearer token (not needed for in-cluster client). It accepts either a token value or a file path to the token
     token: ""
+    ## -- ConfigMap options available within the NGINX Ingress Controller ConfigMap
```
```suggestion
```
Nitpick. It won't appear in the VALUES.md anyway.

traefik/values.yaml
```diff
@@ -409,6 +409,45 @@ providers:
     endpoint: ""
     # -- Kubernetes bearer token (not needed for in-cluster client). It accepts either a token value or a file path to the token
     token: ""
+    ## ConfigMap options available within the NGINX Ingress Controller ConfigMap
```
darkweaver87:
```suggestion
```

PR #1726, opened 2026-03-13
feat(api): add dashboardName option

darkweaver87:
### What does this PR do?

This PR adds support for the `api.dashboardName` option introduced in Traefik v3.7, allowing a custom name for the dashboard.

### Motivation

Support v3.7.

### More

### DIFF
--- a/traefik/values.yaml
+++ b/traefik/values.yaml
@@ -208,7 +208,6 @@ api:  # @schema additionalProperties: false
   # -- Configure API basePath
   basePath: ""  # @schema type:[string, null]; default: "/"
 
-
 # -- Only dashboard & healthcheck IngressRoute are supported.
 # It's recommended to create workloads CR outside of this Chart.
 # @default -- See _values.yaml_


traefik/values.yaml
```diff
@@ -205,6 +205,8 @@ api:  # @schema additionalProperties: false
   debug:  # @schema type:[boolean, null]
   # -- Configure API basePath
   basePath: ""  # @schema type:[string, null]; default: "/"
+  # -- Custom name for the dashboard (v3.7+).
+  dashboardName: ""  # @schema type:[string, null]
```
mloiseleur: Nitpick: wdyt of moving this option just below `dashboard: true` ?

traefik/values.yaml
```diff
@@ -199,13 +199,16 @@ gatewayClass:  # @schema additionalProperties: false
...
   # -- Enable the insecure API (HTTP)
   insecure:  # @schema type:[boolean, null]
   # -- Enable the debug API
   debug:  # @schema type:[boolean, null]
   # -- Configure API basePath
   basePath: ""  # @schema type:[string, null]; default: "/"
 
+
```
```suggestion
```

PR #1734, opened 2026-03-16
feat(ports): add allowACMEByPass to web port values

jnoordsij:
### What does this PR do?

Due to changes introduced in https://github.com/traefik/traefik/pull/12740, which have landed in Proxy v3.6.10, the `allowACMEByPass` can now be used for redirected entrypoints to disable redirects for ACME paths. The most common case for this will be the `web` endpoint, redirecting to `websecure`, now being able to filter out these routes. To make it easier for users to spot this feature, we add the value explicitly with documentation reference.

### Motivation

Seeing https://github.com/traefik/traefik/pull/12740

### More

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -156,7 +156,7 @@ It's possible to redirect all incoming requests on an entrypoint to another entr
 ```yaml
 ports:
   web:
-    # -- If you are handling ACME challenges with another service (e.g. cert-manager),
+    # -- If you are handling ACME HTTP challenges with another service (e.g. cert-manager),
     #    you may enable this option to prevent redirects on ACME challenge routes for this port.
     # allowACMEByPass: true
     http:


mloiseleur:
We have an [example](https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#redirect-permanently-traffic-from-http-to-https) about this :thinking:.

@jnoordsij Do you think we should update it?

jnoordsij:
Ah yes! Probably more common still will be people just leveraging Traefik itself for ACME challenges, but pointing it out more clearly would be useful.

EXAMPLES.md
```diff
@@ -156,6 +156,9 @@ It's possible to redirect all incoming requests on an entrypoint to another entr
 ```yaml
 ports:
   web:
+    # -- If you are handling ACME challenges with another service (e.g. cert-manager),
```
```suggestion
    # -- If you are handling ACME HTTP challenges with another service (e.g. cert-manager),
```

EXAMPLES.md
```diff
@@ -156,6 +156,9 @@ It's possible to redirect all incoming requests on an entrypoint to another entr
 ```yaml
 ports:
   web:
+    # -- If you are handling ACME HTTP challenges with another service (e.g. cert-manager),
```
darkweaver87, line 159: Do you think we can and would be relevant to illustrate / PoC this with cert-manager so that a user can test it "out of the box" ?

jnoordsij: Given this is a simple/generic example I don't really think it would fit here, or really be in the scope of this PR. It may definitely be worth it to explore adding some additional example, but there are quite a lot of choices in setup (e.g. what provider to use, how to setup domains, etc.), so I'm not entirely sure what would be the most reasonable "out of the box" example to provide.

PR #1744, opened 2026-03-17
feat(deployment): add hostUsers support

### LINKED ISSUE #1597: Allow setting `hostUsers` on pods
### Welcome!

To adhere to restrictive cluster security policies, setting `hostUsers` on the pod spec should be possible.

k8s docs here: [User Namespaces Docs](https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/). This reduces blast radius of a pod compromise, and many orgs have added a security policy to require that `hostUsers` be set to `false`.

In most cases this can be set to `false` with negligible impact on the workload. That being said, setting it to `false` by default would risk breaking some installations that mount in certain storage types that lack idmap support, or are using less common network configurations like setting `hostNetwork`.

### Proposal

Add templating for setting `hostUsers`  in the [pod template](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/_podtemplate.tpl), and leave it unset by default, then document it.

This is a very simple feature, if there is maintainer support I can throw together a quick PR.

DrFaust92:
## Summary

- Add `deployment.hostUsers` value to allow setting [`hostUsers`](https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/) on the Traefik pod spec
- Setting `hostUsers: false` enables Kubernetes user namespace isolation, reducing the blast radius of a pod compromise
- Left **unset by default** to avoid breaking installations using storage types without idmap support or `hostNetwork` mode
- Schema allows `boolean | null` so the field is truly optional
- Added unit test coverage: set to `true`, set to `false`, and unset

Closes #1597

## Example

```yaml
deployment:
  hostUsers: false  # enables user namespace isolation
```

## Test plan

- [x] `helm unittest traefik -f tests/pod-config_test.yaml` passes (82/82 tests)

### DIFF
--- a/traefik/templates/_podtemplate.tpl
+++ b/traefik/templates/_podtemplate.tpl
@@ -33,6 +33,9 @@
       automountServiceAccountToken: true
       terminationGracePeriodSeconds: {{ default 60 .Values.deployment.terminationGracePeriodSeconds }}
       hostNetwork: {{ .Values.hostNetwork }}
+      {{- if not (kindIs "invalid" .Values.deployment.hostUsers) }}
+      hostUsers: {{ .Values.deployment.hostUsers }}
+      {{- end }}
       {{- with .Values.deployment.dnsPolicy }}
       dnsPolicy: {{ . }}
       {{- end }}
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -55,6 +55,7 @@ Kubernetes: `>=1.25.0-0`
 | deployment.healthchecksPort | string | `nil` |  |
 | deployment.healthchecksScheme | string | `nil` |  |
 | deployment.hostAliases | list | `[]` | Custom [host aliases](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/) |
+| deployment.hostUsers | string | unset (inherits cluster default) | Whether to use the host user namespace. Setting this to false enables user namespaces, which can improve security by isolating the pod's users from the host. See https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/ |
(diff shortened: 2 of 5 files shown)


traefik/templates/_podtemplate.tpl
```diff
@@ -33,6 +33,9 @@
       automountServiceAccountToken: true
       terminationGracePeriodSeconds: {{ default 60 .Values.deployment.terminationGracePeriodSeconds }}
       hostNetwork: {{ .Values.hostNetwork }}
+      {{- if not (kindIs "invalid" .Values.deployment.hostUsers) }}
```
mloiseleur, line 36:
```suggestion
      {{- with .Values.deployment.hostUsers }}
```

mloiseleur (dismissed):

darkweaver87 (approved):
LGTM once tests will pass :-)

mloiseleur:
@DrFaust92 Sorry, my suggestion was wrong. The first implementation was better. 
Would you please revert the last commit?

mloiseleur (approved):

especially-relative:
should there be a test for invalid configs combinations?

https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/#limitations

DrFaust92:
> should there be a test for invalid configs combinations?
> 
> https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/#limitations

I can add it but I think other tests are that through or should we check any possible edge case

especially-relative:
Depends on if running traefik with host networking is commonplace, that is the only issue that would be likely to trip someone up.

PR #1761, opened 2026-03-25
docs: clarify DaemonSet updateStrategy to prevent rollout deadlock

Ahmed-Sameh-MM:
This merge request is part of the open issue #1753.

### What does this PR do?

Adds a clarification in the DaemonSet installation example regarding the required update strategy.

It explains why the default `updateStrategy` is not suitable for a `DaemonSet` and provides a minimal working configuration to prevent rollout deadlocks.

### Motivation

When using Traefik as a `DaemonSet`, the default update strategy (`maxUnavailable: 0`) can lead to a deadlock during updates. Since a `DaemonSet` cannot schedule more than one pod per node, no pods can be terminated and the rollout cannot proceed.

This change helps users avoid this issue by documenting a working configuration.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -9,6 +9,20 @@ deployment:
   kind: DaemonSet
 ```
 
+The update strategy default value in `traefik/values.yaml` needs to be changed, in order to accommodate for a `DaemonSet`.
+
+Whenever the `DaemonSet` gets updated with a new configuration using `maxUnavailable: 0`, this prevents any pod owned by the DaemonSet from being terminated, and a `DaemonSet` cannot schedule more than one pod per node, which results in a rollout deadlock.
+
+Instead, this minimal working update-strategy configuration can be used.
+
+```yaml
+updateStrategy:
+  type: RollingUpdate
+  rollingUpdate:
+    maxUnavailable: 1
+    maxSurge: 0
+```
+
 ## Configure Traefik Pod parameters
 
 ### Extending /etc/hosts records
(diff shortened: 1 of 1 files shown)


EXAMPLES.md
```diff
@@ -9,6 +9,20 @@ deployment:
...
+Instead, this minimal working update-strategy configuration can be used.
+
+```yaml
+updateStrategy:
+  type: RollingUpdate
+  rollingUpdate:
+    maxUnavailable: 1
+    maxSurge: 0
+```
+
```
mloiseleur, line 17:
````suggestion
  kind: DaemonSet
# The update strategy needs to be changed accordingly
# See https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/ for details
updateStrategy:
  rollingUpdate:
    maxUnavailable: 1
```

````

Wdyt of this shorter version?

Ahmed-Sameh-MM: I like it, it's straight to the point.

mloiseleur (approved):

Ahmed-Sameh-MM:
I think that's everything for this pull request. Will it be merged now?

darkweaver87 (approved):
LGTM

EXAMPLES.md
Ahmed-Sameh-MM: > Wdyt of this shorter version?

One small thing I noticed here is that "maxSurge: 0" needs to be added, so that the update strategy works without errors, as maxUnavilable can't be set with "maxSurge: 1", which is the default value in the helm chart.

Issue #1669, opened 2026-02-13
traefik Helm chart should not install Gateway API CRDs


sathieu:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

39.0.0

### What version of Traefik are you using?

v3.6.7

### What did you expect to happen ?

Gateway API CRDs are not owned by traefik. Please allow installing them by other methods (i.e. allow to disable installation of those CRDs).

### What did you notice instead ?

Gateway API CRDs are always installed

### What are your values ?

not related.

### Additional Information

```shell

```

9numbernine9:
+1. We're already using the Gateway API with Cilium and have the Gateway API CRDs installed by other means; now, the Traefik Helm chart wants to conflict with this previous installation. 😢

I'm sure the Traefik Helm chart is trying to be helpful here, but there should be some kind of 'escape valve' if the chart is going to install these CRDs. (IMHO, the chart shouldn't try to install CRDs that aren't in the `traefik.io` or `hub.traefik.io` API groups in the first place.)

janlonsetteig:
FYI: This is also being discussed here: https://github.com/traefik/traefik-helm-chart/issues/1561#issuecomment-3728593868

mloiseleur:
>  but there should be some kind of 'escape valve' if the chart is going to install these CRDs.

The current feature set of Helm does not give us any way on this. We explored the option of providing a dedicated chart on CRD, as suggested in [upstream documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions), but it's a dead end: Helm can [fail with more than 1 MB of content](https://github.com/traefik/traefik-helm-chart/issues/1561), and Gateway API CRDs are already bigger than that.

A specific version of Traefik is tailored towards a specific version of Gateway API. Sometimes, it can _somehow_ work; sometimes, there are breaking changes between versions of Gateway API.

There are Pros & Cons in excluding Gateway API CRDs from this Chart.

| **Pros**                                                                   | **Cons**                                                                                                                                       |
|----------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------|
[...]

PR #1782, opened 2026-03-31
feat(CRDs): add traefik hub ContentItem

jspdown:
### What does this PR do?

This PR brings the ContentItem CRD for Traefik Hub API management. This new resource is used to describe an additional page on an APIPortal.

### Motivation

Allow users to add additional documentation on their portals

### More

### DIFF
--- a/traefik-crds/tests/crds_test.yaml
+++ b/traefik-crds/tests/crds_test.yaml
@@ -26,7 +26,7 @@ tests:
       hub: true
     asserts:
       - hasDocuments:
-          count: 14
+          count: 15
       - isKind:
           of: CustomResourceDefinition
       - equal:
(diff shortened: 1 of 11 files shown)


traefik-crds/tests/crds_test.yaml
mloiseleur:
```suggestion
          count: 15
```

mloiseleur:
Overall PR LGTM.
We may stop CRD chart after the v40 GA release, when Proxy v3.7 will be GA (see [here](https://github.com/traefik/traefik-helm-chart/issues/1669#issuecomment-4096810602) for details).
We'll probably need to rework this PR after this release and before merging it.

darkweaver87 (commented):
`ContentItem` crds  were introduced in hub-crds `v1.29.0`, currently only `v3.20.ea.8+` use it and `v1.29.0` also include `UpLinks` CRD starting `v1.28.0`. Consequently, this PR needs at least to be based on `v39.1` branch not master.

darkweaver87 (approved):
LGTM

PR #1722, opened 2026-03-13
chore(crds): :wastebasket: remove deprecated traefik-crds chart

darkweaver87:
### What does this PR do?

This PR removes deprecated chart.

### Motivation

Ease maintenance.

### More

- [x] Yes, I updated the tests accordingly
- [x] Yes, I updated the schema accordingly
- [x] Yes, I ran `make test` and all the tests passed

PR #1784, opened 2026-04-02
feat: :sparkles: add `hub-manager` chart

bpsoraggi:
### What does this PR do?

Adds new `hub-manager` chart

### Motivation

To include the new chart in the repository

### More


hub-manager/templates/_helpers.tpl
```diff
@@ -0,0 +1,101 @@
...
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "hhub-manager.chart" -}}
```
darkweaver87: There is many of this kind of typo in this file:
```suggestion
{{- define "hub-manager.chart" -}}
```

hub-manager/templates/_helpers.tpl
```diff
@@ -0,0 +1,101 @@
...
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create the chart image name.
+*/}}
+{{- define "hhub-manager.image-name" -}}
+{{- printf "%s/%s:%s" .Values.image.registry .Values.image.repository (.Values.image.tag | default .Chart.AppVersion) }}
```
darkweaver87: Should this even be configurable ?

bpsoraggi: I left it like this so it follows the same pattern as the Traefik chart and in case tags can change, but I agree we can remove it if it'll always be static

darkweaver87: We'll probably re-introduce it if needed ;-)

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+  # -- Base URL of the workspace service
+  workspaceURL: ""
+  # -- URL of the trace service
+  traceURL: ""
+
+hydra:
+  # -- URL of the Hydra admin API
+  url: ""
+  # -- URL of the Hydra issuer
+  issuerURL: ""
```
mloiseleur: AFAIR, it should not be required in this version.
```suggestion
```

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+    protocol: TCP
+  
+# -- overrides the app.kubernetes.io/name label
+nameOverride: ""
+
+# -- Overrides the resource name for templates (i.e deployment, service, etc..)
+fullnameOverride: ""
+
+# -- This field overrides the default Release Namespace for Helm
+namespaceOverride: ""
```
mloiseleur: They came from legacy migration in Traefik Helm chart.
It should not be needed for this chart.
```suggestion
```

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+    maxUnavailable: 0  # @schema type:[integer, string, null]
+    maxSurge: 1        # @schema type:[integer, string, null]
+
+# This section builds out the service account more information can be found here: https://kubernetes.io/docs/concepts/security/service-accounts/
+serviceAccount:
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+  # -- Whether to automatically mount a ServiceAccount's API credentials
+  automount: true
```
mloiseleur:
```suggestion
  automountServiceAccountToken: false
```

hub-manager/templates/rbac/serviceaccount.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if not .Values.serviceAccount.name -}}
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: {{ include "hub-manager.serviceAccountName" . }}
```
mloiseleur:
```suggestion
  name: {{ include "hub-manager.fullname" . }}
```
Wdyt about using same name on all resources, and so avoid this parameter?
Like for the `RoleBinding`, for instance

hub-manager/templates/rbac/clusterrole.yaml
```diff
@@ -0,0 +1,70 @@
+---
+kind: ClusterRole
```
mloiseleur: Normally, only `Role` and `RoleBinding` should be needed

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
+# Default values for hub-manager
+# This is a YAML-formatted file
+# Declare variables to be passed into templates
+
+image:  # @schema additionalProperties: false
+  # -- Hub-manager image host registry
+  registry: gcr.io
```
mloiseleur:
```suggestion
  registry: ghcr.io
```
GCR is a private registry in this use case.

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+  # -- Hub-manager image repository
+  repository: traefiklabs/hub-manager
```
mloiseleur:
```suggestion
  repository: traefik/hub-manager
```

hub-manager/templates/NOTES.txt
```diff
@@ -0,0 +1 @@
+
```
mloiseleur: Wdyt about adding something similar to [those 4th first lines](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/NOTES.txt#L1C1-L4C20) in this NOTES.TXT ?

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+  # -- Name of Secret with key 'jwt-cert' set to a valid JWT Cert
+  cert: ""
+
+services:
+  # -- Base URL of the offer service
+  offerURL: ""
+  # -- Base URL of the workspace service
+  workspaceURL: ""
+  # -- URL of the trace service
+  traceURL: ""
```
mloiseleur:
```suggestion
```
AFAIK, should be removed

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+  # -- Name of Secret with key 'postgres-encryption-key' set to a valid encryption key
+  encryptionKey: ""
+
+jwt:
+  # -- Name of Secret with key 'jwt-iss' set to a valid JWT Issuer
+  iss: ""
+  # -- Name of Secret with key 'jwt-sub' set to a valid JWT Sub
+  sub: ""
+  # -- Name of Secret with key 'jwt-cert' set to a valid JWT Cert
+  cert: ""
```
mloiseleur:
```suggestion
```
should be removed

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+  maxReplicas:  # @schema type:[integer, null]; minimum:0
+  # -- metrics contains the specifications for which to use to calculate the desired replica count (the maximum replica count across all metrics will be used)
+  metrics: []
+  # -- behavior configures the scaling behavior of the target in both Up and Down directions (scaleUp and scaleDown fields respectively)
+  behavior: {}
+  # -- scaleTargetRef points to the target resource to scale, and is used for the pods for which metrics should be collected, as well as to actually change the replica count
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: "{{ template \"hub-manager.fullname\" . }}"
```
mloiseleur:
```suggestion
autoscaling:  # @schema additionalProperties: false
  # -- Create HorizontalPodAutoscaler object
  enabled: false
  # -- Spec of the HPA
  spec:
    scaleTargetRef:
      apiVersion: apps/v1
      kind: Deployment
      name: "{{ template \"hub-manager.fullname\" . }}"
    minReplicas: 1
    maxReplicas:  2
    metrics:
      - resource:
          name: cpu
          target:
            averageUtilization: 70
            type: Utilization
        type: Resource      
```
Wdyt of simplifying the template and put a default, common example in the values? (Instead of mapping one by one).

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+  imagePullSecrets: []
+  # -- Pod lifecycle actions
+  lifecycle: {}
+
+updateStrategy:  # @schema additionalProperties: false
+  # -- Customize updateStrategy of Deployment
+  type: RollingUpdate
+  rollingUpdate:
+    maxUnavailable: 0  # @schema type:[integer, string, null]
+    maxSurge: 1        # @schema type:[integer, string, null]
```
mloiseleur: Shouldn't this be a sub-part of `deployment:` ?

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+    port: 8080
+  # -- The number of consecutive failures allowed before considering the probe as failed
+  failureThreshold: 2
+  # -- The number of seconds to wait before starting the first probe
+  initialDelaySeconds: 5
+  # -- The number of seconds to wait between consecutive probes
+  periodSeconds: 5
+
+# -- Define [Startup Probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes)
+startupProbe: {}
```
mloiseleur: Shouldn't this be a subpart of `deployment:`? 🤔

hub-manager/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+  username: ""
+  # -- Name of Secret with key 'tracing-password' set to a valid password
+  password: ""
+
+ports:
+  # @schema additionalProperties: false
+  http:
+    port: 8080
+    # -- The port protocol (TCP/UDP)
+    protocol: TCP
```
mloiseleur: Why let the user configure deployment ports ? Shouldn't this be done by default and let user choose on `Service` level, where it's exposed?

hub-manager/templates/_podtemplate.tpl
```diff
@@ -0,0 +1,160 @@
...
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
+        {{- with .Values.deployment.startupProbe}}
+        startupProbe:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
+        {{- with .Values.deployment.lifecycle }}
+        lifecycle:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
```
mloiseleur:
```suggestion
        {{- with .Values.resources }}
        resources: {{- toYaml . | nindent 10 }}
        {{- end }}
        {{- with .Values.deployment.readinessProbe }}
        readinessProbe: {{- toYaml . | nindent 10 }}
        {{- end }}
        {{- with .Values.deployment.livenessProbe }}
        livenessProbe: {{- toYaml . | nindent 10 }}
        {{- end }}
        {{- with .Values.deployment.startupProbe}}
        startupProbe: {{- toYaml . | nindent 10 }}
        {{- end }}
        {{- with .Values.deployment.lifecycle }}
        lifecycle: {{- toYaml . | nindent 10 }}
        {{- end }}
```

hub-manager/values.yaml
bpsoraggi: I don't mind. I did it like this initially so the new chart would be as similar as possible to the Traefik chart. Would it be better to change, or to keep the same pattern as the Traefik chart? 🤔

hub-manager/values.yaml
bpsoraggi: I agree. I only kept it like this to follow the pattern from the Traefik chart

hub-manager/values.yaml
bpsoraggi: Same as `updateStrategy`

hub-manager/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: hub-manager
+description: A Helm chart for Kubernetes
+type: application
+version: 0.1.0
+appVersion: "1.16.0"
```
mloiseleur:
```suggestion
version: 1.0.0
appVersion: "1.0.0"
```

hub-manager/values.yaml
mloiseleur: In traefik chart, it handles also Daemonset deployment.
We'll continue to refactor and improve Traefik chart, so maybe we can avoid some inconsistency pattern when we find ones ?

hub-manager/values.yaml
```diff
@@ -0,0 +1,147 @@
...
+
+# -- Address to listen on
+address: ":8080"
+
+# -- Name of Secret with key 'token' set to a valid license token
+token: ""
+
+logs:
+  # -- Set logs format
+  format: "json" # @schema enum:["common", "json", null]; type:[string, null]; default: "common"
```
mloiseleur:
```suggestion
```
Doesn't exist in hub-manager. It's only json, at least FTM.

hub-manager/values.yaml
```diff
@@ -0,0 +1,147 @@
...
+  # -- Alternative logging levels are TRACE, DEBUG, INFO, WARN, ERROR, FATAL, and PANIC
+  level: "INFO"  # @schema enum:[TRACE,DEBUG,INFO,WARN,ERROR,FATAL,PANIC]; default: "INFO"
+
+deployment:
+  # -- Number of pods of the deployment
+  replicas: 1
+  # -- Number of old history to retain to allow rollback (If not set, default Kubernetes value is set to 10)
+  revisionHistoryLimit: 10  # @schema type:[integer, null];minimum:0
+  # -- Amount of time (in seconds) before Kubernetes will send the SIGKILL signal if shut down does not happen
+  terminationGracePeriodSeconds: 60
```
mloiseleur:
```suggestion
```
🤔 I'm wondering if we are not going in too much detail here. It's a simple app with DB.
Wdyt of simplifying this chart and go with the default for those kind of very specific parameters?
```suggestion
```

hub-manager/values.yaml
```diff
@@ -0,0 +1,147 @@
...
+    port: 8080
+  # -- The number of consecutive failures allowed before considering the probe as failed
+  failureThreshold: 2
+  # -- The number of seconds to wait before starting the first probe
+  initialDelaySeconds: 5
+  # -- The number of seconds to wait between consecutive probes
+  periodSeconds: 5
+
+# -- Define [Startup Probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes)
+startupProbe: {}
```
mloiseleur:
```suggestion
```
On this kind of software, it's a different context than Traefik.
So we may take a different path. 
We can simplify both users and maintainers life by providing reliable hard-coded defaults on probes.
Wdyt ?

hub-manager/values.yaml
mloiseleur: terminationGracePeriod can already be set with `lifecycle` value.

hub-manager/templates/rbac/role.yaml
youkoulayley, line 1: We don't need specific permissions for `hub-manager`, it doesn't look for resources on the cluster.
I think, role, rolebinding and serviceaccounts are not needed.

hub-manager/templates/_helpers.tpl
```diff
@@ -0,0 +1,85 @@
...
+and additionally imposes a limit of 15 characters on the length of the name.
+See also https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services
+and https://www.rfc-editor.org/rfc/rfc6335#section-5.1.
+*/}}
+{{- define "hub-manager.portname" -}}
+{{- $portName := . -}}
+{{- $portName = $portName | lower -}}
+{{- $portName = $portName | trimPrefix "-" | trunc 15 | trimSuffix "-" -}}
+{{- print $portName -}}
+{{- end -}}
```
youkoulayley:
```suggestion
{{- end -}}

```

hub-manager/templates/_podtemplate.tpl
```diff
@@ -0,0 +1,117 @@
...
+        {{- with .Values.volumeMounts }}
+        volumeMounts:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
+        args:
+          - "serve"
+          - "--addr={{ .Values.address }}"
+
+          {{- with .Values.token }}
+          - "--hub.token=$(HUB_TOKEN)"
```
youkoulayley:
```suggestion
          - "--token=$(HUB_TOKEN)"
```

hub-manager/templates/_podtemplate.tpl
```diff
@@ -0,0 +1,117 @@
...
+
+          {{- with .Values.logs }}
+          - "--log-level={{ .level }}"
+          {{- end }}
+
+          {{- with .Values.tracing }}
+            {{- if .insecure }}
+          - "--tracing-insecure"
+            {{- end }}
+          {{- end }}
```
youkoulayley: If you put the tracing explicitly to `false` in the values.yaml. It will set it to `true` here.
I think it would be better to set it with the other env variable for the tracing:

```
- name: TRACING_INSECURE
  value: "{{ .Values.tracing.insecure }}"  
```

hub-manager/templates/_podtemplate.tpl
```diff
@@ -0,0 +1,117 @@
...
+        {{- tpl (toYaml .) $ | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "hub-manager.fullname" . }}
+      automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
+      {{- with .Values.deployment.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+      - image: {{- printf "%s/%s:%s" .Values.image.registry .Values.image.repository (.Values.image.tag | default .Chart.AppVersion) }}
```
youkoulayley:
```suggestion
      - image: {{ printf "%s/%s:%s" .Values.image.registry .Values.image.repository (.Values.image.tag | default .Chart.AppVersion) }}
```

The `-` trim the whitespace after `image:`. Not sure it's what we want.

hub-manager/templates/_podtemplate.tpl
```diff
@@ -0,0 +1,117 @@
...
+        name: {{ include "hub-manager.fullname" . }}
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        {{- with .Values.resources }}
+        resources: {{- toYaml . | nindent 10 }}
+        {{- end }}
+        readinessProbe:
+          httpGet:
+            path: /ready
+            port: 8080
```
youkoulayley: This is the address by default, but it can be changed with the `--address` parameter. Not sure anyone will change it, but it can be. WDYT of making this parametrizable?

hub-manager/templates/_podtemplate.tpl
```diff
@@ -0,0 +1,117 @@
...
+          failureThreshold: 2
+          initialDelaySeconds: 5
+          periodSeconds: 5
+        livenessProbe:
+          httpGet:
+            path: /live
+            port: 8080
```
youkoulayley: Same here

hub-manager/templates/rbac/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+metadata:
+  name: {{ include "hub-manager.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "hub-manager.labels" . | nindent 4 }}
+  annotations:
+  {{- with .Values.serviceAccount.annotations }}
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
```
youkoulayley:
```suggestion
  {{- with .Values.serviceAccount.annotations }}
  annotations:
  {{- toYaml . | nindent 4 }}
  {{- end }}
```

hub-manager/templates/_helpers.tpl
```diff
@@ -0,0 +1,67 @@
...
+
+{{/*
+Change input to a valid name for a port.
+This is a best effort to convert input to a valid port name for Kubernetes,
+which per RFC 6335 only allows lowercase alphanumeric characters and '-',
+and additionally imposes a limit of 15 characters on the length of the name.
+See also https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services
+and https://www.rfc-editor.org/rfc/rfc6335#section-5.1.
+*/}}
+{{- define "hub-manager.portname" -}}
```
youkoulayley: I may be wrong be I think it's not used anywhere.

hub-manager/values.yaml
```diff
@@ -0,0 +1,115 @@
...
+  # -- Name of Secret with key 'postgres-encryption-key' set to a valid encryption key
+  encryptionKey: ""
+
+tracing:
+  # -- Address to send traces
+  address: ""
+  # -- use HTTP instead of HTTPS
+  insecure: false
+  # -- Probability to send traces
+  probability: 0
```
youkoulayley: This is documented as an int in the `VALUES.md` but it's in fact a float.

youkoulayley (changes requested):

youkoulayley:
Last comment: There is no tests for this helm chart.

hub-manager/templates/_podtemplate.tpl
bpsoraggi: It was before, but we thought it would be simpler/better if we just set hard-coded defaults [[ref.](https://github.com/traefik/traefik-helm-chart/pull/1784#discussion_r3049409662)] WDYT?

youkoulayley: If we go that way, I think we need to at least put some documentation on the address parameter in the values.yaml.

hub-manager/templates/deployment.yaml
```diff
@@ -0,0 +1,31 @@
...
+  labels:
+    {{- include "hub-manager.labels" . | nindent 4 }}
+    {{- with .Values.deployment.labels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.deployment.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
```
youkoulayley, line 15: We also miss the migration of the database somewhere.
I'm thinking that maybe we should just have an `initContainer` to run the migration. If we need something better in the future we could update it at that time. WDYT?

hub-manager/templates/_podtemplate.tpl
bpsoraggi: Maybe we could get rid of it altogether? What do you think would be best?

hub-manager/templates/_podtemplate.tpl
```diff
@@ -0,0 +1,111 @@
...
+      annotations:
+        {{- tpl (toYaml .) $ | nindent 8 }}
+      {{- end }}
+    spec:
+      {{- with .Values.affinity }}
+      affinity:
+        {{- tpl (toYaml .) $ | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "hub-manager.fullname" . }}
+      automountServiceAccountToken: false
```
youkoulayley: This is not needed anymore

hub-manager/templates/deployment.yaml
```diff
@@ -0,0 +1,31 @@
...
+  {{- with .Values.deployment.updateStrategy }}
+  strategy:
+    type: {{ .type }}
+    {{- if (eq .type "RollingUpdate") }}
+    rollingUpdate:
+      maxUnavailable: {{ .rollingUpdate.maxUnavailable }}
+      maxSurge: {{ .rollingUpdate.maxSurge }}
+    {{- end }}
+  {{- end }}
+  template: {{ include "hub-manager.podTemplate" . | fromYaml | toYaml | nindent 4 }}
```
darkweaver87: So far we won't use an other kind of deployment than a `Deployment`. Don't you think it will be simpler to inject the `_podtemplate.tp ` here ?

hub-manager/templates/_podtemplate.tpl
```diff
@@ -0,0 +1,111 @@
...
+              secretKeyRef:
+                key: postgres-uri
+                name: {{ .uri }}
+          - name: POSTGRES_ENCRYPTION_KEY
+            valueFrom:
+              secretKeyRef:
+                key: postgres-encryption-key
+                name: {{ .encryptionKey }}
+          {{- end }}
+
```
youkoulayley:
```suggestion
          {{- with .Values.postgres }}
          {{- with .uri }}
          - name: POSTGRES_URI
            valueFrom:
              secretKeyRef:
                key: postgres-uri
                name: {{ .uri }}
          {{- end }}
          {{- with .encryptionKey }}
          - name: POSTGRES_ENCRYPTION_KEY
            valueFrom:
              secretKeyRef:
                key: postgres-encryption-key
                name: {{ .encryptionKey }}
          {{- end }}
          {{- end }}

```
`{{- with .Values.postgres }}` is always true because postgres is an object. I think we should guard each variable independently.

hub-manager/tests/deployment-config_test.yaml
```diff
@@ -0,0 +1,140 @@
...
+    asserts:
+      - notExists:
+          path: spec.replicas
+  - it: should not have replicas when autoscaling is enabled for the deployment
+    set:
+      autoscaling:
+        enabled: true
+    asserts:
+      - notExists:
+          path: spec.replicas
```
youkoulayley: This is a duplicate of the previous test

hub-manager/tests/notes_test.yaml
```diff
@@ -0,0 +1,27 @@
+suite: NOTES.txt output
+templates:
+  - NOTES.txt
+chart:
+  appVersion: v1.0.0
+tests:
+  - it: should use helm managed namespace in release notes output
+    set:
+      rbac:
+        namespaced: true
```
youkoulayley:
```suggestion
```

There is no rbac left in this helm-chart.

hub-manager/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: hub-manager
+description: A Helm chart for Kubernetes
```
youkoulayley:
```suggestion
description: A Helm chart for hub-manager
```

hub-manager/templates/migration-job.yaml
```diff
@@ -0,0 +1,55 @@
...
+    {{- with .Values.migration.labels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    {{- with .Values.migration.annotations }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
```
youkoulayley, line 17: I think we need the same `nodeSelector` and `tolerations` as the deployment to ensure the pod lands on the correct node if needed.

youkoulayley (changes requested):

hub-manager/templates/migration-job.yaml
```diff
@@ -0,0 +1,55 @@
...
+        resources: {{- toYaml . | nindent 10 }}
+        {{- end }}
+        {{- with .Values.securityContext }}
+        securityContext:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
+        args:
+          - "migrate"
+        env:
+          {{- with .Values.postgres }}
```
darkweaver87: Without `.Values.postgres` the YAML will be wrong with an empty `env`.

hub-manager/templates/migration-job.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "hub-manager.fullname" . }}-migration
```
darkweaver87, line 4: This job is not tested. Also for a migration job it might require a `backoffLimit`, probably to 0. Also we may add the same scheduling options as for the deployment so that people can schedule this job wherever they want.

hub-manager/templates/deployment.yaml
bpsoraggi: You mean get rid of `_podtemplate.tpl` and have just one file? If so, it could go either way imo. One is simpler but the other is a bit easier to read

@youkoulayley wdyt?

hub-manager/templates/migration-job.yaml
bpsoraggi: Regarding the scheduling, this job is actually a chart hook. The available hooks are documented [here](https://helm.sh/docs/topics/charts_hooks/#the-available-hooks)

hub-manager/templates/deployment.yaml
darkweaver87: Yes I mean one file. It will be easier to read IMO, especially for yaml indentation, etc.
Let's get @youkoulayley thoughts on it :-)

hub-manager/templates/migration-job.yaml
darkweaver87: I know what is a chart hook but I speak about kubelet scheduling like affinity/anti-affinity, tolerations, etc. It still remains a pod we have a execute somewhere at some point :-)

hub-manager/templates/deployment.yaml
youkoulayley: One file is ok for me, I don't have a strong opinion on this to be honest.

hub-manager/templates/_podtemplate.tpl
mloiseleur: make sense to me. @youkoulayley wdyt ?

youkoulayley: Make sense to me too

hub-manager/templates/deployment.yaml
```diff
@@ -0,0 +1,150 @@
...
+          {{- with .Values.logs }}
+          - "--log-level={{ .level }}"
+          {{- end }}
+
+          {{- if .Values.tracing.insecure }}
+          - "--tracing-insecure"
+          {{- end }}
+        env:
+          {{- with .Values.token }}
+          - name: HUB_TOKEN
```
youkoulayley:
```suggestion
          - name: TOKEN
```

hub-manager/tests/deployment-config_test.yaml
youkoulayley, line 1: We are missing a service for this deployment.

youkoulayley (changes requested):

hub-manager/tests/deployment-config_test.yaml
mloiseleur: @bpsoraggi Yes, I second that. we clearly need a `Service` in order for this `Deployment` to be reached by Traefik Hub.

hub-manager/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.service.enabled }}
```
mloiseleur: I guess this if should be removed ?

hub-manager/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "hub-manager.fullname" . }}
+  namespace: {{ .Release.Namespace . }}
+  labels:
+    {{- include "hub-platform.labels" . | nindent 4 }}
```
mloiseleur:
```suggestion
    {{- include "hub-manager.labels" . | nindent 4 }}
```

hub-manager/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
...
+  {{- end }}
+spec:
+  type: ClusterIP
+  ports:
+    - port: 8080
+      targetPort: http
+      protocol: TCP
+      name: http
+  selector:
+    {{- include "hub-platform.selectorLabels" . | nindent 4 }}
```
mloiseleur:
```suggestion
    {{- include "hub-manager.selectorLabels" . | nindent 4 }}
```

hub-manager/values.yaml
```diff
@@ -0,0 +1,115 @@
...
+
+migration:
+  # -- Additional migration job annotations
+  annotations: {}
+  # -- Additional migration job labels
+  labels: {}
+
+service:
+  # -- Additional annotations
+  annotations: {}
```
mloiseleur, line 55: Wdyt of adding a test to ensure this setting works as expected?

hub-manager/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
...
+{{- end }}
```
youkoulayley:
```suggestion
{{- end }}

```

hub-manager/templates/service.yaml
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    {{- include "hub-platform.labels" . | nindent 4 }}
+  {{- with .Values.service.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  type: ClusterIP
+  ports:
+    - port: 8080
```
youkoulayley:
```suggestion
    - port: 80
```

I think port 80 is better at the service level.

youkoulayley (changes requested):

youkoulayley (approved):

hub-manager/templates/NOTES.txt
```diff
@@ -0,0 +1,10 @@
+{{/* Print release information */}}
+{{- printf "\n\n" -}}
+{{ .Release.Name }} with {{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }} has been deployed successfully on {{ .Release.Namespace }} namespace!
+{{- printf "\n" -}}
+
+1. Get the application URL by running these commands:
+  export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "hub-manager.fullname" . }},app.kubernetes.io/instance={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
+  export CONTAINER_PORT=$(kubectl get pod --namespace {{ .Release.Namespace }} $POD_NAME -o jsonpath="{.spec.containers[0].ports[0].containerPort}")
+  echo "Visit http://127.0.0.1:8080 to use your application"
+  kubectl --namespace {{ .Release.Namespace }} port-forward $POD_NAME 8080:$CONTAINER_PORT
```
mloiseleur:
```suggestion
```
🤔 Is this really relevant in hub-manager context ?

bpsoraggi: Wasn't too sure, so I added it just in case. Would it be better to remove it, then?

mloiseleur: I guess so. It's a backend app, without UI.

hub-manager/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: hub-manager
+description: A Helm chart for hub-manager
+type: application
+version: 1.0.0
```
mloiseleur:
```suggestion
version: 1.0.0-rc.1
```

hub-manager/Chart.yaml
```diff
@@ -0,0 +1,6 @@
...
+appVersion: v1.0.0
```
mloiseleur:
```suggestion
appVersion: v1.0.0-rc.1
```

mloiseleur (changes requested):
Next steps: 
* EXAMPLES.md with a simple (and) complete example on how to use it, that can be reviewed and replayed.

hub-manager/EXAMPLES.md
```diff
@@ -0,0 +1,18 @@
+# EXAMPLES
+
+## Minimal installation
+
+Minimal setup for a single `hub-manager` instance backed by PostgreSQL.
+
+This example assumes a Secret named `hub-manager-config` already exists and contains:
+- `token`
+- `postgres-uri`
+- `postgres-encryption-key`
```
mloiseleur, line 19: Shouldn't we share or at least explain how to get/set this one?

mloiseleur (approved):

hub-manager/templates/migration-job.yaml
darkweaver87: `affinity` is honored by the Deployment but not by this Job, so a user setting `.Values.affinity` would see the migration Pod land somewhere different from the workload.

hub-manager/templates/_helpers.tpl
```diff
@@ -0,0 +1,52 @@
...
+*/}}
+{{- define "hub-manager.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/* Shared labels used for selector*/}}
+{{- define "hub-manager.labelselector" -}}
+app.kubernetes.io/name: {{ template "hub-manager.name" . }}
+{{- end }}
+
```
darkweaver87: While this should not happend, this one can conflict in case multiple helm chart instance are installed in the same instance.
```suggestion
```

hub-manager/templates/service.yaml
```diff
@@ -0,0 +1,20 @@
...
+  {{- end }}
+spec:
+  type: ClusterIP
+  ports:
+    - port: 80
+      targetPort: http
+      protocol: TCP
+      name: http
+  selector:
+    {{- include "hub-manager.labelselector" . | nindent 4 }}
```
darkweaver87:
```suggestion
    {{- include "hub-manager.selectorLabels" . | nindent 4 }}
```

hub-manager/templates/deployment.yaml
```diff
@@ -0,0 +1,150 @@
...
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if not .Values.autoscaling.enabled }}
+  replicas: {{ default 1 .Values.deployment.replicas }}
+  {{- end }}
+  selector:
+    matchLabels:
+      {{- include "hub-manager.labelselector" . | nindent 6 }}
```
darkweaver87:
```suggestion
      {{- include "hub-manager.selectorLabels" . | nindent 6 }}
```

Makefile
```diff
@@ -6,6 +6,7 @@ IMAGE_HELM_DOCS=jnorwood/helm-docs:v1.14.2
 IMAGE_HELM_UNITTEST=docker.io/helmunittest/helm-unittest:3.19.0-1.0.1
 
 traefik/tests/__snapshot__:
+	@mkdir hub-manager/tests/__snapshot__
 	@mkdir traefik/tests/__snapshot__
 	@mkdir traefik-crds/tests/__snapshot__
```
darkweaver87: Make it bullet proof:
```suggestion
	@mkdir -p hub-manager/tests/__snapshot__
	@mkdir -p traefik/tests/__snapshot__
	@mkdir -p traefik-crds/tests/__snapshot__
```

mloiseleur (dismissed):
This PR miss a workflow to release this chart.

hub-manager/templates/_helpers.tpl
mloiseleur: This helper "hub-manager.labelselector" is somehow a unused duplicate of "hub-manager.selectorLabels"
=> I agree that it can be removed

hub-manager/templates/_helpers.tpl
```diff
@@ -0,0 +1,52 @@
...
+{{/*
+Common labels
+*/}}
+{{- define "hub-manager.labels" -}}
+helm.sh/chart: {{ include "hub-manager.chart" . }}
+{{ include "hub-manager.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
```
mloiseleur:
```suggestion
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
```

.github/workflows/release.yaml
```diff
@@ -283,3 +284,130 @@ jobs:
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
+
+      - name: Create release
+        uses: ncipollo/release-action@339a81892b84b4eeb0f6e744e4574d79d0d9b8dd # v1.21.0
+        with:
+          tag: ${{ steps.tag_version.outputs.new_tag }}
+          name: ${{ steps.tag_version.outputs.new_tag }}
+          body: ${{ steps.changelog.outputs.changelog }}
+          prerelease: ${{ contains(steps.chart_version.outputs.CHART_VERSION, '-') }}
```
mloiseleur:
```suggestion
          body: ${{ steps.changelog.outputs.changelog }}
          makeLatest: "false"
          name: ${{ steps.tag_version.outputs.new_tag }}
          prerelease: ${{ contains(steps.chart_version.outputs.CHART_VERSION, '-') }}
          tag: ${{ steps.tag_version.outputs.new_tag }}
```
We do not want hub-manager to appear as "latest" chart, so maybe we can disable the auto latest feature of this action. Wdyt ?

.github/workflows/release.yaml
```diff
@@ -283,3 +284,130 @@ jobs:
...
+          passphrase: ${{ secrets.GPG_PASSPHRASE }}
+        if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
+
+      - name: Publish Helm chart to the ghcr.io registry
+        uses: appany/helm-oci-chart-releaser@d94988c92bed2e09c6113981f15f8bb495d10943 # v0.5.0
+        with:
+          name: hub-manager
+          repository: traefik/helm
+          tag: ${{ steps.chart_version.outputs.CHART_VERSION }}
+          path: ./hub-manager
```
mloiseleur, line 281: 🤔 Do we want to publish hub-manager chart and hub-manager public image in the same docker registry path ?

mloiseleur: Ah no, it will go in helm/hub-manager, like the two others. So it's good 👍

darkweaver87 (approved):
LGTM once @mloiseleur changes will be applied.

.github/workflows/release.yaml
```diff
@@ -283,3 +284,130 @@ jobs:
...
+        uses: mathieudutour/github-tag-action@a22cf08638b34d5badda920f9daf6e72c477b07b # v6.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          custom_tag: ${{ steps.chart_version.outputs.CHART_VERSION }}
+          tag_prefix: ${{ env.hub_manager_tag_prefix }}
+        if: steps.tag_exists.outputs.TAG_EXISTS == 'false'
+
+      - name: Render changelog configuration
+        env:
+          REGEXP: '(\\(.+\\)|(?!\\(CRDs-.+\\)))'
```
mloiseleur:
```suggestion
          REGEXP: '(\\(.+\\)|(?!\\(hub-manager\\)))'
```

mloiseleur:
@bpsoraggi After rebase, it will require for changelog something similar of what we have done for CRD chart:

1. An update in CONTRIBUTING.md (https://github.com/traefik/traefik-helm-chart/pull/1722/changes#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055)
2. An update in release & test-changelog workflows to include _hub-manager_ scope as the key to include it in hub-manager Changelog.

youkoulayley (approved):

youkoulayley (approved):

darkweaver87 (approved):
LGTM

PR #1787, opened 2026-04-06
feat: add service.nameOverride for adopting existing Services

zalbiraw:
## Summary

- Adds `service.nameOverride` to allow setting the Service name independently of the release name
- Enables the orphan-and-adopt migration pattern where Traefik needs to take ownership of an existing Service (e.g., `ingress-nginx-controller`) without `fullnameOverride` affecting all resources

## Use case

When migrating from NGINX Ingress Controller to Traefik using the [hostname retention approach](https://doc.traefik.io/traefik/migrate/nginx-to-traefik/#loadbalancer-hostname-retention), users orphan the NGINX LoadBalancer Service and need Traefik's Helm release to adopt it. Without `service.nameOverride`, enabling `service.enabled: true` creates a new Service (e.g., `traefik-ingress-nginx`) instead of adopting the existing `ingress-nginx-controller` Service. `fullnameOverride` cannot be used because it renames all resources (deployment, serviceaccount, etc.), not just the Service.

**Example values:**

```yaml
service:
  enabled: true
  nameOverride: ingress-nginx-controller
```

## Test plan

- [ ] `helm template` with `service.nameOverride` set produces a Service with the overridden name
- [ ] `helm template` without `service.nameOverride` produces the default Service name (no behavior change)
- [ ] `additionalServices` are unaffected by `service.nameOverride`

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -500,6 +500,7 @@ Kubernetes: `>=1.25.0-0`
 | service.annotationsUDP | object | `{}` | Additional annotations for UDP service only |
 | service.enabled | bool | `true` |  |
 | service.labels | object | `{}` | Additional service labels (e.g. for filtering Service by custom labels) |
+| service.nameOverride | string | `""` | Override the default Service name. Useful for adopting an existing Service (e.g., during migration from another ingress controller). |
 | service.single | bool | `true` |  |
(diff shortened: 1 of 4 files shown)


darkweaver87 (commented):
Shouldn't this PR be based on v40.0 branch if it relates to Nginx migration ?

zalbiraw:
@darkweaver87 sorry about that. Should I not be working off of master for this? I can certainly switch the base to v40 I was not aware

zalbiraw:
@darkweaver87 please let me know if this correct now :D

darkweaver87 (approved):
LGTM

PR #1320, opened 2025-01-27
docs: 📚️ fix typo in Guidelines.md

lbarnkow:
### What does this PR do?

It's just a typo in the docs.

### Motivation

I was reading through `Guidelines.md` as I was diffing a recent release with an old copy we're using internally.

### More

- [ ] Yes, I updated the tests accordingly
- [ ] Yes, I updated the schema accordingly
- [ ] Yes, I ran `make test` and all the tests passed

PR #1791, opened 2026-04-10
feat(chart): :memo: version support with annotations

darkweaver87:
### What does this PR do?

Declares supported Traefik Proxy and Hub version ranges as Chart.yaml annotations
(`traefik.io/proxy-min-version`, `proxy-max-version`, `hub-min-version`, `hub-max-version`),
replacing hardcoded values in templates.

### Motivation

Fixes https://github.com/traefik/traefik-helm-chart/issues/1758

### More

### DIFF
--- a/.github/fixtures/changelog-traefik.md
+++ b/.github/fixtures/changelog-traefik.md
@@ -13,8 +13,6 @@
 - docs: 📚️ fix typo in Guidelines.md #1320 by @lbarnkow
 - chore(release): publish v34.2.0 #1323 by @darkweaver87
 
-
-
 ## 👌 Traefik version support
 
 * Traefik Proxy: v3.6.0 -> v3.6.12 (default)
--- a/.github/fixtures/changelog-traefik.md
+++ b/.github/fixtures/changelog-traefik.md
@@ -1,4 +1,5 @@
-## [v34.2.0](https://github.com/traefik/traefik-helm-chart/compare/v34.1.0...v34.2.0) (2025-01-28)
+# [v34.2.0](https://github.com/traefik/traefik-helm-chart/compare/v34.1.0...v34.2.0) (2025-01-28)
+
 ## :rocket: Features
 
 - feat: :sparkles: add hub tracing capabilities #1322 by @darkweaver87
--- a/.github/fixtures/changelog-traefik.md
+++ b/.github/fixtures/changelog-traefik.md
@@ -14,3 +14,8 @@
 - chore(release): publish v34.2.0 #1323 by @darkweaver87
 
 
+
+## 👌 Traefik version support
+
+* Traefik Proxy: v3.6.0 -> v3.6.12 (default)
+* Traefik Hub: v3.19.0 -> v3.20.0-ea.1
--- a/.github/workflows/changelog.json
+++ b/.github/workflows/changelog.json
@@ -1,5 +1,5 @@
 {
-  "template": "## [#{{TO_TAG}}](#{{RELEASE_DIFF}}) (${DATE})\n#{{CHANGELOG}}${VERSION_SUPPORT}",
+  "template": "# [#{{TO_TAG}}](#{{RELEASE_DIFF}}) (${DATE})\n\n#{{CHANGELOG}}${VERSION_SUPPORT}",
   "categories": [
     {
       "title": "## :boom: BREAKING CHANGES",
--- a/.github/workflows/changelog.json
+++ b/.github/workflows/changelog.json
@@ -43,7 +43,7 @@
     }
   ],
   "pr_template": "- ${{TITLE}} ##{{NUMBER}} by @#{{AUTHOR}}",
-  "max_pull_requests": 300,
+  "max_pull_requests": 500,
   "empty_template": "- no changes",
   "transformers": [
     {
--- a/.github/workflows/changelog.json
+++ b/.github/workflows/changelog.json
@@ -43,7 +43,7 @@
     }
   ],
   "pr_template": "- ${{TITLE}} ##{{NUMBER}} by @#{{AUTHOR}}",
-  "max_pull_requests": 500,
+  "max_pull_requests": 300,
   "empty_template": "- no changes",
   "transformers": [
     {
--- a/.github/workflows/changelog.json
+++ b/.github/workflows/changelog.json
@@ -1,5 +1,5 @@
 {
-  "template": "## [#{{TO_TAG}}](#{{RELEASE_DIFF}}) (${DATE})\n#{{CHANGELOG}}",
+  "template": "## [#{{TO_TAG}}](#{{RELEASE_DIFF}}) (${DATE})\n#{{CHANGELOG}}${VERSION_SUPPORT}",
   "categories": [
     {
       "title": "## :boom: BREAKING CHANGES",
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -71,15 +71,7 @@ jobs:
       - name: Render changelog configuration
         env:
           REGEXP: '(\\(.+\\)|(?!\\(CRDs-.+\\)))'
-        run: |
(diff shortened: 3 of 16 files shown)


mloiseleur (commented):
First review.
1. Changelog is not updated as expected.
2. Tests on notes.txt will probably annoy us. Wdyt of reducing and/or simplifying them ?

darkweaver87:
Thanks for the revie @mloiseleur. Here my changes to address your comments:

- `test-changelog.yaml`: populate `VERSION_SUPPORT` so the fixture gets the `## 👌 Traefik version support` section.
- `changelog.json`: bump `max_pull_requests` 300 → 500 (the tag window holds 352 PRs, #1320 was being cut off — [example failing run](https://github.com/traefik/traefik-helm-chart/actions/runs/24779515411/job/72506506876)).
- `hack/test.sh`: copy chart to `/tmp` and pin version-support annotations there, decoupling tests from the real ranges. Same pin in `test-changelog`.
- Drop 3 redundant "no-warning when in range" NOTES tests.
- Rebased onto `v40.0`; Chart.yaml annotations adjusted (`proxy-max v3.7.0-rc.1`, `hub-min v3.19.3`, `hub-max v3.20.0-rc.1`).

.github/workflows/release.yaml
```diff
@@ -72,7 +72,14 @@ jobs:
         env:
           REGEXP: '(\\(.+\\)|(?!\\(CRDs-.+\\)))'
         run: |
-          export DATE=$(date +%F); cat .github/workflows/changelog.json | envsubst > /tmp/changelog.json; cat /tmp/changelog.json
+          export DATE=$(date +%F)
+          PROXY_MIN=$(awk '/traefik.io\/proxy-min-version:/{print $2}' traefik/Chart.yaml)
+          PROXY_MAX=$(awk '/traefik.io\/proxy-max-version:/{print $2}' traefik/Chart.yaml)
+          HUB_MIN=$(awk '/traefik.io\/hub-min-version:/{print $2}' traefik/Chart.yaml)
+          HUB_MAX=$(awk '/traefik.io\/hub-max-version:/{print $2}' traefik/Chart.yaml)
+          export VERSION_SUPPORT="\n\n## 👌 Traefik version support\n\n* Traefik Proxy: ${PROXY_MIN} -> ${PROXY_MAX} (default)\n* Traefik Hub: ${HUB_MIN} -> ${HUB_MAX}"
```
mloiseleur:
```suggestion
          export VERSION_SUPPORT="## 👌 Traefik version support\n\n* Traefik Proxy: ${PROXY_MIN} -> ${PROXY_MAX} (default)\n* Traefik Hub: ${HUB_MIN} -> ${HUB_MAX}"
```
When looking at the raw file, it seems there are far too many newlines. Wdyt ?

darkweaver87: Good catch! Applied. The root cause is that `#{{CHANGELOG}}` (from mikepenz's action) already ends with `\n\n` — trailing blank line after the last category block. So prepending another `\n\n` in `VERSION_SUPPORT` gave us 4 `\n` (3 blank lines) between the changelog and the version support section. Dropping the leading `\n\n` leaves just the action's own trailing `\n\n` = 1 blank line, which is proper markdown.

Note: the CRDs fixture has a similar residue (2 trailing blank lines) from the same `\n\n` trailing in CHANGELOG. Since we can't control the action's output, there's no clean way to fix it without a hack (post-processing trim), so leaving it as-is.

mloiseleur:
> changelog.json: bump max_pull_requests 300 → 500 (the tag window holds 352 PRs,

That seems very high. We have between 2 and 20 PRs for a single release. 
And it looks like we would have to increase it even more in a year or two. 
:thinking: Wdyt about adding a date filter ? Do we have any other better option, like starting from the previous tag ?

.github/workflows/release.yaml
```diff
@@ -72,7 +72,14 @@ jobs:
...
         run: |
-          export DATE=$(date +%F); cat .github/workflows/changelog.json | envsubst > /tmp/changelog.json; cat /tmp/changelog.json
+          export DATE=$(date +%F)
+          PROXY_MIN=$(awk '/traefik.io\/proxy-min-version:/{print $2}' traefik/Chart.yaml)
+          PROXY_MAX=$(awk '/traefik.io\/proxy-max-version:/{print $2}' traefik/Chart.yaml)
+          HUB_MIN=$(awk '/traefik.io\/hub-min-version:/{print $2}' traefik/Chart.yaml)
+          HUB_MAX=$(awk '/traefik.io\/hub-max-version:/{print $2}' traefik/Chart.yaml)
+          export VERSION_SUPPORT="## 👌 Traefik version support\n\n* Traefik Proxy: ${PROXY_MIN} -> ${PROXY_MAX} (default)\n* Traefik Hub: ${HUB_MIN} -> ${HUB_MAX}"
+          cat .github/workflows/changelog.json | envsubst > /tmp/changelog.json
+          cat /tmp/changelog.json
```
mloiseleur: This code is duplicated between `release.yaml` and `test-changelog.yaml`.
=> Wdyt about putting this code in a dedicated script and/or makefile target ?

darkweaver87: Good point.

darkweaver87:
> > changelog.json: bump max_pull_requests 300 → 500 (the tag window holds 352 PRs,
> 
> That seems very high. We have between 2 and 20 PRs for a single release. And it looks like we would have to increase it even more in a year or two. 🤔 Wdyt about adding a date filter ? Do we have any other better option, like starting from the previous tag ?

`fetchViaCommits: true` seems to fix our issue :-)

.github/workflows/changelog.json
```diff
@@ -1,5 +1,5 @@
 {
-  "template": "## [#{{TO_TAG}}](#{{RELEASE_DIFF}}) (${DATE})\n#{{CHANGELOG}}",
+  "template": "## [#{{TO_TAG}}](#{{RELEASE_DIFF}}) (${DATE})\n#{{CHANGELOG}}${VERSION_SUPPORT}",
```
mloiseleur:
```suggestion
  "template": "# [#{{TO_TAG}}](#{{RELEASE_DIFF}}) (${DATE})\n\n#{{CHANGELOG}}${VERSION_SUPPORT}",
```

.github/fixtures/changelog-traefik.md
mloiseleur:
```suggestion
# [v34.2.0](https://github.com/traefik/traefik-helm-chart/compare/v34.1.0...v34.2.0) (2025-01-28)

## :rocket: Features
```

hack/test.sh
```diff
@@ -2,5 +2,19 @@
 
 set -e
 
-/usr/bin/helm unittest --color ./traefik
+# Test against a copy of the traefik chart with pinned version-support
```
mloiseleur: Seems a bit excessive to copy everything when we need here to set specific value for consistent test.
Wdyt about:

1. Open an issue on https://github.com/helm-unittest/helm-unittest about the capacity to set this kind of field
2. Explore if we can use [postRenderer](https://github.com/helm-unittest/helm-unittest/blob/main/DOCUMENT.md) to avoid this 
3. Use helm plugin install of docker, so it could at least be done in place, by saving current values and restoring them afterwards (example [here](https://oneuptime.com/blog/post/2026-01-17-helm-chart-testing-unittest-conftest/view#github-actions))

darkweaver87: 1. yes this is the long-term response to this problem
2. postRenderer transforms the output but not the evaluation of templates (cf. https://github.com/helm-unittest/helm-unittest/blob/main/pkg/unittest/test_job.go#L491; TLDR if statements are evaluated at rendering time)
3. done in https://github.com/traefik/traefik-helm-chart/pull/1791/commits/0fcd9c8a128978ca424a4f00ebfdce2b1e835ee9

mloiseleur (approved):

PR #1796, opened 2026-04-13
feat(podtemplate): allow disabling automatic GOMEMLIMIT

johnvanhienen:
### What does this PR do?
Allow opting out of the automatic GOMEMLIMIT env var injection by setting deployment.goMemLimitPercentage to 0. When set to 0, the chart skips adding GOMEMLIMIT, letting users provide their own via the env values instead.

### Motivation
When using a Vertical Pod Autoscaler (VPA), the memory limit changes dynamically. The chart's static GOMEMLIMIT (calculated at deploy time of the Helm release) becomes stale after VPA adjusts the limit. Setting GOMEMLIMIT via env with a resourceFieldRef to limits.memory solves this.

The resourceFieldRef solution was implemented before merging #1418 , which calculates it based on the ratio. For people that don't use the VPA this would probably be a great default setting, but for people that do use the VPA it currently makes it difficult. A relatively low GOMEMLIMIT causes the garbage collector to use a lot of CPU. 

Besides the solution above I've also considered not setting the `memory.limit` to prevent the generation of the GOMEMLIMIT env var, but I use the `RequestsAndLimits` [mode](https://kubernetes.io/docs/concepts/workloads/autoscaling/vertical-pod-autoscale/#controlledvalues) on the VPA object, since I do want to have a limit set, since GOMEMLIMIT is just a soft limit.  

By introducing this change people are able to use their own solution for the GOMEMLIMIT or provide it manually via resourceFieldRef.

### More

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -50,7 +50,7 @@ Kubernetes: `>=1.25.0-0`
 | deployment.dnsConfig | object | `{}` | Custom pod [DNS config](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#poddnsconfig-v1-core) |
 | deployment.dnsPolicy | string | `""` | Custom pod DNS policy. Apply if `hostNetwork: true` |
 | deployment.enabled | bool | `true` | Enable deployment |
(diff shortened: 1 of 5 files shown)


darkweaver87 (approved):
LGTM

mloiseleur (approved):

PR #1803, opened 2026-04-17
fix(CRDs): content item admission webhook name

mdeliatf:
### What does this PR do?

Fix the content item admission webhook name.

Fix: https://github.com/traefik/hub-issues/issues/1797

### More


darkweaver87 (approved):
LGTM

Issue #1820, opened 2026-04-29
support pinning by digest for stronger supply chain integrity


zalbiraw:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

## Context

Today, `image.tag` is the only way to pick which Traefik image is deployed.
Tags are *mutable* references — anyone with push access to the registry
namespace, or a successful registry-side compromise / namespace takeover /
typosquat, can repoint the same tag to different content at any time. The
chart's `tag` field also can't carry a digest (`sha256:…`) because the rendered
string is `registry/repository:<tag>` — a digest contains a colon and would
produce a malformed reference.

The `image.tag` doc comment already gestures at this gap
("When a digest is required, `versionOverride` can be used to set the version"),
but there is no actual field to express the digest pin.

## Proposal

Add `image.digest` next to `image.tag` in the `image` section. When set, it
takes precedence over `image.tag` and the rendered reference switches to the
`@`-separated form:

```yaml
image:
  registry: docker.io
  repository: traefik
  digest: sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef

**Comment by zozs** [comment]
Worth noticing that it is possible today by using the "workaround" of combination of tag + digest like this:

```yaml
image:
  tag: v3.6.13@sha256:34d5089d0b414945342848518b383f11f5b3a645504ed87b77ffeb9d683d0e48
```

Which gives the following result:

```
❯ helm template traefik traefik --set image.tag="v3.6.13@sha256:34d5089d0b414945342848518b383f11f5b3a645504ed87b77ffeb9d683d0e48" | grep -C 2 docker.io
        - name: USER
          value: traefik
        image: docker.io/traefik:v3.6.13@sha256:34d5089d0b414945342848518b383f11f5b3a645504ed87b77ffeb9d683d0e48
        imagePullPolicy: IfNotPresent
        lifecycle: null

```

The tag is mostly for documentation purposes in this case, it's the digest that will be validated which is what you typically want anyway :).

umglurf:
Hi

If Traefik Hub is enabled, then this macro is invoked and fails: https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/_helpers.tpl#L185

zalbiraw:
[...]

PR #1821, opened 2026-04-29
feat(image): add image.digest for digest-based pinning

zalbiraw:
### What does this PR do?

Adds `image.digest` to the `image` section of `values.yaml`. When set, it takes precedence over `image.tag` and the rendered image reference switches to the `registry/repository@sha256:<hex>` form. The field is validated by a strict schema pattern (`^sha256:[a-f0-9]{64}$`) so misconfiguration fails at `helm install/upgrade` time rather than on the cluster. `versionOverride` continues to be how the chart's version-checking logic learns the version when pinning by digest.

### Motivation

Closes (https://github.com/traefik/traefik-helm-chart/issues/1820).

Tags are *mutable* pointers; digests are immutable, content-addressable hashes. Pinning by digest is the baseline supply-chain control recommended by NIST SP 800-204D, the CNCF Software Supply Chain Best Practices, and CIS / NSA Kubernetes hardening guidance. Today, `image.tag` cannot carry a digest because the rendered string uses a `:` separator (`registry/repo:<tag>`), which collides with the `sha256:` prefix. The `image.tag` doc comment already acknowledged this gap ("When a digest is required, `versionOverride` can be used to set the version") without offering a field to express it. This PR closes that gap.

### More

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -213,6 +213,7 @@ Kubernetes: `>=1.25.0-0`
 | hub.tracing.additionalTraceHeaders.traceContext.traceId | string | `""` | Name of the header that will contain the trace-id copy. |
 | hub.tracing.additionalTraceHeaders.traceContext.traceParent | string | `""` | Name of the header that will contain the traceparent copy. |
 | hub.tracing.additionalTraceHeaders.traceContext.traceState | string | `""` | Name of the header that will contain the tracestate copy. |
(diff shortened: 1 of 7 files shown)


darkweaver87 (approved):
LGTM

zalbiraw:
@darkweaver87 sorry but I added a few more changes based on feedback on the issue

mloiseleur (approved):

mloiseleur:
@zalbiraw Can you rebase ? We should be good to go after that.

darkweaver87 (approved):
LGTM

zalbiraw:
Sorry @mloiseleur I was AFK most of the day. Thank you for handling it @darkweaver87! 🚀

PR #1834, opened 2026-05-06
ci(renovate): disable renovate self rate limit

mloiseleur:
### What does this PR do?

Disable renovate self rate limit.

### Motivation

This feature is not interesting in our use case.


darkweaver87 (approved):
LGTM

PR #1861, opened 2026-05-21
feat(providers.file)!: switch content to an object

### LINKED ISSUE #1860: support yaml for providers.file.content
### Welcome!

It would be easier for me to lint/validate my file-provider config if it could be actual yaml instead of yaml-in-yaml-string (syntax highlighting and validation, etc.), i.e. instead of:

```yaml
file:
  enabled: true
  content: |
    http:
      middlewares:...
```

```yaml
file:
  enabled: true
  content:
    http:
      middlewares:...
```

Adding it to the configmap should be only a change from:

```
{{ content | toYaml | nindent 4 }}
```

to:

```
{{ content | toYaml | quote | nindent 4 }}
```

In order to be backward-compatible, we could do this as a new `contentYaml` or `contentData` field. Either that or a switch on the type of `content`, whichever is preferable.

Would a PR be welcome? And if so, any preference for adding a switch for the type of `content` or a sibling field?

minrk:
### What does this PR do?

allows structured file content in values.yaml rather than requiring yaml-as-string-in-yaml.

Before:

```yaml
providers:
  file:
    enabled: true
    content: |
      http:
        middlewares: ...
```

After:

```yaml
providers:
  file:
    enabled: true
    content:
      http:
        middlewares: ...
```

This is a backward-incompatible change because it changes the type of `providers.file.content` from string to object, but updating (at least for standard values.yaml deployments) is a matter of removing `|` after `content:`.

If there's a good reason to support both strings and objects, I can do that, too. I don't know if there's a situation where a string is easy and yaml/json is hard.

closes #1860

### Motivation

yaml strings of yaml files in yaml values are harder to read/write/validate than regular yaml structures, where authors get syntax highlighting, autoindent, schema validation, etc.

### More

### Future

Importing a json schema for the yaml config file itself would allow validation of the full content of this file, as [suggested here](https://github.com/traefik/traefik-helm-chart/issues/1860#issuecomment-4505287563), but that is not done in this PR.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -1980,7 +1980,7 @@ hub:
 providers:
   file:
     enabled: true
-    content: |
+    content:
       http:
         uplinks:
           whoami:
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -445,7 +445,7 @@ Kubernetes: `>=1.25.0-0`
 | ports.websecure.transport.respondingTimeouts.readTimeout | string | `nil` |  |
 | ports.websecure.transport.respondingTimeouts.writeTimeout | string | `nil` |  |
 | priorityClassName | string | `""` | [Pod Priority and Preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/) |
-| providers.file.content | string | `""` | File content (YAML format, go template supported) (see https://doc.traefik.io/traefik/reference/install-configuration/providers/others/file/) |
+| providers.file.content | object | `{}` | File content as an object (will be YAML-formatted, go template supported) (see https://doc.traefik.io/traefik/reference/install-configuration/providers/others/file/) |
 | providers.file.enabled | bool | `false` | Create a file provider |
 | providers.file.watch | bool | `true` | Allows Traefik to automatically watch for file changes |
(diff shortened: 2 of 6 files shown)


darkweaver87:
Hello @minrk,

Thanks for this PR. WDYT about supporting both ?

```yaml
# schema: type: [string, object]
content:
  {{- if kindIs "string" .Values.providers.file.content }}
  {{ .Values.providers.file.content | quote | nindent 4 }}
  {{- else }}
  {{ toYaml .Values.providers.file.content | quote | nindent 4 }}
  {{- end }}
```

minrk:
Sure! @mloiseleur [suggested](https://github.com/traefik/traefik-helm-chart/issues/1860#issuecomment-4505287563) doing it as a breaking change, so that's what I did, but supporting both seems fine to me, if that's preferable.

minrk:
providers.file.content now accepts string

minrk:
The schema test doesn't like representing both types using `oneOf`. I'm not sure how to get `helm schema` to generate `oneOf` types, but the chart accepts strings, at least. Not sure what you'd like to do about that.

darkweaver87 (approved):
OK @minrk then don't bother then. I'm fine with the breaking change. LGTM

PR #1862, opened 2026-05-23
fix(ingressroute): fail fast on uppercase ingressRoute keys (RFC 1123)

filip2mac:
## What does this PR do?

Adds a pre-flight validation in `requirements.yaml` that fails with a clear, actionable error when any `ingressRoute` map key contains uppercase characters.

## Motivation

`IngressRoute` names are generated as `{release}-{key}`, where `{key}` is the map key from `.Values.ingressRoute`. Uppercase characters produce names like `release-kubernetesCRD` which Kubernetes rejects:

```
metadata.name: Invalid value: "traefik-kubernetesCRD": a lowercase RFC 1123 subdomain
must consist of lower case alphanumeric characters, '-' or '.'...
```

The root cause is often a user confusing `ingressRoute.<name>` with `providers.<name>`. Rather than silently lowercasing the key, this PR follows the chart's existing pattern (see `requirements.yaml`) of failing fast with a message that explains both the problem and the likely fix:

```
ERROR: ingressRoute key "kubernetesCRD" contains uppercase characters and would generate
an invalid Kubernetes resource name (RFC 1123 requires lowercase). If you meant to enable
a provider, use providers.kubernetesCRD instead.
```

## More

### DIFF
--- a/traefik/templates/requirements.yaml
+++ b/traefik/templates/requirements.yaml
@@ -34,7 +34,7 @@
 
 {{- range $name, $config := .Values.ingressRoute }}
   {{- if regexMatch "[A-Z]" $name }}
-    {{- fail (printf "ERROR: ingressRoute key %q contains uppercase characters and would generate an invalid Kubernetes resource name (RFC 1123 requires lowercase). If you meant to enable a provider, use providers.%s instead." $name $name) }}
+    {{- fail (printf "ERROR: ingressRoute key %q contains uppercase characters and would generate an invalid Kubernetes resource name (RFC 1123 requires lowercase)." $name) }}
   {{- end }}
 {{- end }}
 
--- a/traefik/templates/requirements.yaml
+++ b/traefik/templates/requirements.yaml
@@ -32,6 +32,12 @@
   {{- end }}
 {{- end }}
 
+{{- range $name, $config := .Values.ingressRoute }}
+  {{- if regexMatch "[A-Z]" $name }}
+    {{- fail (printf "ERROR: ingressRoute key %q contains uppercase characters and would generate an invalid Kubernetes resource name (RFC 1123 requires lowercase). If you meant to enable a provider, use providers.%s instead." $name $name) }}
+  {{- end }}
+{{- end }}
+
 {{- if and (semverCompare "<v3.6.2-0" $version) (.Values.providers.kubernetesIngressNGINX).enabled}}
   {{- fail "ERROR: Kubernetes Ingress NGINX provider is only available for traefik >= v3.6.2." }}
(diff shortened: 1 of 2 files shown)


traefik/templates/requirements.yaml
```diff
@@ -31,6 +31,12 @@
   {{- end }}
 {{- end }}
 
+{{- range $name, $config := .Values.ingressRoute }}
+  {{- if regexMatch "[A-Z]" $name }}
+    {{- fail (printf "ERROR: ingressRoute key %q contains uppercase characters and would generate an invalid Kubernetes resource name (RFC 1123 requires lowercase). If you meant to enable a provider, use providers.%s instead." $name $name) }}
```
mloiseleur:
```suggestion
    {{- fail (printf "ERROR: ingressRoute key %q contains uppercase characters and would generate an invalid Kubernetes resource name (RFC 1123 requires lowercase)." $name) }}
```

traefik/tests/requirements-config_test.yaml
```diff
@@ -635,3 +635,18 @@ tests:
...
     asserts:
       - notFailedTemplate: {}
+  - it: should fail when an ingressRoute key contains uppercase characters
+    set:
+      ingressRoute:
+        kubernetesCRD:
+          enabled: true
+    asserts:
+      - failedTemplate:
+          errorMessage: 'ERROR: ingressRoute key "kubernetesCRD" contains uppercase characters and would generate an invalid Kubernetes resource name (RFC 1123 requires lowercase). If you meant to enable a provider, use providers.kubernetesCRD instead.'
```
```suggestion
          errorMessage: 'ERROR: ingressRoute key "kubernetesCRD" contains uppercase characters and would generate an invalid Kubernetes resource name (RFC 1123 requires lowercase).'
```

mloiseleur:
@filip2mac I've reviewed your PR. Globally, it improves the user XP, thanks :+1: . On the details, the second part of the message does not make sense to me, so I suggest to remove it. Feel free to share with me if I missed something.

traefik/templates/requirements.yaml
filip2mac: Done, thanks!

traefik/tests/requirements-config_test.yaml
filip2mac: Done, thanks!

filip2mac:
Thanks for the review! Agreed, the hint about `providers.*` was speculative and not something we can reliably derive from the key name, so removing it keeps the message clean and accurate.

mloiseleur (approved):

darkweaver87 (approved):
LGTM

PR #1877, opened 2026-06-08
chore(ci): track proxy & hub max-version annotations with renovate

jspdown:
### What does this PR do?

Lets Renovate auto-bump the `traefik.io/proxy-max-version` and `traefik.io/hub-max-version` annotations in `Chart.yaml` (and the matching version in the requirements tests' error-message assertions) from the latest stable `traefik` / `ghcr.io/traefik/traefik-hub` releases.

### Motivation

These annotations were bumped by hand because Renovate only tracked the bare `appVersion`, so new Traefik Hub / Proxy releases never opened a PR.

### More


darkweaver87 (approved):
LGTM

PR #1884, opened 2026-06-09
feat(version): :sparkles: relax max-version guard to warn on minor/patch, fail only on major mismatch

### LINKED ISSUE #1883: Reconsider hard-failing on the Traefik Proxy/Hub max version
## Context

[PR #1791](https://github.com/traefik/traefik-helm-chart/pull/1791) (fixing #1758) declared the supported Traefik Proxy and Hub version ranges as `Chart.yaml` annotations:

```yaml
annotations:
  traefik.io/proxy-min-version: v3.6.0
  traefik.io/proxy-max-version: v3.7.1
  traefik.io/hub-min-version: v3.19.3
  traefik.io/hub-max-version: v3.20.2
```

Alongside moving the previously hardcoded values into annotations, the PR **added a new hard failure on the max version**. Before #1791, only the *min* version was enforced.

Current logic in `traefik/templates/requirements.yaml`:

```
min:  version < min                         -> fail   (any version, stable or not)
max:  isStableVersion AND version > max      -> fail
```

(`NOTES.txt` additionally prints a *warning* — not a failure — for non-stable versions above max.)

## The concern

Enforcing the **min** version is sound: the chart emits configuration/CRDs that depend on features introduced in Traefik >= min. Running an older proxy would produce broken config, so failing fast at template time is the right behavior. No objection here.

Enforcing the **max** version as a hard failure is more questionable:

1. **It blocks benign patch/minor upgrades.** A user on chart `40.2.0` who sets `image.tag: v3.7.2` (a bugfix or CVE patch) gets:

   ```
   ERROR: This version of the Chart only supports Traefik Proxy up to v3.7.1.
   ```

   even though nothing the chart renders actually breaks. The chart isn't detecting a known incompatibility — it simply doesn't know that version exists yet, because it was released after the chart.

2. **Max encodes absence of knowledge, not presence of breakage.** "We haven't tested above X" is different from "X is known to break." Hard-failing on *untested* forces a chart release for every new proxy release, even pure patches.

3. **It forces chart + image upgrades into lockstep.** A common, legitimate pattern is bumping the proxy image ahead of a chart release — e.g. to pick up a security fix quickly. The max gate makes that impossible without also bumping the chart.

4. **The current rule is internally inconsistent.** Max only fails for **stable** versions. A pre-release *above* max (e.g. `v3.8.0-rc.1`) only triggers a warning in `NOTES.txt`, while a **stable patch** (`v3.7.2`, generally the safer choice) hard-fails. The "warn, don't fail" decision was already made for pre-releases; the stable path is arguably where it matters more.

## Proposal

Relax the max-version handling so it no longer blocks benign upgrades, in two complementary parts:

- **Downgrade max to a warning for patch/minor versions.** Mirror the existing `NOTES.txt` behavior for stable versions above max, instead of failing. Internally consistent with how pre-releases are already treated.
- **Keep a hard failure only on a major-version mismatch.** Block e.g. `v4.x` on a chart that targets `v3.x`, where breaking CRD/config changes are plausible.

Together these keep protection against genuinely breaking jumps while no longer blocking patch upgrades or decoupled image bumps. **Min stays a hard failure.**

The same reasoning applies symmetrically to the Hub max version (`traefik.io/hub-max-version`).

darkweaver87:
### What does this PR do?

Reworks the Traefik Proxy/Hub max-version install guard so it no longer blocks benign upgrades, implementing the proposal in #1883.

Previously, a stable version above `proxy-max-version` / `hub-max-version` hard-failed at template time, while pre-releases above max only triggered a warning. This forced a chart release for every new proxy/hub patch and prevented bumping the image ahead of the chart (e.g. to pick up a CVE fix quickly).

New behavior:

- **Minor/patch above max → warning, not failure.** A version above max but with the same major (e.g. `v3.99.0` on a chart targeting `v3.7.4`) now templates successfully and prints a "use at your own risk" warning in `NOTES.txt`, consistent with how pre-releases were already treated.
- **Major-version mismatch → hard failure.** A version with a higher major (e.g. `v4.x` on a `v3.x` chart, stable or pre-release) still fails, since breaking CRD/config changes are plausible.
- **Min-version guard unchanged.** Versions below the supported minimum keep failing — they are genuinely incompatible.

The same logic is applied symmetrically to the Hub max version.

This supersedes the previous `skipMaxVersionGuard` opt-in flag approach: no new chart value is introduced.

Closes #1883

### More

### DIFF
--- a/traefik/templates/requirements.yaml
+++ b/traefik/templates/requirements.yaml
@@ -5,8 +5,9 @@
   {{- if (semverCompare (printf "<%s-0" $proxyMinVersion) $version) }}
     {{- fail (printf "ERROR: This version of the Chart only supports Traefik Proxy %s+" $proxyMinVersion) -}}
   {{- end }}
-  {{- if and (eq (include "traefik.isStableVersion" $version) "true") (semverCompare (printf ">%s" $proxyMaxVersion) $version) }}
-    {{- fail (printf "ERROR: This version of the Chart only supports Traefik Proxy up to %s." $proxyMaxVersion) -}}
+  {{/* Only a major-version jump is a hard failure; minor/patch above max is just a warning (see NOTES.txt). */}}
+  {{- if semverCompare (printf ">=%d.0.0-0" (add1 (int (semver $proxyMaxVersion).Major))) $version }}
+    {{- fail (printf "ERROR: This version of the Chart does not support Traefik Proxy %s, which is a major version above the supported %s. Breaking changes are likely." $version $proxyMaxVersion) -}}
   {{- end }}
 {{- end }}
 
@@ -151,8 +152,9 @@
   {{- if semverCompare (printf "<%s-0" $hubMinVersion) $hubVersion }}
     {{ fail (printf "ERROR: this Chart supports *only* Traefik Hub >= %s." $hubMinVersion) }}
   {{- end }}
-  {{- if and (eq (include "traefik.isStableVersion" $hubVersion) "true") (semverCompare (printf ">%s" $hubMaxVersion) $hubVersion) }}
-    {{ fail (printf "ERROR: this Chart only supports Traefik Hub up to %s." $hubMaxVersion) }}
+  {{/* Only a major-version jump is a hard failure; minor/patch above max is just a warning (see NOTES.txt). */}}
+  {{- if semverCompare (printf ">=%d.0.0-0" (add1 (int (semver $hubMaxVersion).Major))) $hubVersion }}
+    {{ fail (printf "ERROR: this Chart does not support Traefik Hub %s, which is a major version above the supported %s. Breaking changes are likely." $hubVersion $hubMaxVersion) }}
   {{- end }}
 
   {{- if and (not $.Values.tracing.otlp.enabled) .Values.hub.tracing.additionalTraceHeaders.enabled }}
--- a/traefik/templates/requirements.yaml
+++ b/traefik/templates/requirements.yaml
@@ -131,16 +131,10 @@
 {{- end }}
 
 {{- if $.Values.hub.token -}}
-  {{ $hubVersion := $.Values.oci_meta.enabled | ternary $.Values.oci_meta.images.hub.tag $.Values.image.tag }}
-  {{ $hubVersion = ($.Values.global.azure.enabled | ternary $.Values.global.azure.images.hub.tag $hubVersion) }}
-  {{- if and (not $hubVersion) $.Values.versionOverride }}
-    {{- $hubVersion = $.Values.versionOverride }}
-  {{- end }}
+  {{- $hubVersion := include "traefik.hubVersion" $ }}
   {{ if not $hubVersion }}
     {{ fail "When using Traefik Hub, image.tag must be set, or image.digest with versionOverride." }}
   {{- end -}}
-
-  {{ $hubVersion = (split "@" (default "v3" $hubVersion))._0 }}
 
   {{/* Consider non semver versions as latest one  */}}
   {{- if not (regexMatch "v[0-9]+.[0-9]+.[0-9]+" (default "" $hubVersion)) -}}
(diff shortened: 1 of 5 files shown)


mmatur (commented):
Thanks for the PR

My concern: this fixes the symptom, not the root cause.

The max-version guard fails on `> proxy-max-version` / `hub-max-version`, but that annotation is tracked at **patch** granularity (`v3.7.4`). By semver, a patch bump (`z` in `x.y.z`) is backward-compatible. A new Traefik patch can't introduce a field or feature the Chart doesn't already know. So hard-failing on a patch overshoot is a structural false positive: it treats "not yet validated" as "broken".

The result is that every new Traefik/Hub patch requires either a Chart release or the operator setting this flag. Even though Traefik, Traefik Hub and the Chart are separate products on separate release cadences.

WDYT about handling the patch case at the guard level and keeping the flag for minor overshoots only?

darkweaver87:
@mmatur I finally implemented the proposal #1883

pepordev:
> @mmatur I finally implemented the proposal #1883

any ETA for this to land? we need some features from the v3.7.5 and while planning to do a pr to bump the versions i saw this one.

traefik/templates/requirements.yaml
```diff
@@ -5,8 +5,9 @@
   {{- if (semverCompare (printf "<%s-0" $proxyMinVersion) $version) }}
     {{- fail (printf "ERROR: This version of the Chart only supports Traefik Proxy %s+" $proxyMinVersion) -}}
   {{- end }}
-  {{- if and (eq (include "traefik.isStableVersion" $version) "true") (semverCompare (printf ">%s" $proxyMaxVersion) $version) }}
-    {{- fail (printf "ERROR: This version of the Chart only supports Traefik Proxy up to %s." $proxyMaxVersion) -}}
+  {{/* Only a major-version jump is a hard failure; minor/patch above max is just a warning (see NOTES.txt). */}}
+  {{- if semverCompare (printf ">=%d.0.0-0" (add1 (int (semver $proxyMaxVersion).Major))) $version }}
+    {{- fail (printf "ERROR: This version of the Chart does not support Traefik Proxy %s, which is a major version above the supported %s. Breaking changes are likely." $version $proxyMaxVersion) -}}
```
mloiseleur:
```suggestion
    {{- fail (printf "ERROR: This version of the Chart does not support Traefik Proxy %s, which is a major version above the supported %s." $version $proxyMaxVersion) -}}
```

traefik/templates/requirements.yaml
```diff
@@ -151,8 +152,9 @@
   {{- if semverCompare (printf "<%s-0" $hubMinVersion) $hubVersion }}
     {{ fail (printf "ERROR: this Chart supports *only* Traefik Hub >= %s." $hubMinVersion) }}
   {{- end }}
-  {{- if and (eq (include "traefik.isStableVersion" $hubVersion) "true") (semverCompare (printf ">%s" $hubMaxVersion) $hubVersion) }}
-    {{ fail (printf "ERROR: this Chart only supports Traefik Hub up to %s." $hubMaxVersion) }}
+  {{/* Only a major-version jump is a hard failure; minor/patch above max is just a warning (see NOTES.txt). */}}
+  {{- if semverCompare (printf ">=%d.0.0-0" (add1 (int (semver $hubMaxVersion).Major))) $hubVersion }}
+    {{ fail (printf "ERROR: this Chart does not support Traefik Hub %s, which is a major version above the supported %s. Breaking changes are likely." $hubVersion $hubMaxVersion) }}
```
mloiseleur:
```suggestion
    {{ fail (printf "ERROR: this Chart does not support Traefik Hub %s, which is a major version above the supported %s." $hubVersion $hubMaxVersion) }}
```

mloiseleur (commented):
I do not see any warning or major version guard test for Traefik Hub in this PR.

@darkweaver87 🤔 Shouldn't we have the same logic for Traefik Hub ?

mloiseleur (approved):
LGTM. Code is clearer and both Proxy & Hub are covered.

PR #1885, opened 2026-06-09
feat(hub): :sparkles: install out-of-box with only hub.token set

darkweaver87:
### What does this PR do?

Aligns a default Traefik Hub install with Traefik Proxy. With only `hub.token` set and the image left at its defaults, the chart now resolves the full Hub image automatically:

- `registry` → `ghcr.io`
- `repository` → `traefik/traefik-hub`
- `tag` → `traefik.io/hub-max-version`

So `--set hub.token=xxx` installs a working Hub on the latest supported version, just as Proxy installs `Chart.AppVersion` out of the box. Any field the user provides explicitly is respected; the `image.digest` case still requires `versionOverride`.

Driven by three helpers (`traefik.imageRegistry`, `traefik.imageRepository`, `traefik.defaultTag`) used by `image-name`, `NOTES.txt` and the `requirements.yaml` guard.

### Motivation

Hub previously failed with `image.tag must be set` and forced the user to spell out registry, repository and tag, while Proxy worked out of the box. This makes both behave the same.

### More

- [x] Yes, I updated the tests accordingly
- [ ] Yes, I updated the schema accordingly
- [x] Yes, I ran `make test` and all the tests passed

PR #1886, opened 2026-06-12
fix(notes): :memo: use traefik.image-name so NOTES match deployed image

mloiseleur:
### What does this PR do?

Makes the install NOTES print the same image the container actually runs.
For `oci_meta` and `global.azure` installs, NOTES showed the plain `image.registry/image.repository` instead of the OCI/Azure image. NOTES now delegates to `traefik.image-name`, the helper the container spec uses.

### Motivation

The post-install message should never lie about what was deployed. Single source of truth for the image string across container spec and NOTES.

### More

> [!NOTE]
> Overlaps #1885 (both touch the NOTES image line) — whichever lands second will need a trivial rebase.


darkweaver87 (approved):
LGTM

PR #1889, opened 2026-06-12
ci: track max-version assertions with relaxed guard wording

mloiseleur:
## Context

Renovate keeps the Proxy/Hub `*-max-version` test assertions in sync via a `customManager` regex. 

#1884 reworded the guard message (`only supports... up to vX` → `... above the supported vX`), so the regex went stale and matched nothing.

## Motivation

With the stale regex, the assertions in `requirements-config_test.yaml` drift on every release, and each version-bump PR fails `make test` on the outdated string (e.g. #1888). This updates the matchString to the new wording so the assertions track `Chart.yaml` again.

Verified: `renovate-config-validator` passes; regex captures the supported version on the proxy/hub assertion lines and ignores the hard-coded test inputs.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -92,7 +92,7 @@
       versioningTemplate: 'docker',
       matchStrings: [
         'traefik\\.io/(?<product>proxy|hub)-max-version:\\s*(?<currentValue>\\S+)',
-        'only supports Traefik (?<product>Proxy|Hub) up to (?<currentValue>v\\d+\\.\\d+\\.\\d+)',
+        'support Traefik (?<product>Proxy|Hub) \\S+, which is a major version above the supported (?<currentValue>v\\d+\\.\\d+\\.\\d+)',
         // Default Hub image assertion (container-config_test.yaml) — bumped with hub-max-version.
         'ghcr\\.io/traefik/traefik-(?<product>hub):(?<currentValue>v\\d+\\.\\d+\\.\\d+)',
       ],


.github/renovate.json5
```diff
@@ -92,7 +92,7 @@
       versioningTemplate: 'docker',
       matchStrings: [
         'traefik\\.io/(?<product>proxy|hub)-max-version:\\s*(?<currentValue>\\S+)',
-        'only supports Traefik (?<product>Proxy|Hub) up to (?<currentValue>v\\d+\\.\\d+\\.\\d+)',
+        'support Traefik (?<product>Proxy|Hub) \\S+, which is a major version above the supported (?<currentValue>v\\d+\\.\\d+\\.\\d+)',
```
bpsoraggi, line 95:
```suggestion
        'support Traefik (?<product>Proxy|Hub) \\S+, which is one major version above the supported (?<currentValue>v\\d+\\.\\d+\\.\\d+)',
```

Just a minor wording nitpick. Otherwise LGTM

darkweaver87 (approved):
LGTM, please run renovate after merge to check if it updates https://github.com/traefik/traefik-helm-chart/pull/1888

.github/renovate.json5
mloiseleur: The regex has to match the template's literal output (requirements.yaml:10 / :155), which says "a major version above". 

If I change only the regex to "one major version", it matches nothing and the manager goes dead again — the bug this PR fixes. Also "a major version above" = "a higher major version"; "one" would imply exactly one increment, but the guard fires for any higher major (e.g. v99.0.0).

PR #1891, opened 2026-06-15
fix(deployment): omit spec.replicas when replicas is null

### LINKED ISSUE #1864: Allow omitting Deployment.spec.replicas for externally managed scaling
### Welcome!

I would like to use/deploy Traefik with an external KEDA ScaledObject that manages the replica count of the Traefik Deployment. That allows me use additional metrics which are not available for default HPA.

In this setup, the Deployment’s spec.replicas field should be omitted from the rendered manifest so that KEDA becomes the sole owner of replica scaling.

Many GitOps environments use Argo CD together with KEDA.
Currently, the chart always renders spec.replicas when autoscaling.enabled=false.

When an external KEDA ScaledObject manages the Deployment:
1. KEDA updates Deployment.spec.replicas
2. Argo CD detects drift from the Helm-rendered value
3. Argo CD continuously reconciles the Deployment back to replicas: 1
4. KEDA scaling behavior is disrupted

There are could be several solutions to this:

1. When deployment.replicas is null, do not render the replicas field.
2. Introduce a dedicated flag `manageReplicas: false`
3. Introduce better integration with KEDA

mloiseleur:
## What does this PR do?

When `deployment.replicas` is set to `null` (and `autoscaling.enabled` is `false`), the rendered Deployment no longer includes a `spec.replicas` field.

## Motivation

Closes #1864.

The chart always rendered `spec.replicas` when `autoscaling.enabled: false`, defaulting to `1`. This prevents an external controller — such as KEDA, which brings its own `ScaledObject`/HPA — from owning the replica count: a GitOps tool like Argo CD detects drift and reconciles the Deployment back to the chart value, fighting the scaler.

Default behavior is unchanged (`replicas: 1`); only an explicit `null` drops the field.

## More

- Added unit test covering `replicas: null` → no `spec.replicas`.
- Documented the case in `EXAMPLES.md` (external scaler / KEDA).
- Regenerated `VALUES.md` and `values.schema.json`.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -85,6 +85,20 @@ autoscaling:
         averageUtilization: 80
 ```
 
+## Install with an external scaler (KEDA)
+
+When an external controller like [KEDA](https://keda.sh/) brings its own
+`ScaledObject` (and HPA), the chart's HPA stays disabled and `spec.replicas` must
+be omitted. Otherwise a GitOps tool like Argo CD reconciles the Deployment back to
+the chart value, fighting the scaler.
+
+```yaml
+deployment:
+  replicas: null
+autoscaling:
+  enabled: false
+```
+
 ## Install with Argo Rollouts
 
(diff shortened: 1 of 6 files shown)


darkweaver87 (approved):
LGTM

PR #1900, opened 2026-06-18
feat(hub): :sparkles: support hub.existingSecret for the license token

darkweaver87:
### What does this PR do?

Adds `hub.existingSecret` — an explicit way to reference an existing `Secret` (with a `token` key) that holds the Traefik Hub license — and makes `hub.token` unambiguously the *inline literal token*.

Today `hub.token` is overloaded by **length**: a value of 65+ chars is treated as a literal token (the chart creates the `traefik-hub-license` Secret), while a shorter value is silently treated as the *name* of an existing Secret — and the doc comment even says "Name of `Secret`", which is misleading.

- **`hub.existingSecret`** (new): name of an existing Secret to use (BYO). Takes precedence over `hub.token`; also enables Hub.
- **`hub.token`**: the inline literal token (chart creates `traefik-hub-license`).
- **Backward compatible**: a ≤ 64-char `hub.token` with no `existingSecret` still works as an existing-Secret name, now with a deprecation notice in `NOTES.txt` pointing to `hub.existingSecret`. Nothing breaks.

### Motivation

Removes a quiet footgun (length-based magic with a misleading comment) and aligns the BYO-license path with the upcoming experimental chart, where `hub.token` is always the literal token and BYO is explicit — so users can adopt the explicit form now and migrate cleanly later.

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -16,18 +16,45 @@ Create chart name and version as used by the chart label.
 
 {{/*
 Image defaults. An explicit image value always wins; otherwise the chart picks the
-Traefik Hub default when hub.token is set, and the Traefik Proxy default otherwise.
+Traefik Hub default when hub.token or hub.existingSecret are set, and the Traefik Proxy default otherwise.
 */}}
 {{- define "traefik.imageRegistry" -}}
-{{- .Values.image.registry | default (ternary "ghcr.io" "docker.io" (not (empty .Values.hub.token))) -}}
+{{- .Values.image.registry | default (ternary "ghcr.io" "docker.io" (eq (include "traefik-hub.enabled" .) "true")) -}}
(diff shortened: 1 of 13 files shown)


traefik/templates/_helpers.tpl
```diff
@@ -19,29 +19,56 @@ Image defaults. An explicit image value always wins; otherwise the chart picks t
 Traefik Hub default when hub.token is set, and the Traefik Proxy default otherwise.
```
mloiseleur:
```suggestion
Traefik Hub default when hub.token or hub.existingSecret are set, and the Traefik Proxy default otherwise.
```

mloiseleur (changes requested):
We should not invite or open users to put this kind of sensitive token in clear in _values_.

I'll open a PR to notify deprecation support of this hidden capacity.

Issue #1898, opened 2026-06-17
Why the enduser hate?


githubcdr:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik-helm-chart/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What version of the Traefik's Helm Chart are you using?

41.0.0

### What version of Traefik are you using?

latest

### What did you expect to happen ?

Hi,

Why would you introduce such human unfriendly breaking changes like https://github.com/traefik/traefik-helm-chart/pull/1887

defaultmode -> defaultMode, wow that's crazy

Just wanted to share my frustration as an enduser with this "freedom" to cause issues whenever you like

C

### What did you notice instead ?

Too much time wasted

### What are your values ?

I have to ask AI

### Additional Information

```shell
Announce breaking changes, provide a grace period and supply an automated script.
```

mloiseleur:
There is no enduser hate.

@githubcdr The motivation is in the PR:
> Improve user experience by reducing divergence between this chart and Traefik's own config.

We have observed that many users have been confused by the divergence of configuration between this chart and Traefik.
See for instance:
* https://github.com/traefik/traefik-helm-chart/issues/1602
* https://github.com/traefik/traefik-helm-chart/issues/1712
* https://github.com/traefik/traefik-helm-chart/issues/1860

That's why we have started this effort to align syntax between upstream and this chart, as much as possible.

It's a huge refactor so, to ease enduser life, we do it progressively. 
Last major version, v40, was focused on `Service` & `ingressNGINX` provider, for instance ([source](https://github.com/traefik/traefik-helm-chart/releases/tag/v40.0.0)).

> Announce breaking changes

AFAIK, the change has been announced:

* Following _semver_, it's a new major version
* The release notes display upgrade notes + bomb emoji on those changes:

<img width="1191" height="521" alt="Image" src="https://github.com/user-attachments/assets/8530f7ba-2cc8-4612-8e5c-190665748bd9" />

* same in Changelog.md

So, I wonder 🤔 , what else is missing ?

>  "provide a grace period"

Nothing forces a user to upgrade to this major version, so I'm not sure to follow.

Can you be more precise ? How could we provide such _grace period_ ?

> supply an automated script

That's a good idea, thanks 👍!

githubcdr:
Hi,

[...]

PR #1887, opened 2026-06-12
fix(logs)!: align syntax with upstream

mloiseleur:
## Motivation

Improve user experience by reducing divergence between this chart and Traefik's own config. 

Values now match the `log` and `accessLog` keys from the [Traefik logs reference](https://doc.traefik.io/traefik/observe/logs-and-access-logs/), so config can be copy-pasted directly.

### Before

```yaml
logs:
  general:
    level: DEBUG
    format: json
    filePath: "/var/log/traefik.log"
    noColor: true
    otlp:
      enabled: true
  access:
    enabled: true
    format: json
    timezone: "Europe/Paris"
    filters:
      statuscodes: "200,300-302"
      retryattempts: true
      minduration: "10ms"
    fields:
      general:
        defaultmode: keep
        names:
          ClientUsername: drop
      headers:
        defaultmode: drop
      queryParameters:
        defaultmode: drop
```

### After

```yaml
log:
  level: DEBUG
  format: json
  filePath: "/var/log/traefik.log"
  noColor: true
  otlp:
    enabled: true
accessLog:
  enabled: true
  format: json
  timezone: "Europe/Paris"
  filters:
    statusCodes: "200,300-302"
    retryAttempts: true
    minDuration: "10ms"
  fields:
    defaultMode: keep
    names:
      ClientUsername: drop
    headers:
      defaultMode: drop
    queryParameters:
      defaultMode: drop
```

### Key mapping

| Before | After |
|---|---|
| `logs.general.*` | `log.*` |
| `logs.access.*` | `accessLog.*` |
| `logs.access.filters.statuscodes` | `accessLog.filters.statusCodes` |
| `logs.access.filters.retryattempts` | `accessLog.filters.retryAttempts` |
| `logs.access.filters.minduration` | `accessLog.filters.minDuration` |
| `logs.access.fields.general.defaultmode` | `accessLog.fields.defaultMode` |
| `logs.access.fields.general.names` | `accessLog.fields.names` |
| `logs.access.fields.headers.defaultmode` | `accessLog.fields.headers.defaultMode` |
| `logs.access.fields.queryParameters.defaultmode` | `accessLog.fields.queryParameters.defaultMode` |

`accessLog.enabled` and `accessLog.timezone` are chart-only UX helpers (no native equivalent) and are kept.

## Testing

- `make test` → 688 passed
- `make lint` → success
- `make docs` + `make schema` regenerated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1901, opened 2026-06-18
docs(migrate): add v40 to v41 values migration helper

mloiseleur:
### What does this PR do?

Adds `hack/migrate/`: a cross-platform helper that rewrites a user values override
from v40 to v41 (the #1887 logging key renames). 

One `yq` expression holds the mapping; thin `.sh`/`.ps1` launchers run it via Docker. 

Dry-run by default, `-i` rewrites in place.
Warns (no auto-fix) on the `providers.file.content` type change. 

### Motivation

v41 renamed the logging keys (`logs.*` → `log.*` / `accessLog.*`); 
Enduser love. See #1898.

### More

Tested locally against `sample-v40-values.yaml` (covers every remapped key): `.sh` on
mikefarah yq v4 via Docker and `.ps1` on PowerShell 7 produce identical output; dry-run,
in-place + `.bak`, and idempotent re-run all verified.


darkweaver87:
Nice one @mloiseleur, this does the job for v41 👍 and it'll make a few users happy I think.

One thing before we commit to the approach though — wdyt about doing it in Go instead of yq + the shell/powershell wrappers? Since we're going to keep aligning the chart values with Traefik's config, we'll have more of these breaking renames down the line, and a small Go tool would let us add each one as a tiny tested transform instead of maintaining a new yq + `.sh` + `.ps1` trio every time.

Couple of other upsides: we can ship prebuilt binaries (no Docker, no yq, nothing to install) 📦, it runs offline (no Docker Hub pull / rate limits), and being a Linux team we can't really test the `.ps1` or macOS by hand 🐧 — whereas a Go binary is `go test`'d once and cross-compiles the same everywhere. It can also actually convert `providers.file.content` (string → object) instead of just warning about it ✨

I threw together a quick PoC to see how it feels — same mapping and sample as yours, comment-preserving and idempotent: https://github.com/darkweaver87/traefik-helm-chart/tree/feat/migrate-go-v40-v41/hack/migrate

Not trying to derail this PR 🙂 just think it might save us pain as the migrations pile up. Happy to turn it into a proper PR if you're into it!

mloiseleur:
Thanks @darkweaver87, solid PoC and I get the long-game reasoning. 🙏

Heads up though — I pushed a refactor after your comment that I think closes most of the gap:

- **No more script trio.** The `.sh` and `.ps1` launchers are gone. It's now a single declarative `v40-to-v41.yq` expression — nothing per-OS to hand-test on Windows/macOS, the same file runs identically everywhere.
- **`providers.file.content` is converted, not warned about** — `from_yaml` turns the string into a real object inline. ✨
- **Offline / no Docker Hub** — with a local `yq` there's no image pull and no rate limits; Docker is just the no-install fallback.
- **Tested** — there's a golden-fixture regression check (`sample-v40-values.yaml` → `expected-v41-values.yaml`).

For future migrations the maintenance unit would be "add another dedicated, tested `yq` file" (one per major version jump — minor and patch releases won't need one), which feels comparable to "add a tested Go transform" — but without shipping/hosting a per-platform binary or a Go toolchain + cross-compile release pipeline.

Where I think Go genuinely wins is if a migration ever needs real logic — conditionals, validation, value reshaping beyond key renames. `yq` is great at structural moves but gets awkward there. 

Proposal: keep this PR on `yq`, and if/when a migration outgrows it, revisit your Go tool then — it'd be the right call at that point.

If that doesn't sound fair to you, no worries — let's leave it open for other maintainers / community feedback before deciding. 🙂

hack/migrate/README.md
```diff
@@ -0,0 +1,64 @@
+# Migrate values from chart v40 to v41
```
jnoordsij: Maybe we should just name this "migrate values", add some 'general' explanation of the concept and requirements (yq or docker), and then add "v40 to v41" as first (and for now only) subheading. Then it already slightly prepares for adding more future migrations (or even historic ones if there's popular demand for it).

jnoordsij (approved):
I think this looks great! I agree with @mloiseleur that `yq` is probably the easier choice maintenance-wise for now, but great to know we have the option of more advanced tooling if the need arises. For now I'd say probably landing this as is so people can benefit from it ASAP is probably the way to go.

hack/migrate/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+All mapping logic lives in a single file, [`v40-to-v41.yq`](./v40-to-v41.yq), run with
+[mikefarah `yq` v4](https://github.com/mikefarah/yq). Run it against **your override file**
+(the one you pass with `helm -f`), not the chart's `values.yaml`.
+
+Back up first, since `-i` rewrites in place:
+
+```sh
+cp myvalues.yaml myvalues.yaml.bak
+```
```
darkweaver87: Generic intro + `## Requirements` (local `yq` or Docker), then `## v40 to v41` as the first subheading. Future migrations just append a new `## vNN to vNN+1` section.

````suggestion
# Migrate values

Some chart releases rename or restructure values. This directory holds one migration
per breaking release, each expressed as a single
[mikefarah `yq` v4](https://github.com/mikefarah/yq) expression. Run it against **your
override file** (the one you pass with `helm -f`), not the chart's `values.yaml`.

## Requirements

Either a local `yq` v4, or just Docker (`mikefarah/yq:4` image, nothing to install).

Back up first, since `-i` rewrites in place:

```sh
cp myvalues.yaml myvalues.yaml.bak
```

## v40 to v41

Chart **v41.0.0** renamed the logging keys to match upstream Traefik
([PR #1887](https://github.com/traefik/traefik-helm-chart/pull/1887)). The mapping lives
in [`v40-to-v41.yq`](./v40-to-v41.yq).

| Before (v40) | After (v41) |
|---|---|
| `logs.general.*` | `log.*` |
| `logs.access.*` | `accessLog.*` |
| `logs.access.filters.statuscodes` | `accessLog.filters.statusCodes` |
| `logs.access.filters.retryattempts` | `accessLog.filters.retryAttempts` |
| `logs.access.filters.minduration` | `accessLog.filters.minDuration` |
| `logs.access.fields.general.defaultmode` | `accessLog.fields.defaultMode` |
| `logs.access.fields.general.names` | `accessLog.fields.names` |
| `logs.access.fields.headers.defaultmode` | `accessLog.fields.headers.defaultMode` |
| `logs.access.fields.queryParameters.defaultmode` | `accessLog.fields.queryParameters.defaultMode` |

It also converts `providers.file.content` from a string to an object.

### Usage
````

**Review by darkweaver87 — COMMENTED** [review]
Following up on @jnoordsij's suggestion in https://github.com/traefik/traefik-helm-chart/pull/1901#discussion_r3549831926 — here are applyable suggestions restructuring the README into a generic "Migrate values" doc with `v40 to v41` as the first subheading.
@jnoordsij does it look good to you ? If so I will apply those and merge this PR.

File: hack/migrate/README.md [hunk 3]
```diff
@@ -0,0 +1,64 @@
...
+
+**With Docker** (no local install, only Docker required)
+
+```sh
+docker run --rm -v "$PWD":/w -w /w mikefarah/yq:4 \
+  --from-file /w/v40-to-v41.yq myvalues.yaml      # add -i after --from-file to rewrite in place
+```
+
+## Known limitations
```
darkweaver87:
```suggestion
### Known limitations
```

hack/migrate/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Comments are preserved by `yq`, but comments attached to keys that move
+  (`logs.*`) may shift. Review the diff before applying with `-i`.
+- The expression is idempotent: running it on an already-migrated file is a no-op.
+
+## Fixtures
```
darkweaver87: Kept under the v40→v41 section since the golden files are migration-specific — that way each future section stays self-contained.

```suggestion
### Fixtures
```

hack/migrate/README.md
darkweaver87: Good call — pushed a suggestion doing exactly that: generic intro + `## Requirements` (local `yq` or Docker), then `## v40 to v41` as the first subheading, with Usage/Limitations/Fixtures demoted under it. The next migration just appends a `## vNN to vNN+1` section.

@mloiseleur is away, so applying these three suggestions should unblock the merge.

jnoordsij:
> @jnoordsij does it look good to you ? If so I will apply those and merge this PR.

This looks great to me, thanks for the efforts!

darkweaver87 (approved):
LGTM

PR #1907, opened 2026-06-29
docs(hub): deprecate inline literal token in values

mloiseleur:
### Context

`hub.token` accepts two things by length: a value ≥65 chars is treated as the
inline literal license token (the chart creates the `traefik-hub-license`
Secret from it), while a shorter value is the name of an existing Secret.

The inline-literal path puts a sensitive token in `values.yaml` — it leaks into
release state, version control and CI logs. This was never the documented intent.

### Motivation

Deprecate the inline-literal path via a NOTES.txt warning and announce its
removal in a future release, steering users to bring their own Secret and set
`hub.token` to its name.

This is a counter-proposal to #1900, which goes the other way (promotes the
inline literal as primary, deprecates the existing-Secret-name path).

### DIFF
--- a/traefik/templates/NOTES.txt
+++ b/traefik/templates/NOTES.txt
@@ -15,6 +15,17 @@
 {{- end -}}
 
 
+{{/* Deprecation: inline literal hub.token in values (chart creates the license Secret). */}}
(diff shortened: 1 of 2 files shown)


darkweaver87 (approved):
LGTM

PR #1925, opened 2026-07-08
ci(artifacthub): structured changelog with link on major

mloiseleur:
### What does this PR do?

Emit the structured `artifacthub.io/changes` annotation from `hack/changelog.sh`:
- **Patch/minor** → per-kind entries (`feat`→added, `fix`→fixed, breaking or anything else→changed), gitmoji shortcodes stripped, release commit dropped.
- **Major (`X.0.0`)** → single `changed` entry linking to the release Upgrade Notes.

<details><summary>Generated output — last 5 versions</summary>

```yaml
# 41.1.1 (patch)
  artifacthub.io/changes: |
    - kind: fixed
      description: "fix: update hardened image registry"

# 41.1.0 (minor)
  artifacthub.io/changes: |
    - kind: fixed
      description: "fix: support enabling only kubernetesIngressNGINX with rbac.namespaced"
    - kind: fixed
      description: "fix(rbac): align namespaced RBAC with kubernetesIngressNGINX namespace scoping"
    - kind: fixed
      description: "fix(ports): support IPv6 hostIP values"
    - kind: fixed
      description: "fix(deployment): accept named port for healthchecksPort"
    - kind: added
      description: "feat(rbac): grant configmaps write for hub api management"
    - kind: added
      description: "feat(hub): add hardened image variant"
    - kind: added
      description: "feat(hub): add hub.enabled to run Traefik Hub in proxy mode"
    - kind: added
      description: "feat(deps): update traefik docker tag to v3.7.9"
    - kind: added
      description: "feat(deps): update traefik docker tag to v3.7.8"
    - kind: added
      description: "feat(deps): update traefik docker tag to v3.7.7"
    - kind: added
      description: "feat(deps): update ghcr.io/traefik/traefik-hub docker tag to v3.20.7"
    - kind: added
      description: "feat(CRDs): update Traefik Hub to v1.33.0"
    - kind: changed
      description: "docs: change metrics.prometheus labels types from string to bool"
    - kind: changed
      description: "docs(values): fix rendered types for nullable fields"
    - kind: changed
      description: "chore: file provider scrambled output"

# 41.0.2 (patch)
  artifacthub.io/changes: |
    - kind: fixed
      description: "fix(crds): add missing hub.traefik.io_uplinks to kustomization"
    - kind: added
      description: "feat(deps): update traefik docker tag to v3.7.6"
    - kind: added
      description: "feat(deps): update ghcr.io/traefik/traefik-hub docker tag to v3.20.6"

# 41.0.1 (patch)
  artifacthub.io/changes: |
    - kind: fixed
      description: "fix(ingressroute): fail fast on uppercase ingressRoute keys (RFC 1123)"
    - kind: added
      description: "feat(hub): support traefik hub v3.20.5"
    - kind: changed
      description: "docs(hub): deprecate inline literal token in values"

# 41.0.0 (major)
  artifacthub.io/changes: |
    - kind: changed
      description: "This is a major release with breaking changes. Read the upgrade notes before upgrading."
      links:
        - name: Upgrade Notes
          url: https://github.com/traefik/traefik-helm-chart/releases/tag/v41.0.0
```

</details>

Rewriting the annotation deletes only its own block, so an annotation sorting after `artifacthub.io/changes` is no longer dropped. An empty result aborts instead of writing a placeholder: it means `Chart.yaml` and `Changelog.md` are out of sync, and no release is changeless.

Covered by golden tests in `.github/fixtures/artifacthub-changes/`, run by `make test-changelog`, which the Test workflow calls next to `make test-ns`. Each case pins one behaviour: kind mapping, escaping and block replacement on a patch, the upgrade notes link on a major, a first release that is `X.0.0` without being breaking, and a release whose only commit is the release itself, which is refused. The tests also parse the emitted annotation and check every kind against the ones Artifact Hub accepts, so regenerating the golden files cannot bless a broken one.

Reading the chart version moved from `yq` to `sed`: `yq -r` is python-yq only and mikefarah yq has no `-r` flag, so the script could not run on a GitHub runner. The tests do use `yq`, in syntax both flavours run.

### Motivation

Artifact Hub buried the v41 breaking changes in a flat commit list (#1898). Artifact Hub has no "breaking" kind, so a major collapses to one entry pointing at the upgrade guide — impossible to miss.

### More

- [x] Yes, I ran `make test-changelog` and all the tests passed


darkweaver87:
Nice one — the breaking-change link on major is exactly what we were missing after #1898. Few things I'd tweak before it goes in:

The gitmoji stripping doesn't fully work — the regex only catches text shortcodes like `:rocket:`, but our changelog already has the rendered unicode emoji, so the output still shows `chore(release): 🚀 publish 41.0.2`. Since the description is plain text on Artifact Hub, might as well strip the unicode ones too.

Also, every release ends up listing its own `chore(release): publish X` commit as a `changed` badge — pure noise, we can filter it out.

Small hardening: `esc` only escapes `"`, so a stray backslash would break the YAML string. Won't happen with conventional commits realistically, but it's a one-liner to cover.

Last one, and the bigger one: right now everything that isn't feat/fix falls into `changed`, so we only use 2 of the 6 kinds. Could map a bit more:

- `feat` → added
- `fix` → fixed
- `perf` → changed
- `revert` → removed
- else → changed

`deprecated` and `security` don't have a conventional-commit type, so let's just skip those two.

darkweaver87:
Pushed 8c922f1 with all four points, ready for another pair of eyes while @mloiseleur is away.

- **Unicode gitmoji**: GNU sed can't match multibyte ranges, so the shortcode `sed` became one `perl -CSD` pass covering both forms. Perl ships by default where GNU sed doesn't, so no extra install.
- **Release commits**: filtered with `grep -v '^\* chore(release)'`.
- **Escaping**: backslash first, then quote.
- **Kinds**: `perf` → changed, `revert` → removed; `deprecated`/`security` have no conventional-commit type.

Re-ran against the real `Changelog.md` for 41.0.2 / 41.0.1 / 41.0.0 (matches the PR description minus the filtered release lines), plus a synthetic one for escaping and accented text. The emitted annotation round-trips through `yq`.

darkweaver87 (approved):
LGTM

darkweaver87:
@bpsoraggi could you have a look ?

bpsoraggi (approved):
LGTM

PR #1929, opened 2026-07-13
fix: support enabling only kubernetesIngressNGINX with rbac.namespaced

### LINKED ISSUE #1928: rbac.namespaced doesn't allow using kubernetesIngressNGINX provider only
### Welcome!

40.3.0

v3.7.4

Allow to use kubernetesIngressNGINX with rbac.namespaced

When only kubernetesIngressNGINX is enabled, rbac.namespaced cannot be used

> Error: execution error at (traefik/templates/requirements.yaml:30:8): ERROR: namespaced rbac requires Kubernetes CRD or Kubernetes Ingress provider.

Enabling `kubernetesIngress` alongside allowed the generation.

### What are your values ?

```yml
# hardening
rbac:
  namespaced: true

providers:
  # https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-crd/
  kubernetesCRD:
    # Support kubernetesIngressNGINX provider only to minimize supported configs
    # Cluster-wide resources are also prone to conflict
    enabled: false
  # https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-ingress/
  kubernetesIngress:
    # Support kubernetesIngressNGINX provider only to minimize supported configs
    enabled: false
  # Migrate from NGINX Ingress Controller (ingress-nginx) to Traefik with minimal configuration changes
  # https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-ingress-nginx/
  kubernetesIngressNGINX:
    enabled: true
    ingressClass: nginx
  # https://doc.traefik.io/traefik/reference/install-configuration/providers/others/file/
  file:
    enabled: true
    # Avoid unnecessary filesystem polling and eliminate risk of spurious reload
    watch: false
    # https://doc.traefik.io/traefik/reference/routing-configuration/http/tls/tls-certificates/#default-certificate
    # Chart template (provider-file-cm.yaml) has a rendering bug with toYaml + nindent
    # that produces a stray "|" line in the ConfigMap. Fixed by generate-templates.sh.
    content: |
      tls:
        stores:
          default:
            defaultCertificate:
              certFile: /default-tls-certificate/tls.crt
              keyFile: /default-tls-certificate/tls.key

```

### Additional Information

```shell

```

rtsui-harmonicinc:
### Motivation

Fixes #1928 

### More

### DIFF
--- a/traefik/templates/requirements.yaml
+++ b/traefik/templates/requirements.yaml
@@ -27,8 +27,8 @@
   {{- if .Values.providers.kubernetesGateway.enabled }}
     {{- fail "ERROR: Kubernetes Gateway provider requires ClusterRole. RBAC cannot be namespaced." }}
   {{- end }}
-  {{- if and (not .Values.providers.kubernetesIngress.enabled) (not .Values.providers.kubernetesCRD.enabled) }}
-    {{- fail "ERROR: namespaced rbac requires Kubernetes CRD or Kubernetes Ingress provider." }}
+  {{- if and (not .Values.providers.kubernetesIngress.enabled) (not .Values.providers.kubernetesCRD.enabled) (not .Values.providers.kubernetesIngressNGINX.enabled) }}
+    {{- fail "ERROR: namespaced rbac requires Kubernetes CRD, Kubernetes Ingress NGINX, or Kubernetes Ingress provider." }}
   {{- end }}
 {{- end }}
 
--- a/traefik/tests/rbac-config_test.yaml
+++ b/traefik/tests/rbac-config_test.yaml
@@ -95,6 +95,31 @@ tests:
           count: 0
         template: rbac/clusterrolebinding.yaml
 
+  - it: rbac.namespaced should not fail when only KubernetesIngressNGINX provider is enabled
+    set:
+      rbac:
+        namespaced: true
+      providers:
+        kubernetesIngressNGINX:
+          enabled: true
+        kubernetesIngress:
+          enabled: false
+        kubernetesCRD:
+          enabled: false
+    asserts:
+      - isKind:
+          of: Role
+        template: rbac/role.yaml
+      - isKind:
+          of: RoleBinding
+        template: rbac/rolebinding.yaml
+      - hasDocuments:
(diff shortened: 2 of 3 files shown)


darkweaver87:
Hi @rtsui-harmonicinc,

Thanks for fixing this! Would you mind adding a unit test that covers the positive case — rbac.namespaced=true with only kubernetesIngressNGINX enabled should now render successfully and produce the Role? The current test only checks the failure path.

Thanks

rtsui-harmonicinc:
hi @darkweaver87 ,

it seems to be related in L1035 of rbac-config_test.
I may edit that test to explicitly disable k8s crd provider as well

darkweaver87:
Hi @rtsui-harmonicinc,

Thanks for the follow-up! Unfortunately the test you added in `rbac-config_test.yaml` doesn't actually guard the fix: that suite is scoped to `templates: [rbac/*, deployment.yaml]`, so `requirements.yaml` is never rendered and the `fail` guard never fires. I checked locally — the new test passes identically against the *unfixed* `requirements.yaml`. It only re-asserts that the Role/RoleBinding render, which was already the case before your change.

The positive case belongs in `requirements-config_test.yaml` (that suite is scoped to `requirements.yaml`), right next to the existing negative test:

```yaml
  - it: should be possible to namespace rbac with only kubernetesIngressNGINX provider
    set:
      rbac:
        namespaced: true
      providers:
        kubernetesCRD:
          enabled: false
        kubernetesIngress:
          enabled: false
        kubernetesIngressNGINX:
          enabled: true
    asserts:
      - notFailedTemplate: {}
```

I verified this one fails without the fix and passes with it.

And yes, good catch on L1035 — that test leaves `kubernetesCRD` at its default `true`, so it doesn't isolate the NGINX-only case. Worth disabling it there as you suggested.

Thanks!

rtsui-harmonicinc:
thanks for the explanation, i have added the tests accordingly

darkweaver87 (approved):
LGTM

PR #1930, opened 2026-07-15
feat(rbac): grant configmaps write for hub api management

ArthurHlt:
### What does this PR do?

Grants `create`, `patch` and `delete` verbs on `configmaps` in the Traefik ClusterRole when Traefik Hub API management is enabled (`hub.apimanagement.enabled`). The permission is added through a dedicated rule so the default cluster role stays read-only and `nodes`/`services` keep only `get`/`list`/`watch`

### Motivation

Traefik Hub API management needs write access to configmaps to operate, but that access should not be granted by default nor extended to other core resources. Gating it behind `hub.apimanagement.enabled` keeps the cluster role least-privilege for every other configuration.

### More

### DIFF
--- a/traefik/templates/rbac/clusterrole.yaml
+++ b/traefik/templates/rbac/clusterrole.yaml
@@ -188,6 +188,14 @@ rules:
     verbs:
       - list
       - watch
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    verbs:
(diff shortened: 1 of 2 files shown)


mmatur (approved):

darkweaver87 (approved):
LGTM

PR #1933, opened 2026-07-16
fix(ports): support IPv6 hostIP values

### LINKED ISSUE #1788: IPv6 as hostIP is not quoted correctly & IPv6 is not braced when being passed over arguments
### Welcome!

39.0.6

v3.6.12

#### values.yaml
```
hostIP: "::"
```
#### generated deployment.yaml
```
hostIP: ::      # invalid yaml, missing quote
```

#### Workaround:
```
hostIP: "'::'"
```

#### Generated argument:
```
--entryPoints.web.address=fd::1:80/tcp     # invalid host
```

#### Workaround (doesn't work)
Setting hostIP as `[::]` solve the problem but k8s doesn't accept braces in the hostIP field

Generation error/Invalid CLI argument error

### What are your values ?

```
ports:
  web:
    port: 80
    hostPort: 80
    hostIP: "::"
    protocol: TCP
```

### Additional Information

```shell

```

ihopenre-eng:
### What does this PR do?

- Quotes `hostIP` in container ports and HTTP probe hosts so IPv6 values render as valid YAML.
- Brackets IPv6 `hostIP` values in Traefik entrypoint addresses while preserving IPv4, hostname, and empty-host rendering.
- Adds Deployment regression coverage for both regular ports and the `ports.traefik.hostIP` probe path.

This supersedes #1859 and incorporates its unresolved review feedback by covering IPv6 probe hosts and keeping Deployment assertions in the Deployment test suite.

### Motivation

Fixes #1788.

Without this change, `hostIP: ::` produces invalid YAML in container port and probe host fields, plus an unbracketed Traefik entrypoint address.

### More

- [x] Schema update is not required; no values or types changed
- [x] The full Helm unit suite passes (698 tests), and `helm lint traefik` passes

### DIFF
--- a/traefik/tests/ports-config_test.yaml
+++ b/traefik/tests/ports-config_test.yaml
@@ -187,7 +187,7 @@ tests:
           path: spec.template.spec.containers[0].args
           content: "--entryPoints.traefik.address=:8080/tcp"
 
-  - it: should quote IPv6 hostIP and bracket the entrypoint address
+  - it: should render IPv6 hostIP in ports and bracket the entrypoint address
     template: deployment.yaml
     set:
       ports:
@@ -205,7 +205,7 @@ tests:
           path: spec.template.spec.containers[0].args
           content: "--entryPoints.web.address=[::]:8000/tcp"
 
-  - it: should quote IPv6 hostIP when used as the probes host
+  - it: should render IPv6 hostIP as the probes host
     template: deployment.yaml
     set:
       ports:
--- a/traefik/tests/ports-config_test.yaml
+++ b/traefik/tests/ports-config_test.yaml
@@ -187,6 +187,41 @@ tests:
           path: spec.template.spec.containers[0].args
           content: "--entryPoints.traefik.address=:8080/tcp"
 
+  - it: should quote IPv6 hostIP and bracket the entrypoint address
+    template: deployment.yaml
+    set:
+      ports:
+        web:
+          hostIP: "::"
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].ports
+          content:
+            name: web
+            containerPort: 8000
+            hostIP: "::"
+            protocol: TCP
+      - contains:
+          path: spec.template.spec.containers[0].args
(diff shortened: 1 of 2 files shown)


traefik/tests/ports-config_test.yaml
```diff
@@ -187,6 +187,41 @@ tests:
           path: spec.template.spec.containers[0].args
           content: "--entryPoints.traefik.address=:8080/tcp"
 
+  - it: should quote IPv6 hostIP and bracket the entrypoint address
```
darkweaver87: I'd just rename the title to describe what the test verifies. The `equal: "::" `check passing is the proof the fix works (unquoted `host: ::` wouldn't parse to that value), so the test is correct; only the wording is off.

```suggestion
  - it: should render IPv6 hostIP in ports and bracket the entrypoint address
```

traefik/tests/ports-config_test.yaml
```diff
@@ -187,6 +187,41 @@ tests:
...
+          content:
+            name: web
+            containerPort: 8000
+            hostIP: "::"
+            protocol: TCP
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--entryPoints.web.address=[::]:8000/tcp"
+
+  - it: should quote IPv6 hostIP when used as the probes host
```
darkweaver87: Same here:
```suggestion
  - it: should render IPv6 hostIP as the probes host
```

darkweaver87 (commented):
LGTM, just few nit picks :-)

ihopenre-eng:
Thanks for the review @darkweaver87 :-) Applied both title suggestions in 33dfda5 — the `equal: "::"` assertions are unchanged, only the wording:

- `should quote IPv6 hostIP and bracket the entrypoint address` → `should render IPv6 hostIP in ports and bracket the entrypoint address`
- `should quote IPv6 hostIP when used as the probes host` → `should render IPv6 hostIP as the probes host`

darkweaver87 (approved):
LGTM

PR #1934, opened 2026-07-16
feat(hub): :sparkles: add hub.enabled to run Traefik Hub in proxy mode

darkweaver87:
### What does this PR do?

Adds `hub.enabled` to install the Traefik Hub distribution without a license token, running it in proxy mode: a drop-in Traefik Proxy.

```yaml
hub:
  enabled: true
```

This installs `ghcr.io/traefik/traefik-hub` with no `--hub.*` argument, no token volume/mount and no API portal. Setting `hub.token` later enables API Gateway on the same release, without changing the image. `hub.enabled` is nullable and defaults to `true` when `hub.token` is set, so existing values files render identically.

Choosing the distribution (image registry/repository/tag, marketplace images, Hub version bounds, Hub→Proxy version mapping) now keys off `hub.enabled`, while everything licensed (CLI args, RBAC, admission, API portal, modsec/uplink guards) still keys off `hub.token`. `hub.enabled: false` with a token set is rejected.

Also gates `hub-apiportal.yaml` on `hub.token`, which previously rendered a portal whenever `hub.apimanagement.enabled` was set, with no agent behind it.

### Motivation

Traefik Hub can now run without a license (traefik/traefik-hub#1245), giving a single deployment artifact that runs as a plain proxy today and unlocks commercial capabilities later by adding a license. The chart had no way to express this: `hub.token` being set was the only thing selecting the Hub distribution, so the Hub image without a token was unreachable except by hand-setting `image.repository`, which silently misresolved the version.

### More

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -145,7 +145,8 @@ Kubernetes: `>=1.25.0-0`
 | hub.apimanagement.admission.selfManagedCertificate | bool | `false` | By default, this chart handles directly the tls certificate required for the admission webhook. It's possible to disable this behavior and handle it outside of the chart. See EXAMPLES.md for more details. |
 | hub.apimanagement.enabled | bool | `false` | Set to true in order to enable API Management. Requires a valid license token. |
 | hub.apimanagement.openApi.validateRequestMethodAndPath | bool | `false` | When set to true, it will only accept paths and methods that are explicitly defined in its OpenAPI specification |
-| hub.hardened | bool | `false` | Use the hardened image variant. It appends `-hardened` to the tag and defaults the image to `registry.traefik.io/traefik-hub`. Requires a valid license token and Traefik Hub >= v3.21.0-ea. |
+| hub.enabled | bool | `true` when `hub.token` is set | Install Traefik Hub. Without `hub.token`, it runs in proxy mode: a drop-in Traefik Proxy, which requires Traefik Hub >= v3.21.0-ea. |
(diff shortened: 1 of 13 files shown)


ArthurHlt (approved):
lgtm and tested manually

traefik/VALUES.md
```diff
@@ -145,6 +145,7 @@ Kubernetes: `>=1.25.0-0`
 | hub.apimanagement.admission.selfManagedCertificate | bool | `false` | By default, this chart handles directly the tls certificate required for the admission webhook. It's possible to disable this behavior and handle it outside of the chart. See EXAMPLES.md for more details. |
 | hub.apimanagement.enabled | bool | `false` | Set to true in order to enable API Management. Requires a valid license token. |
 | hub.apimanagement.openApi.validateRequestMethodAndPath | bool | `false` | When set to true, it will only accept paths and methods that are explicitly defined in its OpenAPI specification |
+| hub.enabled | string | `true` when `hub.token` is set | Install Traefik Hub. Without `hub.token`, it runs in proxy mode: a drop-in Traefik Proxy, which requires Traefik Hub >= v3.21.0-ea. |
```
bpsoraggi: Should this not be boolean?

darkweaver87: Good catch — `hub.enabled` is nullable, so helm-docs inferred `string` from the empty YAML node. Fixed with the `(bool)` prefix convention from #1936; it now renders as `| hub.enabled | bool | `true` when `hub.token` is set | ... |`. Rebased on master and regenerated the docs and schema.

bpsoraggi (approved):
LGTM

mmatur (approved):
LGTM

PR #1935, opened 2026-07-16
docs: :books: change metrics.prometheus labels types from string to bool

AbhinavPraveen:
### What does this PR do?

Corrects the documentation to give some fields under `metrics.prometheus` the correct type.

### Motivation

Install failure when using string types:
```
Upgrade failed: values don't meet the specifications of the schema(s) in the
following chart(s):
traefik:
- at '/metrics/prometheus/addEntryPointsLabels': got string, want null or
boolean
- at '/metrics/prometheus/addRoutersLabels': got string, want null or boolean
- at '/metrics/prometheus/addServicesLabels': got string, want null or
boolean
```

### More

N/A, Docs only

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -322,9 +322,9 @@ Kubernetes: `>=1.25.0-0`
 | metrics.otlp.pushInterval | string | `""` | Interval at which metrics are sent to the OpenTelemetry Collector. Default: 10s |
 | metrics.otlp.resourceAttributes | object | `{}` | Defines additional resource attributes to be sent to the collector. |
 | metrics.otlp.serviceName | string | `nil` | Service name used in OTLP backend. Default: traefik. |
-| metrics.prometheus.addEntryPointsLabels | string | `nil` | Enable metrics on entry points. Default: true |
-| metrics.prometheus.addRoutersLabels | string | `nil` | Enable metrics on routers. Default: false |
-| metrics.prometheus.addServicesLabels | string | `nil` | Enable metrics on services. Default: true |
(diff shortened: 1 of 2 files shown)


darkweaver87:
Thanks for catching this — the rendered type is genuinely misleading and your fix was the right call, just in the wrong file.

`traefik/VALUES.md` is generated from `traefik/values.yaml` by helm-docs (`make docs`), so a hand-edit there gets overwritten on the next regeneration. The actual fix is a helm-docs type override in `values.yaml`: helm-docs infers the type from the YAML node, and since these fields are empty it falls back to `string` — it does not read the `@schema` annotation. Prefixing the doc comment with `(bool)` overrides it.

I took the liberty of pushing that to your branch (2949b24) — the regenerated `VALUES.md` came out byte-identical to what you had written by hand, so your PR output is unchanged, it is just reproducible now. Let me know if you would rather revert it and do it yourself.

One note on the original error: the schema was already correct (`type:[boolean, null]`), so the install failure came from passing a string, not from a schema bug. The misleading doc is exactly what caused that, so this is still worth fixing.

This is not limited to these three fields — ~79 other rows in `VALUES.md` have the same problem (`api.debug`, `api.insecure`, `autoscaling.minReplicas`, all of `metrics.otlp.*`…). I will open a follow-up PR for the sweep once this one is merged, and credit this PR as the origin.

AbhinavPraveen:
I'm happy for you to merge this as is or for you to push to my branch if needed @darkweaver87. I appreciate the offer to credit this PR though, thanks.

darkweaver87 (approved):
LGTM

PR #1936, opened 2026-07-17
docs(values): :books: fix rendered types for nullable fields

darkweaver87:
### What does this PR do?

Follow-up to #1935, applied to the rest of the chart.

- **Nullable fields retyped**: helm-docs infers the type from the YAML node, so intentionally-empty fields fell back to `string`. A `(bool)` / `(int)` prefix on the doc comment overrides it.
- **Descriptions recovered**: some comments used `## --` instead of `# --`, so helm-docs skipped them and the fields rendered with an empty description. For fields whose default is computed by the templates (`readinessPath`, `livenessPath`, `healthchecksHost`…), the `Default:` clause is expressed with `# @default --` so the rendered Default column matches, instead of contradicting it.
- **`ports.websecure.transport`** had no `# --` line, unlike `ports.web.transport`: helm-docs recursed and emitted a row per child instead of one documented row.

### Motivation

#1935 hit an install failure because the docs said `string`, so the value was quoted and the schema rejected it. The same trap remained on many other fields, including `api.insecure` and `autoscaling.minReplicas`.

### More

Comments only, no behavior change. `values.schema.json` gains the descriptions recovered from the `## --` blocks; no type or constraint touched.

`ports.websecure.http3.enabled` is the only description I reworded (raw comment lines → markdown links). `deployment.healthchecksPort` is deliberately left untouched here — it needs a schema change (named ports), handled in a separate PR.

Follow-up: some fields have no doc comment at all and still render with an empty description.

- [x] Yes, I updated the tests accordingly _(comments only)_


traefik/VALUES.md
```diff
@@ -513,24 +507,24 @@ Kubernetes: `>=1.25.0-0`
 | providers.kubernetesIngressNGINX.modsec.enabled | bool | `false` | Enable ModSec engine. Requires Traefik Hub >= v3.20.0-ea.8. |
 | providers.kubernetesIngressNGINX.modsec.owaspCoreRules | bool | `false` | Enable OWASP Core Rules. |
 | providers.kubernetesIngressNGINX.modsec.snippet | string | `""` | Custom ModSec rules snippet. |
-| providers.kubernetesIngressNGINX.proxyBodySize | string | `nil` | Default maximum size of a client request body in bytes (default: 1048576) |
-| providers.kubernetesIngressNGINX.proxyBufferSize | string | `nil` | Default buffer size for reading the response body in bytes (default: 8192) |
+| providers.kubernetesIngressNGINX.proxyBodySize | int | `nil` | Default maximum size of a client request body in bytes (default: 1048576) |
+| providers.kubernetesIngressNGINX.proxyBufferSize | int | `nil` | Default buffer size for reading the response body in bytes (default: 8192) |
 | providers.kubernetesIngressNGINX.proxyBuffering | string | `nil` | Defines whether to enable response buffering (default: false) |
```
bpsoraggi: Shouldn't this be bool (`providers.kubernetesIngressNGINX.proxyBuffering`)?

I've caught some other examples as well, but not sure what's supposed to be left out and what was just missed on accident

darkweaver87: Good catch — accidental miss, nothing deliberately left out: the sweep only caught keys with an empty value, not the ones written as explicit `foo: null`.

Fixed in 67260d5, along with every other nullable key still rendering as `string`:

* `bool`: `proxyRequestBuffering`, `proxyBuffering`, `allowCrossNamespaceResources`, `allowSnippetAnnotations`, `strictValidatePathType`, `ports.web.uplink`, `hub.sendlogs`
* `int`: `ports.websecure.containerPort`, `ports.websecure.hostPort`, `ports.websecure.http3.advertisedPort`

The remaining `string` rows are collapsed children of `transport` / `encodedCharacters` / `podDisruptionBudget` under a parent `@default -- nil`.

bpsoraggi (commented):
I think you might have still missed a few, let me know if I'm wrong:
- `accessLog.fields.queryParameters.defaultMode` (bool?) - L38
- `hub.redis.database` (int?) - L230

darkweaver87:
Thanks — I did another full sweep. Results:

- `hub.redis.database`: good catch, it's a Redis DB index (integer). Fixed → now `[integer, null]`, rendering as `int` (and its test fixture updated).
- `accessLog.fields.queryParameters.defaultMode`: this one's intentional — it's a string enum (`keep`/`drop`/`null`), matching its sibling `defaultMode` fields under `fields` and `fields.headers`, so `string` is correct rather than `bool`.

While sweeping I also found the real culprit behind a couple of look-alikes: fields whose `# @schema type:` sat on a **separate line above** the `# --` comment (instead of inline on the value) were still rendered as `string` by helm-docs, even though `values.schema.json` typed them correctly. Fixed the rendering for:

- `deployment.hostUsers` → `bool`
- `hub.providers.multicluster.children.cluster-1.serversTransport.{disableHTTP2,insecureSkipVerify}` → `bool`, `maxIdleConnsPerHost` → `int`

The remaining `string | nil` rows are all genuine strings (formats, enums, names, registries, tracing verbosity, duration timeouts). Also rebased on `master`.

bpsoraggi (approved):
LGTM

PR #1937, opened 2026-07-20
fix(deployment): :bug: accept named port for healthchecksPort

darkweaver87:
### What does this PR do?

Allows `deployment.healthchecksPort` to be a named port (e.g. `web`), not only a numeric one.

### Motivation

The value is rendered into the liveness/readiness probes' `httpGet.port`, which is an `IntOrString` in Kubernetes — a named container port is valid there, and the chart's container ports are named. The schema restricted it to `integer`, rejecting a valid configuration. This widens it to `[string, integer, null]`, matching the existing `targetPort` field which feeds the same kind of `IntOrString`.

### More

Adds a unit test rendering a named `healthchecksPort` into both probes. Note that `helm-unittest` does not validate against `values.schema.json` (only `helm install` / `helm lint` do), so the test guards the rendering rather than the schema relaxation itself.


bpsoraggi (approved):
LGTM, reviewed changes locally

PR #1938, opened 2026-07-20
fix(rbac): :bug: align namespaced RBAC with kubernetesIngressNGINX namespace scoping

darkweaver87:
### What does this PR do?

Follow-up to #1929. Fixes two gaps in `rbac.namespaced` handling for the Kubernetes Ingress NGINX provider.

**1. `watchNamespace` was not taken into account when generating Roles.** `role.yaml`/`rolebinding.yaml` build their namespace list from `kubernetesCRD.namespaces`, `kubernetesIngress.namespaces`, `knative.namespaces` and `hub.namespaces` only. With `rbac.namespaced=true` and `providers.kubernetesIngressNGINX.watchNamespace=foo`, the pod is started with `--providers.kubernetesingressnginx.watchnamespace=foo` but no Role is created in `foo`, so Traefik gets 403s on every Ingress it is supposed to watch. The namespace list now mirrors the CLI argument, and the `ingresses`/`ingresses/status` rules are scoped to that namespace instead of being granted everywhere.

**2. `watchNamespaceSelector` is incompatible with namespaced RBAC.** It requires cluster-wide `namespaces` list/watch, which a `Role` cannot grant. Unlike the `ingressclasses` lookup — which Traefik degrades gracefully on a `Forbidden` error — this path hard-fails at startup (`listing namespaces: %w`). The combination is now rejected at template time with an explicit message, and the constraint is documented on the value.

### Motivation

Raised while reviewing #1929: relaxing the namespaced-RBAC guard for NGINX-only setups makes both of these reachable. This PR is based on `master` and is independent of #1929, so the two can merge in either order.

### More


ArthurHlt (approved):
lgtm

youkoulayley (approved):

PR #1942, opened 2026-07-22
feat(hub): add hardened image variant

jspdown:
### What does this PR do?

Add a `hub.hardened` option to deploy the hardened image variant. When `hub.hardened.enabled` is `true`, the image is built as:

    <hub.hardened.registry>/<hub.hardened.repository>:<image.tag>-hardened

with the registry and repository defaulting to `registry.traefik.io` and `traefik-hub`. A FIPS tag (e.g. `v3.20.0-fips`) becomes `v3.20.0-fips-hardened`. In this mode `image.registry` and `image.repository` are not used; only `image.tag` provides the version.

### Motivation

Allow deploying the hardened and FIPS-hardened images.

### More

### DIFF
--- a/traefik/templates/_helpers.tpl
+++ b/traefik/templates/_helpers.tpl
@@ -34,7 +34,9 @@ Traefik Hub default when hub.token is set, and the Traefik Proxy default otherwi
 Create the chart image name.
 */}}
 {{- define "traefik.image-name" -}}
-{{- if .Values.oci_meta.enabled -}}
+{{- if .Values.hub.hardened.enabled -}}
+{{- printf "%s/%s:%s-hardened" .Values.hub.hardened.registry .Values.hub.hardened.repository (.Values.image.tag | default .Chart.AppVersion) }}
+{{- else if .Values.oci_meta.enabled -}}
  {{- if .Values.hub.token -}}
 {{- printf "%s/%s:%s" .Values.oci_meta.repo .Values.oci_meta.images.hub.image .Values.oci_meta.images.hub.tag }}
  {{- else -}}
@@ -370,7 +372,7 @@ Hash: {{ sha1sum ($cert.Cert | b64enc) }}
     {{- $found -}}
 {{- end -}}
 
-{{/* 
+{{/*
 Validate localPlugin configuration and determine plugin type
 Returns: hostPath, inline, or localPath
 */}}
@@ -393,7 +395,7 @@ Returns: hostPath, inline, or localPath
     {{- end -}}
 {{- end -}}
 
-{{/* 
+{{/*
 Get hostPath for a plugin (handles both old and new structure)
 */}}
 {{- define "traefik.getLocalPluginHostPath" -}}
@@ -407,7 +409,7 @@ Get hostPath for a plugin (handles both old and new structure)
     {{- end -}}
 {{- end -}}
 
-{{/* 
+{{/*
 Get inline plugin files (new structure only)
 */}}
 {{- define "traefik.getLocalPluginInlineFiles" -}}
@@ -417,7 +419,7 @@ Get inline plugin files (new structure only)
     {{- end -}}
 {{- end -}}
 
-{{/* 
+{{/*
 Get localPath plugin configuration (new structure only)
 */}}
 {{- define "traefik.getLocalPluginLocalPath" -}}
@@ -433,7 +435,7 @@ Get localPath plugin configuration (new structure only)
     {{- end -}}
 {{- end -}}
 
-{{/* 
+{{/*
 Check if a volume name exists in additionalVolumes
 */}}
 {{- define "traefik.volumeExistsInAdditionalVolumes" -}}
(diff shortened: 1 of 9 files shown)


darkweaver87 (dismissed):
_(Superseded — see the review below.)_

traefik/templates/_helpers.tpl
```diff
@@ -18,7 +18,9 @@ Create chart name and version as used by the chart label.
 Create the chart image name.
 */}}
 {{- define "traefik.image-name" -}}
-{{- if .Values.oci_meta.enabled -}}
+{{- if .Values.hub.hardened.enabled -}}
+{{- printf "%s/%s:%s-hardened" .Values.hub.hardened.registry .Values.hub.hardened.repository (.Values.image.tag | default .Chart.AppVersion) }}
```
darkweaver87: Two issues after rebase: `.Chart.AppVersion` is the Proxy version while the repository here is `traefik-hub`, so this renders `:v3.7.8-hardened` where the baseline gives `:v3.20.6`; and `image.digest` is silently ignored, dropping digest pinning. Folding into `traefik.imageRegistry` / `imageRepository` + the tag branch fixes both, leaving the digest branch untouched.

traefik/templates/_helpers.tpl
```diff
@@ -18,7 +18,9 @@ Create chart name and version as used by the chart label.
 Create the chart image name.
 */}}
 {{- define "traefik.image-name" -}}
-{{- if .Values.oci_meta.enabled -}}
+{{- if .Values.hub.hardened.enabled -}}
```
darkweaver87: First position means this silently overrides `oci_meta.enabled` and `global.azure.enabled`. Intentional? A `fail` would be clearer.

traefik/values.yaml
```diff
@@ -1205,6 +1205,13 @@ hub:
   # -- Name of `Secret` with key 'token' set to a valid license token.
   # It enables API Gateway.
   token: ""
+  hardened:
```
darkweaver87: With the boolean approach this collapses to one value:

```yaml
  # -- Use the hardened image variant. Appends the `-hardened` suffix to the image tag
  # and defaults the registry to `registry.traefik.io`. Requires the Traefik Hub distribution.
  hardened: false
```

and `image.registry` / `image.repository` keep working as overrides.

traefik/tests/hub-hardened_test.yaml
```diff
@@ -0,0 +1,75 @@
...
+      hub:
+        hardened:
+          enabled: true
+          repository: acme/traefik-hub
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[0].image
+          value: registry.traefik.io/acme/traefik-hub:v3.3.0-hardened
+
+  - it: should ignore image registry and repository when hardened
```
darkweaver87: This pins down the behaviour I'd like changed: someone mirroring the hardened image internally sets `image.registry` and gets no error and no effect. Either honour it or `fail`.

After the rebase, please also cover: hardened + `hub.token` with no `image.tag` (asserting the Hub tag default), hardened + `image.digest`, and hardened on a non-Hub distribution (asserting the new `fail`).

darkweaver87 (commented):
Thanks! Needs a rebase — the branch is ~147 commits behind `master`, and `584f20f` / `eea9619` reworked image defaulting. Replayed on current `master`:

```
# hardened + hub.token          -> registry.traefik.io/traefik-hub:v3.7.8-hardened
# same values, no hardened      -> ghcr.io/traefik/traefik-hub:v3.20.6
```

The hardened path pulls the `traefik-hub` repository but tags it with `.Chart.AppVersion`, the *Proxy* version; `master` resolves the Hub tag from the `traefik.io/hub-max-version` annotation (`traefik.defaultTag`). So it asks for a tag that doesn't exist. Blocker.

**Suggested design:** a plain `hub.hardened: false` boolean feeding the existing helpers, instead of a map with its own registry/repository:

```gotemplate
{{- define "traefik.imageRegistry" -}}
{{- if .Values.image.registry -}}{{ .Values.image.registry }}
{{- else if .Values.hub.hardened -}}registry.traefik.io
{{- else -}}{{ ternary "ghcr.io" "docker.io" (not (empty .Values.hub.token)) }}
{{- end -}}
{{- end -}}
```

Same shape for the repository, `-hardened` appended in the tag branch of `traefik.image-name`. You inherit the Hub tag default, `image.digest` keeps working, and `image.registry` / `image.repository` stop being silently discarded.

Plus: a guard that `hub.hardened` isn't set on a non-Hub distribution, and a `NOTES.txt` hint when it's set with empty `deployment.imagePullSecrets` — a warning, not a `fail`, since a user-supplied `serviceAccount.name` may carry pull secrets the chart can't see.

**Note on #1934:** it makes `hub.enabled`, not `hub.token`, the distribution selector, so the guard should key off that — a tokenless proxy-mode install is a legitimate consumer of the hardened image. It also touches `_helpers.tpl`, `requirements.yaml` and `NOTES.txt`; worth agreeing on a merge order.

Docs and schema are clean, and the suite passes on `master` — the failure is a wrong-but-valid tag, not a render error, hence the test ask below.

darkweaver87 (approved):
LGTM

bpsoraggi (approved):
LGTM

PR #1946, opened 2026-07-28
feat(hub): :sparkles: support AWS EC2 provider

zalbiraw:
### What does this PR do?

Adds `hub.providers.ec2` so the Traefik Hub EC2 tag-based discovery provider can be configured through chart values, matching the pattern used by the other Hub providers. Modelled on #1808 (Nutanix Prism Central), including the indexed-list handling for `filters` and the allow-empty `securityGroupPortDiscovery` struct. Gated on Traefik Hub `>= v3.21.0` in `requirements.yaml`.

### Motivation

Traefik Hub gained [an EC2 tag-based discovery provider](https://doc.traefik.io/traefik-hub/api-gateway/reference/install/providers/ref-provider-ec2), but the chart had no mapping for it. Today `hub.providers.ec2` values pass schema validation — because `hub.providers` is the one permissive node in an otherwise `additionalProperties: false` chain — and are then silently never read: a user following the documented Helm example installs a gateway with no EC2 provider, no error and no warning. The interim workaround is `additionalArguments`.

### More

### DIFF
--- a/traefik/tests/hub-ec2-config_test.yaml
+++ b/traefik/tests/hub-ec2-config_test.yaml
@@ -1,0 +1,119 @@
+suite: Traefik Hub EC2 provider
+templates:
+  - deployment.yaml
+set:
+  image:
+    registry: ghcr.io
+    repository: traefik/traefik-hub
+    tag: v3.21.0
+  hub:
+    token: "xxx"
+tests:
+  - it: should not configure ec2 by default
+    asserts:
+      - notContains:
+          path: spec.template.spec.containers[0].args
+          content: "--hub.providers.ec2.region="
+
+  - it: should configure ec2 when enabled
+    set:
+      hub:
+        providers:
+          ec2:
+            enabled: true
+            region: "us-east-1"
+            accessKeyID: "AKIAEXAMPLE"
+            secretAccessKey: "secret"
+            exposedByDefault: false
+            refreshSeconds: 30
+            defaultRule: "Host(`{{ normalize .Name }}`)"
+            constraints: "Tag(`env=prod`)"
+            ipMode: "public"
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--hub.providers.ec2.region=us-east-1"
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--hub.providers.ec2.accessKeyID=AKIAEXAMPLE"
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--hub.providers.ec2.secretAccessKey=secret"
+      - contains:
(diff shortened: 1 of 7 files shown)


darkweaver87 (approved):
LGTM

traefik/tests/requirements-config_test.yaml
```diff
@@ -735,6 +735,56 @@ tests:
             enabled: true
     asserts:
       - notFailedTemplate: {}
+  - it: should fail when an ec2 filter is missing a name
```
mloiseleur, line 782: The requirements on version check is untested. 

Would you please add a test for it? Something like:

```yaml     
    - it: should fail when using ec2 provider on Traefik Hub < v3.21.0
      set: {image: {registry: ghcr.io, repository: traefik/traefik-hub, tag: v3.20.0}, hub: {token: "xxx", providers: {ec2: {enabled: true}}}}
      asserts:
        - failedTemplate:
            errorMessage: "ERROR: hub.providers.ec2 is only available for Traefik Hub >= v3.21.0."
```

traefik/tests/hub-ec2-config_test.yaml
```diff
@@ -0,0 +1,119 @@
...
+    repository: traefik/traefik-hub
+    tag: v3.21.0
+  hub:
+    token: "xxx"
+tests:
+  - it: should not configure ec2 by default
+    asserts:
+      - notContains:
+          path: spec.template.spec.containers[0].args
+          content: "--hub.providers.ec2.region="
```
mloiseleur, line 19: Reviewing this test, I found a limitation.
It should be possible to enable this provider without any params, using AWS env vars or similar.
In the current state of this PR, it does not seems to be possible.

```console
$ helm template traefik --set image.tag=v3.21.0-ea.1 --set hub.enabled=true --set hub.providers.ec2.enabled=true ./traefik/ | grep ec2

```

It was also the case for other hub providers, so I've opened https://github.com/traefik/traefik-helm-chart/pull/1956 to fix this. Would you please update this PR accordingly?

zalbiraw: fixed! thank you

traefik/tests/requirements-config_test.yaml
zalbiraw: fixed!

zalbiraw (commented):
done!

mloiseleur (approved):

mloiseleur:
@zalbiraw I pushed a small commit in this PR (f623c1c) to improve further userXP. 
Please review it and tell me if I missed something :pray: .
I guess we can merge this PR afterwards.

zalbiraw:
Great additions, looks great on my end thank you.

PR #1948, opened 2026-07-30
docs(hub-manager): 📚 rewrite EXAMPLES into a README install guide

darkweaver87:
### What does this PR do?

Renames `hub-manager/EXAMPLES.md` to `README.md` and rewrites it as an install guide mirroring the main chart README: prerequisites, Secret, `hub-manager` install (repo + OCI), then wiring the `traefik` chart with an offline license. Both charts install into a shared `hub` namespace, so the license Secret is reused and the platform URL is concrete. Also aligns the main README Kubernetes prerequisite (v1.22.0 → v1.25.0) with `kubeVersion` and updates the markdown-lint target.

### Motivation

Before releasing the `hub-manager` chart, its `EXAMPLES.md` was not shown when browsing the chart directory and only listed examples rather than guiding a full install alongside Traefik.

### More

_Documentation-only change — no tests or schema affected._


youkoulayley (approved):

bpsoraggi (approved):
LGTM

We can re-include an `EXAMPLES.md` in the future, if we see need for it

PR #1955, opened 2026-08-03
fix(rbac): allow namespaced rbac with only the knative provider

karpovantonme:
### Problem

Enabling the knative provider together with `rbac.namespaced` fails the chart, even though `role.yaml` is written to handle exactly that combination.

```yaml
experimental:
  knative: true
rbac:
  namespaced: true
providers:
  kubernetesCRD: {enabled: false}
  kubernetesIngress: {enabled: false}
  kubernetesIngressNGINX: {enabled: false}
  knative: {enabled: true}
```

```
$ helm template traefik ./traefik -f values.yaml
Error: execution error at (traefik/templates/requirements.yaml:38:8):
ERROR: namespaced rbac requires Kubernetes CRD, Kubernetes Ingress NGINX, or Kubernetes Ingress provider.
```

The guard predates the provider. `git log -S` puts it at 4f39c87 (2025-06-18, `feat(hub): initial support for AI Gateway`), while knative landed in fbd7922 (2025-11-10, `feat: knative provider`) five months later. That commit taught `role.yaml` to emit knative rules under `rbac.namespaced`:

```
{{- if (and (has . $knativeNamespaces) $.Values.providers.knative.enabled) }}
  - apiGroups:
      - networking.internal.knative.dev
```

but the guard was not updated, so the chart never reaches that branch.

Nothing about knative needs cluster scope. Its rules are `networking.internal.knative.dev` `ingresses` and `ingresses/status`, byte-for-byte the same in `role.yaml` and `clusterrole.yaml`. Compare with `kubernetesGateway`, which does need cluster scope and has its own explicit guard two lines above.

### Change

Add `knative` to the guard and to the message it prints.

### Test

Added `should be possible to namespace rbac with only knative provider` to `tests/requirements-config_test.yaml`, next to the existing cases for the other three providers.

Written first, per `AGENTS.md`. It fails on `master` with the error above and passes with the change.

I also updated the expected `errorMessage` in `should not be possible to namespace rbac without one of the required providers`, since the wording now mentions knative.

### Checks

```
$ helm unittest traefik
Test Suites: 50 passed, 50 total
Tests:       734 passed, 734 total
```

And the render that used to fail now emits the Role with the knative rules in it.

Ran with helm 4.2.3 and the helm-unittest plugin directly rather than through `make test`, since Docker was not up on this machine. Same suites either way, but saying which one I actually ran.

### One thing I noticed but did not touch

The `failedTemplate` assertions in that suite have `errorMessage` at the same indent level as `failedTemplate:` itself:

```yaml
    asserts:
      - failedTemplate:
        errorMessage: "ERROR: ..."
```

which parses as two sibling keys, so `failedTemplate` gets an empty body and the message is never compared. Changing the wording of any guarded error currently cannot fail these tests. Happy to fix the indentation across the suite in a separate PR if you want it, but it looked like a change worth keeping out of this one.

### DIFF
--- a/traefik/tests/requirements-config_test.yaml
+++ b/traefik/tests/requirements-config_test.yaml
@@ -372,7 +372,24 @@ tests:
           enabled: false
     asserts:
       - failedTemplate:
-        errorMessage: "ERROR: namespaced rbac requires Kubernetes CRD, Kubernetes Ingress NGINX, or Kubernetes Ingress provider."
+        errorMessage: "ERROR: namespaced rbac requires Kubernetes CRD, Kubernetes Ingress NGINX, Kubernetes Ingress, or Knative provider."
+  - it: should be possible to namespace rbac with only knative provider
+    set:
+      experimental:
+        knative: true
+      rbac:
+        namespaced: true
+      providers:
+        kubernetesCRD:
+          enabled: false
+        kubernetesIngress:
+          enabled: false
+        kubernetesIngressNGINX:
+          enabled: false
+        knative:
+          enabled: true
+    asserts:
+      - notFailedTemplate: {}
   - it: should be possible to namespace rbac with only kubernetesIngressNGINX provider
     set:
       rbac:
--- a/traefik/tests/requirements-config_test.yaml
+++ b/traefik/tests/requirements-config_test.yaml
@@ -372,7 +372,7 @@ tests:
           enabled: false
     asserts:
       - failedTemplate:
-        errorMessage: "ERROR: namespaced rbac requires Kubernetes CRD, Kubernetes Ingress NGINX, Kubernetes Ingress, or Knative provider."
+          errorMessage: "ERROR: namespaced rbac requires Kubernetes CRD, Kubernetes Ingress NGINX, Kubernetes Ingress, or Knative provider."
   - it: should be possible to namespace rbac with only knative provider
     set:
       experimental:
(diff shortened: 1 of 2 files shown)


traefik/tests/requirements-config_test.yaml
```diff
@@ -372,7 +372,24 @@ tests:
           enabled: false
     asserts:
       - failedTemplate:
-        errorMessage: "ERROR: namespaced rbac requires Kubernetes CRD, Kubernetes Ingress NGINX, or Kubernetes Ingress provider."
+        errorMessage: "ERROR: namespaced rbac requires Kubernetes CRD, Kubernetes Ingress NGINX, Kubernetes Ingress, or Knative provider."
```
```suggestion
          errorMessage: "ERROR: namespaced rbac requires Kubernetes CRD, Kubernetes Ingress NGINX, Kubernetes Ingress, or Knative provider."
```
Indentation is required here.

mloiseleur (commented):
Thanks. LGTM. 
Please fix the indentation and we should be good to go.

mloiseleur (approved):

karpovantonme:
Fixed.

Turns out that indentation was doing real work. Checked it both ways: with the fix, swapping the expected text makes the suite fail (1 failed, 733 passed). Without it the same swap still passes, because YAML reads `errorMessage` as a sibling of `failedTemplate`, so only the failure gets asserted and the message itself never gets compared.

The rest of the `failedTemplate` assertions in that file have the same shape, so those messages are sitting there uncompared too. Kept this PR to one thing though. Can send the rest separately if you want them.

PR #1947, opened 2026-07-29
chore(release): 🚀 publish 41.1.0 and hub-manager 1.0.0

darkweaver87:
### What does this PR do?

Release a new version of this Chart.

### Motivation

Support of Traefik v3.7.9, Traefik-Hub v3.20.7 & v3.21.0-ea.1 and Traefik-Hub offline

PR #1961, opened 2026-08-05
ci(renovate): scope updates per chart on VALUES.md & AppVersion

mloiseleur:
## Context

The `VALUES.md` custom manager uses an unanchored regex file pattern, so it also matches `hub-manager/VALUES.md` and writes the Traefik Proxy version into its badge. And `traefik/Chart.yaml` lost its `# renovate: image=traefik` hint in #1947, so the `helmChartYamlAppVersions` preset stopped updating `appVersion`.

## Motivation

Both show up in #1954: `hub-manager` advertises `AppVersion: v3.7.10`, and `traefik/Chart.yaml` keeps `appVersion: v3.7.9` while its `VALUES.md` badge moves on.

This anchors the pattern to the traefik chart, restores the hint, and adds a dedicated manager so `hub-manager` tracks stable `ghcr.io/traefik/hub-manager` releases. A `make check-renovate` step now fails the build if the hint disappears again, since a YAML tool rewriting `Chart.yaml` drops it without any signal.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -83,7 +83,9 @@
       customType: 'regex',
       datasourceTemplate: 'docker',
       depNameTemplate: 'traefik',
-      managerFilePatterns: ['/VALUES.md/'],
+      // Anchored: an unanchored `/VALUES.md/` regex also matches hub-manager/VALUES.md,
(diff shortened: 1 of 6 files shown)


bpsoraggi (approved):
LGTM, tested locally

PR #1963, opened 2026-08-06
fix(changelog): only update the chart being released

mloiseleur:
### What does this PR do?

`make changelog` now only touches the chart being released.

`hack/changelog.sh` skips a chart when the top section of its `Changelog.md` does not match its `Chart.yaml` version, and restores the `Changelog.md` helm-changelog wrote for it. The `Makefile` restores `.github/fixtures` after the helm-changelog run.

Adds a `no-release` golden test case covering the skip.

### Motivation

Since #1925, preparing a `traefik` release also rewrote `hub-manager`, which had no release of its own:

- helm-changelog opens a `## Next Release` section on every chart touched since its last release, and #1961 touched `hub-manager/Chart.yaml`.
- That second `## ` heading pushed `release_count` to 2, so the already published `1.0.0` matched the major-release branch and its `artifacthub.io/changes` annotation was replaced by the upgrade-notes link.
- helm-changelog has no path filter, so the same run also rewrote the four golden fixtures under `.github/fixtures`, which are test inputs.

### More


bpsoraggi (approved):
LGTM

PR #1972, opened 2026-08-17
fix(rbac): restore secretResourceNames for namespaced role

### LINKED ISSUE #1745: rbac.secretResourceNames support removed even for namespaced deployments
### Welcome!

39.0.5

packaged within chart

rbac.secretResourceNames to keep working for namespace scoped deployments

support `rbac.secretResourceNames` fully removed by  https://github.com/traefik/traefik-helm-chart/pull/1683

### What are your values ?

```
rbac:
  enabled: true
  namespaced: true
  secretResourceNames:
    - tls-secret #! adjust
```

### Additional Information

https://github.com/traefik/traefik-helm-chart/pull/1683 breaks my setup.

I've been running namespaced Traefik with rbac.secretResourceNames scoped to only the TLS secret within the namespace. This has been working correctly for at least 6 months — certs were being renewed and Traefik could access them without issue.

#1653 specifically mentions cluster-scoped deployment, where restricting secretResourceNames breaks Traefik's informer because Kubernetes RBAC doesn't support resourceNames filtering on a cluster-scope list.

However, with namespaced: true, no ClusterRole is created at all — only a namespace-scoped Role. In that case, resourceNames restrictions on list work correctly because the scope is already bounded to the namespace.

Could you confirm whether removing secretResourceNames is intentional for namespaced deployments as well, or whether this was only meant to address the cluster-scoped case? If the former, it would be worth documenting that namespaced users need to manage secret RBAC manually outside of Helm or role out fix for that

HarnageaGabriel:
## What
Restores `rbac.secretResourceNames`, but scoped only to the namespaced `Role` case (`rbac.namespaced: true`). `templates/rbac/clusterrole.yaml` is left untouched.

## Why
#1683 removed `rbac.secretResourceNames` entirely to fix a real bug: Kubernetes RBAC does not support `resourceNames` on a `ClusterRole`'s list/watch rules, so setting it broke Traefik's secret informer when `rbac.namespaced: false`.

That fix over-removed the value: when `rbac.namespaced: true`, Traefik only gets a namespaced `Role` (not a `ClusterRole`), and `resourceNames` on a `Role` works correctly. Restricting it to one secret (e.g. Traefik's own TLS certificate) instead of every secret in the namespace is a legitimate, useful restriction that #1683 removed as collateral damage.

The maintainer confirmed this direction in [the issue](https://github.com/traefik/traefik-helm-chart/issues/1745#issuecomment-3080146449): re-implement a version of `secretResourceNames` tailored for the namespaced use case.

## Changes
- `traefik/templates/rbac/role.yaml`: re-add the conditional `resourceNames` block for the secrets rule (same logic #1683 removed).
- `traefik/templates/rbac/clusterrole.yaml`: **unchanged** — cluster-scoped RBAC keeps unrestricted secret access, per #1683.
- `traefik/values.yaml`: re-add `rbac.secretResourceNames: []`, with an updated doc comment clarifying it only applies when `rbac.namespaced: true`.
- `traefik/values.schema.json`, `traefik/VALUES.md`: regenerated via `helm schema` / `helm-docs`.
- `traefik/tests/rbac-config_test.yaml`: re-add only the two `Role`-scoped test cases from #1683's removed tests (empty vs. set `secretResourceNames` under `rbac.namespaced: true`). No `ClusterRole` test cases are re-added.

## Test plan
- [x] `helm unittest -f tests/rbac-config_test.yaml traefik` — 41/41 pass, including the 2 new `Role` cases.
- [x] `helm unittest ./traefik` (full suite) — same 47 pre-existing, unrelated failures as on `master` (tracing/otlp arg formatting), zero regressions from this change (708 passed vs. 706 on `master`, the +2 being the new tests).
- [x] Confirmed `templates/rbac/clusterrole.yaml` renders identically to `master` (no `resourceNames` for secrets in `ClusterRole`, regardless of `secretResourceNames`).

Fixes #1745

### DIFF
--- a/traefik/VALUES.md
+++ b/traefik/VALUES.md
@@ -553,6 +553,7 @@ Kubernetes: `>=1.25.0-0`
 | rbac.aggregateTo | list | `[]` | Enable user-facing roles https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles |
 | rbac.enabled | bool | `true` | Whether Role Based Access Control objects like roles and rolebindings should be created |
 | rbac.namespaced | bool | `false` | When set to true: <br /> 1. It switches respectively the use of `ClusterRole` and `ClusterRoleBinding` to `Role` and `RoleBinding`.<br /> 2. It adds `disableClusterScopeResources` on Ingress and CRD (Kubernetes) providers<br /> **NOTE**: `IngressClass`, `NodePortLB` and **Gateway** provider cannot be used with namespaced RBAC. <br /> See [upstream documentation](https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-ingress/#opt-providers-kubernetesIngress-disableClusterScopeResources) for more details. |
+| rbac.secretResourceNames | list | `[]` | List of Kubernetes secrets that are accessible for Traefik when `rbac.namespaced` is true. If empty, then access is granted to every secret. Ignored when `rbac.namespaced` is false (ClusterRole), since Kubernetes RBAC does not support `resourceNames` on cluster-scoped list/watch rules. |
 | readinessProbe.failureThreshold | int | `1` | The number of consecutive failures allowed before considering the probe as failed. |
 | readinessProbe.initialDelaySeconds | int | `2` | The number of seconds to wait before starting the first probe. |
 | readinessProbe.periodSeconds | int | `10` | The number of seconds to wait between consecutive probes. |
--- a/traefik/templates/rbac/role.yaml
+++ b/traefik/templates/rbac/role.yaml
@@ -45,6 +45,9 @@ rules:
       - ""
     resources:
       - secrets
+    {{- if gt (len $.Values.rbac.secretResourceNames) 0 }}
+    resourceNames: {{ $.Values.rbac.secretResourceNames }}
+    {{- end }}
     verbs:
       - get
       - list
--- a/traefik/templates/requirements.yaml
+++ b/traefik/templates/requirements.yaml
@@ -39,6 +39,10 @@
   {{- end }}
 {{- end }}
 
+{{- if and (not .Values.rbac.namespaced) (not (empty .Values.rbac.secretResourceNames)) }}
+  {{- fail "ERROR: rbac.secretResourceNames requires rbac.namespaced to be true. Kubernetes RBAC does not support resourceNames on cluster-scoped list/watch rules." }}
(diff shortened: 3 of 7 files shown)


traefik/VALUES.md
```diff
@@ -548,6 +548,7 @@ Kubernetes: `>=1.25.0-0`
 | rbac.aggregateTo | list | `[]` | Enable user-facing roles https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles |
 | rbac.enabled | bool | `true` | Whether Role Based Access Control objects like roles and rolebindings should be created |
 | rbac.namespaced | bool | `false` | When set to true: <br /> 1. It switches respectively the use of `ClusterRole` and `ClusterRoleBinding` to `Role` and `RoleBinding`.<br /> 2. It adds `disableClusterScopeResources` on Ingress and CRD (Kubernetes) providers<br /> **NOTE**: `IngressClass`, `NodePortLB` and **Gateway** provider cannot be used with namespaced RBAC. <br /> See [upstream documentation](https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-ingress/#opt-providers-kubernetesIngress-disableClusterScopeResources) for more details. |
+| rbac.secretResourceNames | list | `[]` | List of Kubernetes secrets that are accessible for Traefik when `rbac.namespaced` is true. If empty, then access is granted to every secret. Ignored when `rbac.namespaced` is false (ClusterRole), since Kubernetes RBAC does not support `resourceNames` on cluster-scoped list/watch rules. |
```
mloiseleur, line 556: >  Ignored when `rbac.namespaced` is false (ClusterRole), since Kubernetes RBAC does not support `resourceNames` on cluster-scoped list/watch rules.

Would you please add a fail-fast check in `requirements.yaml` for better UserXP?

HarnageaGabriel: Added — see 4acafff: `requirements.yaml` now fails template rendering when `rbac.secretResourceNames` is set while `rbac.namespaced` is false, with message `ERROR: rbac.secretResourceNames requires rbac.namespaced to be true. Kubernetes RBAC does not support resourceNames on cluster-scoped list/watch rules.` Follows the same `fail` pattern as the other checks in that file. Covered by new tests in `requirements-config_test.yaml` (invalid combo fails, valid combo passes).

mloiseleur (approved):

darkweaver87 (approved):
LGTM

PR #1982, opened 2026-08-27
feat: add `aliasHeadersStrategy` value

jnoordsij:
### What does this PR do?

Add an `aliasHeadersStrategy` value to allow usage with Traefik v3.7.12.

See also #1981.

### Motivation

https://doc.traefik.io/traefik/v3.7/migrate/v3/#v3712

### More

- [x] Yes, I updated the tests accordingly
- [x] Yes, I updated the schema accordingly
- [x] Yes, I ran `make test` and all the tests passed

PR #1983, opened 2026-08-27
feat: deprecate `underscoreHeadersStrategy` value

jnoordsij:
### What does this PR do?

Deprecate the `underscoreHeadersStrategy` value, which has been replaced with `aliasHeadersStrategy` in Traefik v3.7.12.

The emitted deprecation notice is worded vaguely by design, so that we may freely decide on whether or not to already drop this in the next major chart release, wait a bit or even wait until Traefik v4.

Requires #1981 and #1982 to be merged first.

### Motivation

Help users keep their values up-to-date.

### More


mloiseleur (approved):

darkweaver87 (approved):
LGTM

PR #1998, opened 2026-09-07
fix: grant ConfigMap write verbs to namespaced role

jspdown:
### What does this PR do?

Adds the `configmaps` write verbs (`create`, `patch`, `delete`) to the namespaced `Role`, which so far only the `ClusterRole` had.

### Motivation

API Management stores OpenAPI specs in ConfigMaps, so it fails to start with `rbac.namespaced=true`.

### More


mloiseleur (approved):

darkweaver87 (approved):
LGTM

komalsukhani (approved):

PR #2001, opened 2026-09-09
fix: use kubeversion to determine PDB apiVersion

### LINKED ISSUE #2000: PodDisruptionBudget defaulting to removed apiversion when templated
### Welcome!

39.0.9

3.6.x

When using `helm template` or Kargo's function [helm-template](https://docs.kargo.io/user-guide/reference-docs/promotion-steps/helm-template) with the chart, for the rendered manifest to be compatible with a cluster running same version as provided `--kube-version`.

PodDisruptionBudget is created using API version `policy/v1beta` instead of `policy/v1`.

### What are your values ?

```
# Only included relevant bit
podDisruptionBudget:
  enabled: true
  minAvailable: 1
```

### Additional Information
This appears to happen due to the template using `.Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget"` to determine whether to use `v1` or to fallback to `v1beta1`.

The `.Capabilities.APIVersions.has` [property is not available during templating](https://github.com/helm/helm/issues/10760) unless using server-side validation of output, and thus cannot be done on a system without the cluster context present.

Might I suggest using `.Capabilities.KubeVersion.GitVersion` instead or defaulting to `policy/v1` since `policy/v1beta1` has been removed from the lowest Kubernets version supported by this chart?

Thank you!

Koronei:
### What does this PR do?

Ensure predictable templating by using `Capabilities.KubeVersion` instead of `Capabilities.ApiVersions` to determine the `apiVersion` to use for `PodDisruptionBudget`.

### Motivation

Releasing traefik through pre-rendered manifests and GitOps can produce incompatible resources, in this case seemingly limited to the PDB.

https://github.com/traefik/traefik-helm-chart/issues/2000

### More

### DIFF
--- a/traefik/templates/poddisruptionbudget.yaml
+++ b/traefik/templates/poddisruptionbudget.yaml
@@ -1,9 +1,5 @@
 {{- if .Values.podDisruptionBudget.enabled -}}
-{{- if semverCompare ">=1.25-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: policy/v1
-{{- else }}
-apiVersion: policy/v1beta1
-{{- end }}
 kind: PodDisruptionBudget
 metadata:
   name: {{ template "traefik.fullname" . }}
--- a/traefik/templates/poddisruptionbudget.yaml
+++ b/traefik/templates/poddisruptionbudget.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.podDisruptionBudget.enabled -}}
-{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+{{- if semverCompare ">=1.25-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: policy/v1
 {{- else }}
 apiVersion: policy/v1beta1


traefik/templates/poddisruptionbudget.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.podDisruptionBudget.enabled -}}
-{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+{{- if semverCompare ">=1.25-0" .Capabilities.KubeVersion.GitVersion -}}
```
darkweaver87: Since the chart floor is already 1.25, this whole conditional can go — including the `{{- end }}` on line 6 (outside this hunk, so I can't wrap it in a suggestion):

```gotemplate
{{- if .Values.podDisruptionBudget.enabled -}}
apiVersion: policy/v1
kind: PodDisruptionBudget
```

WDYT ?

darkweaver87 (commented):
Thanks for digging into this! The intent is right, but the guard is actually unreachable either way: `Chart.yaml` declares `kubeVersion: '>=1.25.0-0'` and Helm enforces it (`helm template --kube-version 1.21.0` → `chart requires kubeVersion: >=1.25.0-0`), so `.Capabilities.KubeVersion` can never be below 1.25 here — and `policy/v1beta1` is removed as of 1.25 anyway.

The real bug in #2000 is that `.Capabilities.APIVersions` defaults to just `{"v1"}` when rendering without a cluster, so `helm template` silently falls into that dead branch. Simplest fix is to drop the conditional entirely.

traefik/templates/poddisruptionbudget.yaml
Koronei: Completely agree. The condition is not necessary given the requirements of the chart.

darkweaver87 (approved):
LGTM, @mloiseleur or @jnoordsij any thought on this ?

jnoordsij (approved):
LGTM, I love cleanups!

Seems to be ok indeed with our minimum Kubernetes version. I don't think these versions are really used in the wild anymore. In particular I wonder if the upstream application will even work with older versions still...

PR #2004, opened 2026-09-16
fix: add enabled option to ports

### LINKED ISSUE #1971: traefik: values.schema.json does not allow `false` as port value
### Welcome!

41.2.0

3.7.10

When I deploy traefik I disable the `web` port. With argocd versions older than 3.5.0, which uses helm 3.x, I could simply set `ports.web=null` and the port is gone. With argocd version 3.5.0 and newer, which uses helm 4.x, the port now gets rendered again. Looking into the template code reveals that the port config is checked to not be falsey, which is why the null-value worked. Setting `ports.web=false` works, but is not accepted by the schema and so I have to disable schema validation in argocd.

so my expectation is:

ports.web=false disables the port and is accepted by the schema.

ports.web=false works, but the schema doesn't accept it.

### What are your values ?

-

### Additional Information

```shell
Alternatively an explicit `enabled` option on ports might be reasonable.
```

dmaes:
### What does this PR do?

Add an explicit `enabled` option to ports.

### Motivation

Currently, a port must be disabled by setting it to `null`.
However, Helm v4 has a bug where, in subcharts, it merges `null` values as if they were an empty dict. Meaning we can no longer disable ports this way when using Helm v4 (or any other tool that uses Helm v4, like ArgoCD v3.5.0)

Fixes #1971 

### More


pschichtel:
lgtm

mloiseleur (commented):
This PR is a breaking change for all users who have configured custom ports.

pschichtel:
@mloiseleur how? I've only seen checks for enabled=false

dmaes:
@mloiseleur I tested it with custom ports, works fine, `enabled` defaults to `true` when not defined. And the pre-existing unit test for custom ports is still green, without me touching it.

